PR #90, opened 2017-09-20
location.pathname typo

laurentpayot:


bahmutov:
Thank you for noticing and fixing this @laurentpayot

PR #140, opened 2017-10-16
replace <record_key> with example key because '<' was not escaping properly. close #128

### LINKED ISSUE #128: Record key is hidden (not being escaped properly)
screen shot 2017-10-09 at 12 35 15 pm

jennifer-shehane:


jennifer-shehane:
Yeah, I didn't even look into this one at all - didn't want to waste the time.

PR #149, opened 2017-10-17
add examples for binding to Cypress events

brian-mann:
- uncaught exceptions
- test failures
- page navigation
- window before load
- window confirm
- window alert


jennifer-shehane:
It doesn't appear that any of your changes from `develop` yesterday were actually released to `docs.cypress.io` from `master`, I see that you manually merged - I would recommend opening a PR from develop to master so there is at least a history trail to track down when these errors in deployment happen.

PR #163, opened 2017-10-23
Fix display of footer in mobile view. close #157

### LINKED ISSUE #157: Footer does not display properly in mobile view
img_8171

jennifer-shehane:


bahmutov:
really nice, search is huge. I was trying to find things in buildkite API docs that does not have search and it was a huge pain.

PR #185, opened 2017-10-30
unable to locate package libsound

walkerrandolphsmith:


brian-mann:
Good catch. Looks like we have it correctly in our Dockerfile but improperly listed it in the CI docs.

https://github.com/cypress-io/cypress-docker-images/blob/master/Dockerfile

walkerrandolphsmith:
Awesome!

Issue #177, opened 2017-10-26
Write introduction for build & test app tutorial


avanslaars:
Create an entry point for the tutorial that gives a high-level overview of the topics that will be covered. When the video is complete, this page should include a "highlight reel" in either video or gif form, depending on length

Issue #174, opened 2017-10-25
Write First Spec page


avanslaars:
- [ ] Write out tutorial
- [ ] Link keywords to appropriate docs 
- [ ] Publish and link to video
- [ ] Create screenshots & gifs

PR #191, opened 2017-10-31
Issue 160 - Add tutorials section to docs

### LINKED ISSUE #160: Add top-level section for Tutorials
Tutorials will appear in the main navigation and we will use the left nav to handle multiple tutorials, each with their own multi-page navigation when necessary.

avanslaars:
Added top level tutorial section, along with the "Introduction", "Project setup" and "Testing inputs" pages for the "Build and test TodoMVC with Cypress" tutorial.
Closes #160, #173, #174, #177


jennifer-shehane:
I had a lot of feelings on this, so much so that I branched from this PR here so you can preview: https://github.com/cypress-io/cypress-documentation/pull/192

**My thoughts:**
- The videos should be front and center. The layout of the current 'guides' doesn't work. I made a VERY quick mockup of a 'video' layout, where there is no right side nav, the video is larger, and text content is below. I think we could make this better - emphasize how long the video is, maybe even put a little github logo of the author, well - anything! This should be about the video, not conform to our previous format.
- The introduction is not that great. :( Sorry. It does explain what the tutorial will cover, but does not address the WHY. Why do I need to watch this? Why is this critical to me as a developer? What will I walk away knowing that I didn't know before? What can I tell my boss about why I need to spend an hour today watching these tutorials. Also - it's missing some of the core buzzwords. I know it doesn't like, totally teach you how to build a React app, but we're building a React app - this seems important to mention. 
- I still like the idea of the 'contents' table with how long the videos are. I know you said it like...doesn't totally work cause there's text content, but this is honestly about the videos I think.

jennifer-shehane:
- I added tests for the tutorial pages 
- I added a 'More tutorials to come' at the bottom of the list. 
- I renamed the tutorial to 'Test a React Todo App', I think saying 'Build and test' is kind of a stretch since we start with 'starter' code. This is just common, how you should test things with true TDD imo. And the title just looks simpler. What do you think?

I think it's probably ok to push now. We can improve on the layout later as more videos come together.

avanslaars:
That all works for me. The reason I initially avoided React in the title was to prevent people who use Vue, Angular, etc from avoiding looking at it, but I guess we run the risk of those same devs just bailing out once they see React anyway.

Thanks for all your help in getting this ready.

PR #192, opened 2017-10-31
Updates to tutorials structure

jennifer-shehane:
- add box shadow + margin to videos
- move ‘what you’ll learn below videos’
- add specific ‘video’ layout that displays videos as larger
- remove most of content from introduction :grimacing:
- add a ‘contents’ section to main page.

### DIFF
--- a/source/tutorials/build-and-test-todo/introduction.md
+++ b/source/tutorials/build-and-test-todo/introduction.md
@@ -3,36 +3,17 @@ title: Build and test TodoMVC with Cypress
 comments: false
 ---
 
-{% note info %}
-# {% fa fa-graduation-cap %} What you'll learn
-
-- How to run Cypress from an `npm` script
-- How to Use Cypress to test while developing features
-- How to use the Cypress UI to help create your tests
-- How to visit pages, interact with elements and make assertions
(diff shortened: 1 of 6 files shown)


avanslaars:
I think this looks great! What do you think about adding a row to the table for the upcoming video while we're dripping this content out? Something to make it clear that this tutorial isn't done yet and people should come back to look for the next step?

jennifer-shehane:
That's probably a good idea. I wouldn't get too carried away with outlining specific deliverables, but giving a general understanding that there's more to come would be great.

brian-mann:
You could also simply label all the videos as like (1/5), (2/5) to indicate there are more in the series.

avanslaars:
Yeah, I was thinking something fairly general and just one entry.

I've had quite a few occasions where something seems like a logic grouping, and when a video is done, it makes sense to split it into two separate videos. I worry that the `1/n` approach kind of locks us in and if it doesn't indicate clearly that something is coming in the future, will we get issues opened like "tutorial is only showing the first 2 videos, but it says there should be 5".

I guess we can look at variations and decide from there. I'm going to merge this into my original PR and I'll add something from there.

PR #215, opened 2017-11-03
Update continuous-integration.md

anned20:
We were missing x264 from our testing environment and that caused Cypress to hang at processing the video

### DIFF
--- a/source/guides/guides/continuous-integration.md
+++ b/source/guides/guides/continuous-integration.md
@@ -83,7 +83,7 @@ Here are a few CI example repositories using our Docker images
 If you are not using one of the above CI providers then make sure your system has these dependencies installed.
 
 ```shell
-apt-get install xvfb libgtk2.0-0 libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2
+apt-get install xvfb libgtk2.0-0 libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 x264
 ```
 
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
Hi @anned20. Thanks for contributing! Could you sign our CLA before we accept your contribution?

brian-mann:
@bahmutov do we install `x264` in our linux docker images?

bahmutov:
No we don't and @anned20  what is your testing environment?

anned20:
Sorry, I fucked up. I thought that fixed it, but it didn't. 

So, our testing environment is a Vagrant box with Debian Jessie. The tests are working, but just before the video processes, Cypress hangs. I thought installing x264 fixed it, but I was wrong.

bahmutov:
@anned20 is the runtime 32-bit or 64-bit? because we include 64 bit ffmpeg which might crash on 32 bit. This is why in windows build we manually include both ffmpeg version (increasing its size). In the future we should build separate 32-bit and 64-bit builds for every platform

anned20:
I'll have to check that. Will do that tomorrow or Monday. Thank you for your help!

anned20:
I just checked and the runtime is 64bit. The Vagrant image is based on "debian/jessie64". The tests are passing (failing), the screenshots are getting print and then it just hangs.

anned20:
It's not throwing an error either with debug mode on

PR #220, opened 2017-11-04
Issue 194

brian-mann:


source/guides/references/best-practices.md
```diff
@@ -77,6 +77,45 @@ WIP.
...
+```js
+// DONT DO THIS. IT DOES NOT WORK
+// THE WAY YOU THINK IT DOES.
+
+const button = cy.get('button')
+
+const form = cy.get('form')
+
+// nope, fails
+button.click()
```
keeyip-at-outreach: It's 2019 but I'm having a hard time understanding why `button.click()` fails in this example.

In this case, wouldn't `button` be a reference to a chainer? So then wouldn't `button.click()` simply be enqueued and thus *work*?

Perhaps a clearer example not inappropriate use would be to try and make a synchronous check on `button`:
```
// Incorrect way to check the button's disabled state.
if (button.its('disabled')) {
  ...
}

// Correct way
button.its('disabled').then(disabled => {
  if (disabled) {
    ...
  }
});

Comment (jennifer-shehane) [reply source/guides/references/best-practices.md]: We recommend questions relating to how to use Cypress be asked in our [community chat](https://gitter.im/cypress-io/cypress). Also try searching our [existing GitHub issues](https://gitter.im/cypress-io/cypress/issues), reading through our [documentation](https://docs.cypress.io), or searching [Stack Overflow](https://stackoverflow.com/questions/tagged/cypress) for relevant answers.

Alternatively, up for any of our [paid plans](https://www.cypress.io/pricing/) gives you access to our team for support via email.

Comment (keeyip-at-outreach) [reply source/guides/references/best-practices.md]: This isn't a question about how to use Cypress, I am pointing out a potential error in the documentation. The documentation is claiming that `button = get(..); button.click()` fails, but since `button` is a reference to a `Chainer` object, `button.click()` should *eventually* succeed.

This code example might have been written at a time when the `Chainer` object was not as sophisticated, or perhaps to try to discourage users from assuming that they are writing synchronous code, but this sample code is incorrect because the claim that `button.click()` fails is contrary to what is actually going on. As a developer trying to get a firm understanding of how Cypress works the code examples should be as correct as possible.

My request is that someone take a deeper look at this code example for correctness because this very same code example is shown on several introductory sections of the documentation; someone trying to learn Cypress will be quite confused if this basic example is incorrect or inconsistent to how Cypress actually works.

Comment (keeyip-at-outreach) [reply source/guides/references/best-practices.md]: Thanks for taking another look.

But I think the new example still seems like it can be more direct at illustrating the "Cypress is not *synchronous* (as you might assume)" statement.

Here's why:
`cy.visit()` *is essentially a synchronous operation* (to my knowledge) because it will immediately wipe out the existing DOM, cancel any form submissions, etc.., and navigate to a totally different URL; so the reason `a.first().click()` fails in your example is due to the previous DOM/subject being wiped clean after `cy.visit()`.

Here's a more direct example of Cypress's async mode of operation:
```
const button = cy.get('button[type="submit"]');

// Accessing `button[0]` fails because `button` is not a jQuery object,
// and `button[0]` is not a native DOM element.
// Instead, `button` is actually a *chainer* object that will
// look for the `button[type="submit"]` element at a later time.
try {
  // This will fail.
  console.log(button[0].innerText);
} catch (err) {
  // You will get into this catch statement instead.
}

// You can use `.then()` to get the jQuery object when it's **available**
button.then($button => {
  // `$button` is now a populated jQuery result.
  console.log($button[0].innerText);
});

// But `button.click()` is sufficient to instruct Cypress to click on the button whenever it is available.
// Cypress has created methods that mirror jQuery's convenient API, but Cypress' methods
// will also take care of automatically waiting for DOM elements to load.
button.click();
```

jennifer-shehane: Please open a pull request or issue with your proposed suggestions. A PR from 2017 isn't the best place to continue this discussion - no one outside of the original contributors for this PR will be able to see this message or be able to give their opinions. Thanks!

PR #234, opened 2017-11-12
Search background in safari

calendee:
**Issue**
In Safari, the search input field is white.  Since the text color of the search field is also white, the user cannot see the search string being entered.

**Fix**
To overcome default webkit search field styling, change the appearance property to be a textfield and then applied the same background color as the header.


bahmutov:
Thank you, the change is working on safari and does fix the white text problem.

PR #235, opened 2017-11-12
typo on "lower than" expectation

euZebe:


bahmutov:
Great, thanks for noticing and submitting a quick fix

Issue #193, opened 2017-10-31
Add example to cy.stub() for stubbing window.prompt behavior


jennifer-shehane:
```js
cy.visit('http://localhost:3000')
cy.window().then((win) => {
   cy.stub(win, 'prompt').returns('what i want typed into it')
})
```

richdouglasevans:
Can this issue be closed?

Also, good work Team Cypress on really digging into that `first timers only` drive 👍

srinivasrk:
Does it handle prompts generated anywhere in the page ?

PR #294, opened 2017-12-01
Add window.prompt stubbing example to api section

microbouji:
This addresses #193. I followed the pattern of the "Stubbing window.fetch" example recipe to make the change before the app runs.

Also, thank you folks for being friendly to new contributors. The "first-timers-only" label of this issue is what made the difference for me between starting 24 pull requests or not.


microbouji:
Sorry about missing that! Fixed it.

bahmutov:
great example, thank you @microbouji

PR #307, opened 2017-12-06
Reference recommendation not to use after(Each)

DanAtShenTech:
Link to the recommendation not to use the anti-pattern of cleaning up state with after()/afterEach()

### DIFF
--- a/source/guides/core-concepts/writing-and-organizing-tests.md
+++ b/source/guides/core-concepts/writing-and-organizing-tests.md
@@ -174,7 +174,7 @@ describe('Hooks', function() {
 - All `after()` hooks run (once)
 
 {% note danger %}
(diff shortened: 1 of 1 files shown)


DanAtShenTech:
I couldn't determine how to use the 'url' helper in order to make the proper link.  If you agree with this addition to the docs, will you update the PR with the necessary change to 'url'?

brian-mann:
The URL helper references the unique name of the document - in this case it would be "best-practices"

And then if you use a hash it will go out and validate that that hash could be found. So the url helper should be...

```
{% url 'text' nameOfDoc %}
{% url 'text' nameOfDoc#hash %}
{% url 'the text you want here' best-practices#Using-after-or-afterEach-hooks %}
```

DanAtShenTech:
Thanks Brian.

jennifer-shehane:
Sorry, yup, I did not write the link right. Got caught up in fixing the `docs` url and the markdown link. Brian's change is the fix. ^

PR #315, opened 2017-12-07
Added vimeo link to debugging guide

avanslaars:
Just one link for now... Can't embed with a timestamp yet, so for now, it's a link out to vimeo.

Mostly wanted to get feedback before adding more of these. I'll look into adding the Vimeo JS api to the project so we can control the embeds as a separate effort.


brian-mann:
I'm all for linking to timestamps - but if we can't figure out how to do it with a video embed, then we at least have to rename all the videos on vimeo so that they match the titles in the docs.

We should also have a link from the vimeo videos to the tutorials as well. As it stands, its very difficult to correlate the names of the videos to what we call them on the docs.

I'm referring to looking at this link: https://vimeo.com/242961930#t=264s

avanslaars:
I've updated video titles and made sure the newest videos have links back to the tutorial

jennifer-shehane:
@avanslaars @brian-mann There is also a way to set a custom url, so that the title is in the url to easily distinguish. It's in the Advanced settings:

screen shot 2017-12-11 at 10 20 39 am

avanslaars:
@jennifer-shehane - That's great! I had no idea that existed :). I can add titles that match the docs and we can link to those instead.

PR #327, opened 2017-12-12
Remove wording that command.js is loaded automatically

DanAtShenTech:
Include statement to un-comment line in index.js in order to import command.js.
Reference:  https://stackoverflow.com/questions/47695535/cypress-custom-command-is-not-recognized-when-invoked/47707642#47707642

### DIFF
--- a/source/api/cypress-api/custom-commands.md
+++ b/source/api/cypress-api/custom-commands.md
@@ -6,7 +6,7 @@ comments: false
 Cypress comes with its own API for creating custom commands and overwriting existing commands. The built in Cypress commands use the very same API that's defined below.
 
 {% note info  %}
-A great place to define or overwrite commands is in your `cypress/support/commands.js` file, since it is loaded before any test files are evaluated.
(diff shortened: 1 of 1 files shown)


brian-mann:
It is not commented by default.

We seed both the `cypress/support/index.js` and the `cypress/support/commands.js` file, and the `index.js` is already built to import the other.

You would have to **undo** work to prevent these from working.

jennifer-shehane:
The current `index.js` that is seeded. The require statement (last line) is commented out. 
```js
// ***********************************************************
// This example support/index.js is processed and
// loaded automatically before your test files.
//
// This is a great place to put global configuration and
// behavior that modifies Cypress.
//
// You can change the location of this file or turn off
// automatically serving support files with the
// 'supportFile' configuration option.
//
// You can read more here:
// https://on.cypress.io/configuration
// ***********************************************************

// Import commands.js using ES2015 syntax:
import './commands'

// Alternatively you can use CommonJS syntax:
// require('./commands')
```

brian-mann:
Read the text comments, they explain exactly what is happening. It is importing `commands` exactly as I described. Nothing needs to change here.

PR #341, opened 2017-12-20
update to include Windows debugging

aaronherman:
To enable Windows debugging, you have to set the environment variable differently. Improved documentation for Cypress (and instead of relying on looking at visionmedia/debug) will be helpful.


aaronherman:
No worries! I had to run the two commands separately. If they're together, then `cypress open` does not run. I can update documentation to account for that?

jennifer-shehane:
@aaronherman Yes, perhaps add another line to the shell command then, since there would be no output without running either `cypress open` or `cypress run` after the set env var.

aaronherman:
Done!

PR #356, opened 2018-01-03
document CLI helpers for starting and stopping server, close #355

### LINKED ISSUE #355: Describe how to run server and tests and shutdown
Often see this question 

```
Hi, what is the recommended say to start your normal dev server before cypress run? 
I tried an NPM script as "test": "yarn start; yarn cypress run" but running into 
background-process issues and the server isn't closed afterwards.
```

Need to show two ways we do it: via `run-p --race server test` using `npm-run-all` and using https://github.com/bahmutov/start-server-and-test

bahmutov:
Explain `run-p --race` and `start-server-and-test`


jennifer-shehane:
@brian-mann said he had issues with this PR that was externally discussed with you to be addressed.

brian-mann:
Yeah I am not going to merge this in. It conflicts with what is **already** written in this guide. I would never recommend `run-p` in any situation involving Cypress - never. Waiting on the server to load is **always** the best solution.

https://github.com/bahmutov/start-server-and-test is a good step but I wouldn't recommend this in its current form without changes to the API.

We already provide step-by-step examples on how to issue the command line to boot your server and run tests. If we want to provide an NPM module then I would make it an official Cypress one and I would improve the interface before handing it off.

bahmutov:
yeah, @brian-mann thinks `run-p --race` is unnecessary, should just go to `wait on server to start`

bahmutov:
Ok, maybe I would agree with skipping `run-p --race` - even if _we use it all the time_. But how would we need to update https://github.com/bahmutov/start-server-and-test ? I don't see any opened issues for it, and it is news to me that there is something missing there.

brian-mann:
The name is confusing, and the arguments are especially confusing.

Seeing it written in `package.json` is a giant (?)

I would rename it to be:

- `start-server-and-run-tests`
- `boot-server-and-run-tests`
- `wait-on-server-and-run-tests` <-- probably this one

I would accept a `--silent` flag to silence the server `stdout` else it will mix into the output from your tests.

I would accept a `--wait-on` argument for the url of the server to wait on, so it's clear.

I would accept a `--server` command, so its clear that's what's booting the server.

I would accept a `--tests` command so its clear that's what's booting the tests.

Something like this...

```bash
wait-on-server-and-run-tests --wait-on http://localhost:3030 --server npm run server --tests cypress run
```

It is debatable whether you should default to `npm run`. I get how doing that would truncate the command, and avoid the arguments. But this is the kind of thing that is written once and once only and I believe being clear to other users (who didn't write it) is far more practical than saving keystrokes.

brian-mann:
Also doing it my way (passing args) would enable you to not pass a `--server` argument in the case where the system under test is not actually booting the server, but is simply waiting for it to become available in the case of another system booting it.

bahmutov:
ok, now finally I got some feedback, which can be opened as an issue. Otherwise this was sitting silent in the pull request queue. I disagree with variety of options and CLI flags. The package should do one thing and do it well. Maybe the name could be adjusted to `start-wait-test`. That is the most common situation after all.

brian-mann:
Regardless of the changes to the module this content literally conflicts with the other content on the same guide. It has to be consolidated and zipped up properly. Else we tell the user two conflicting things within paragraphs of each other.

bahmutov:
ok I removed the `run-p` section. The user now is instructed how to do this manually step by step or how to use `start-server...`

jennifer-shehane:
I've been seeing this test fail more and more often, seems flaky.

 1) "before each" hook for "posts to Algolia api with correct index on search"

bahmutov:
I have expanded travis and circle snippets too - the user should be able to copy / paste them now.

PR #373, opened 2018-01-13
Switch example back to using jQuery selector

jamoonlore:
Just realized that the example was using a jQuery selector extension. So it's not a typo.


jamoonlore:
Sorry about that

PR #375, opened 2018-01-13
Fix redundant link and add punctuation

jamoonlore:


jamoonlore:
Sorry, I want to undo this one.

Issue #236, opened 2017-11-16
Update overriding env variables wording to be clearer


jennifer-shehane:
> Values in here will overwrite conflicting values in cypress.json.

From discussion on https://github.com/cypress-io/cypress/issues/909, this should really say:

> Values in here will overwrite conflicting environment variables in cypress.json.

clairecliu:
I would like to give this a try.

danverde:
The [docs](https://docs.cypress.io/guides/references/configuration.html#Environment-Variables) currently say:

> You can also use environment variables to override configuration values. This is especially useful in Continuous Integration or when working locally. This gives you the ability to change configuration options without modifying any code or build scripts.
> 
> By default, any environment variable that matches a corresponding configuration key will override the cypress.json value.

From what I understood of [cypress-io/cypress#909](https://github.com/cypress-io/cypress/issues/909) this is incorrect & needs to be updated.

PR #382, opened 2018-01-13
Update overriding env variables wording to be clearer

clairecliu:
- updated overriding env variables wording to be clearer in source guide.

address #236


surfmuggle:
My attempt to just use `cy.visit()` failed with: `cy.visit() must be called with a url or an options object containing a url as its 1st` In gitter Gleb Bahmutov told me that after setting baseUrl you still have to use `cy.visit("/")` 

It would be nice to have this in the docs as well.

PR #431, opened 2018-03-02
Changed first instance of "subject" to be quoted

davidgilbertson:
Maybe a bit picky, but this changes the _first_ instance of the term 'subject' to be quoted, not the second.


davidgilbertson:
Speaking of 'like', Cypress is pretty amazing so far. And kudos on the thorough documentation. I love how it describes the reasoning behind the way the thing works along the way.

Issue #216, opened 2017-11-03
Add assertions to '.select()' doc


jennifer-shehane:
See https://github.com/cypress-io/cypress-example-kitchensink/issues/23

ootsuka:
Hi Jennifer, I am a newbie to open source and wondering if I can work on this issue? If yes where's a good place to start?

jennifer-shehane:
Hey @ootsuka, the document that needs to be updated can be found here: https://github.com/cypress-io/cypress-documentation/blob/develop/source/api/commands/select.md which is the source for this page: https://docs.cypress.io/api/commands/select.html

I think we just wanted the current [Examples](https://docs.cypress.io/api/commands/select.html#Examples) updated to include assertions like instead of this:

```js
cy.get('select').select('apples')
``` 

It would show this:

```js
cy.get('select').select('apples')
 .should('have.value', '456')
```

ootsuka:
Thank you. I am trying to include the assertions but I am having trouble with 
```
cy.get('.action-select-multiple')
        .select(['apples', 'oranges', 'bananas'])
```

Do you mind share how you would do the assertion for this one?

jennifer-shehane:
@ootsuka This one definitely is trickier! 

The value of a select with multiple options selected will be an array of the option's values. 

So, we'll need to pull out the value first using [`.invoke()`](https://on.cypress.io/invoke) so that we can make a better comparison with *just* the array. 

We'll also need to use the `deep.equal` assertion since comparing arrays equality is tricky (see [https://github.com/chaijs/deep-eql](https://github.com/chaijs/deep-eql) for more on how that works). 

```js
cy.get('.action-select-multiple')
  .select(['apples', 'oranges', 'bananas']).invoke('val')
  .should('deep.equal', ['456', '457', '458'])
```

ootsuka:
Yes. I was trying to go for the deep equal but could not get the value cause I did not know the invoke(‘val’).
Thanks a ton!

brian-mann:
The question you have to ask yourself is why add an assertion after doing a `.select([])`. If it's because you have application logic that may modify the value - then this is a good test.

However, if you aren't modifying anything, then you don't need an assertion because you can assume Cypress is doing the right thing. If you tell Cypress to select something that's an invalid value - it will already throw an error. So if the `.select()` resolves, then you know it's correct.

jennifer-shehane:
[...]

PR #441, opened 2018-03-05
Add assertion after select

ootsuka:
Add assertion after select addressing #216


ootsuka:
Thanks for all the help!

PR #444, opened 2018-03-06
Warning about cypress.env.json not being watched

FredyC:
Unless I did something wrong on my side, but simply when I created the `cypress.env.json`, it wasn't read. Closing browser with test did not help. I had to exit whole Test runner and then it worked. Also changing values in existing file won't be noticed.


jennifer-shehane:
Can you open this as an issue in the [main repo](https://github.com/cypress-io/cypress/issues/new)? Because this should be watched and I'm hesitant to write a workaround for it without first having an issue.

FredyC:
Looks like the fix is already merged. That's certainly better solution. Closing :)

jennifer-shehane:
Yes, it will go out in our next version release.

PR #445, opened 2018-03-06
Add doc for cy.task()

chrisbreiding:
Depends on https://github.com/cypress-io/cypress/pull/1422 being released.


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "hexo": {
-    "version": "3.6.0"
+    "version": "3.4.1"
```
jennifer-shehane: did you run `npm i`? This should be version 3.6.0.

source/api/commands/task.md
```diff
@@ -0,0 +1,177 @@
...
+cy.task(event, arg, options)
+```
+
+## Usage
+
+**{% fa fa-check-circle green %} Correct Usage**
+
+```javascript
+// in test
+cy.task('log', 'This will be output to the terminal')
```
jennifer-shehane: Oh boy, I worry that this will confuse people between this and `cy.log()` and like, console.log to console....I worry that this example may not be evident enough that this is a log in Node. Is there another simple example we could use that doesn't involve `console.log` that communicates, this is NODE?

source/api/commands/task.md
```diff
@@ -0,0 +1,177 @@
...
+    case 'log':
+      console.log(arg) // This will log to the terminal
+      return null // Signal that the 'log' event has been handled
+  }
+})
+```
+
+***Read a JSON file's contents***
+
+Note: this serves as a demonstration only. We recommend using fixtures or {% url "`cy.readFile`" readfile %} for a more robust implementation of reading a file with Cypress.
```
jennifer-shehane: I don't think we should show an example that we don't recommend using. We should probably come up with some more solid examples...which takes more time unfortunately.

brian-mann:
Agree - we should almost exclusively use `db` oriented tasks that require libs from your app. That's the primary use case of `cy.task` although you could use it for other things as well...

@chrisbreiding the other issue here is providing examples of how to reuse `babel` in the case where you're requiring app files that require being preprocessed.

brian-mann:
@jennifer-shehane i will update the examples to be better. Chris doesn't need to do it.

package.json
chrisbreiding: `npm i` didn't seem to do anything, so I had to manually change it back.

source/api/commands/task.md
chrisbreiding: Yeah, it's hard to come up with terse, non-convoluated examples for this. I like this example because it's something users have asked for (to log to the terminal) and now they'll be able to do it pretty easily. Plus, maybe it's good for them to ask how this is different from `cy.log`? It's already confusing that `cy.log` logs to one place and not to another...

source/api/commands/task.md
chrisbreiding: Fair enough. I just didn't want some hand-wavy example with some pseudo-code for creating a user in a database or else some super complicated example that actually creates a user in a database.

source/api/commands/task.md
brian-mann: Uh then let's just make `cy.log` log out to `stdout` like... right now, lol...

arun290636:
@jennifer-shehane I have added below code to index.js:

var mysql = require('mysql');
module.exports = (on, config) => {
  // `on` is used to hook into various events Cypress emits
  // `config` is the resolved Cypress config
on('task', {
  'getDiscountCode': () => {
    var result;
    mysql.createConnection({
      host: "host",
      user: "user",
      password: "pass",
      database: "dbName"
    }).then(function(conn){
      // do stuff with conn
       result = conn.query('select query');
      conn.end();
  }); 
  debugger
  return result; 
  }
})
}

and spec file:
describe('connect to database', function () {
    it('database connection', function () {
      cy.visit('/')
      var rs = cy.task('getDiscountCode');
})
})

error on cypress - mysql.createConnection is not a function.

package.json:

"devDependencies": {
    "cypress": "2.0.1",
    "mysql": "^2.15.0"

Your help will be appreciated. Basically i want to connect to database and read some test data.

chrisbreiding:
@arun290636 I tried this out and the error is `mysql.createConnection().then` is not a function, because it doesn't return a promise. It [directly returns the connection](https://github.com/mysqljs/mysql#introduction).

brian-mann:
You wouldn't want to connect to the DB in the task as that will create endless amounts of connections for each time you run the task, which will effectively create a memory leak.

You want to connect to the DB once outside of the callback function so it only ever happens once and once only.

arun290636:
@chrisbreiding Thanks for your quick response. Database connection is working fine now but unable to return result from SQL query. is it easy to write output from SQL query to a JSON file in cypress? SAmple example will be helpful. Basically i want to read some test data from SQL database. It will be a just select query.

brian-mann:
@arun290636 because you're not using promises correctly. You're returning `result` which will always be undefined because it's being mutated asynchronously which is happening way after you've already returned `results`.

You need to return a promise if you're doing an async operation and then the results will be awaited and resolved with `cy.task`.

brian-mann:
@arun290636 unfortunately we can't provide 1:1 support for writing code. We're a small team and its not feasible to do this. We can only offer premium support to companies wishing for 1:1 support.

@chrisbreiding @jennifer-shehane `cy.task` is completely missing from the Plugins API in the docs. It only exists as a command in the API.

We also don't ever once explain about returning a promise and that it is awaited. You can crib the explanation from the Configuration API if necessary because I agree it is confusing.

https://docs.cypress.io/api/plugins/configuration-api.html#Usage

PR #453, opened 2018-03-08
Add best practice around setting `baseUrl`

### LINKED ISSUE #447: Explain Cypress GUI 'reloading' when 'baseUrl' not set
Adding a baseUrl to your cypress.json should fix the 'flashing' of initial load -> refresh -> run tests. Cypress essentially does not know before opening where you are planning to visit (unless specified in cypress.json), so it initially opens on localhost + a random port, then switches to the localhost + port you specified in your visit.

Cypress internally switches to the domain that you use in your cy.visit() so that your application’s code will run under the current domain at all times.

We had a lot of CORS related issues before when we did not set the main window to the same url as the application under test. Previously we implemented an endless amount of hacks and internal translations to figure out the domain you were supposed to be on without actually being on the domain. This caused code to behave different and caused subtle issues.

The entire proxy layer was written to handle https certificates, continue to inject (even on https pages), and know when to automatically bypass injection so you can open other tabs while testing in Cypress. This also allows us to unlock the ability to do things like whitelisting or blacklisting specific 3rd party domains, or even be able to stub not just XHR’s but any kind of HTTP request.

jennifer-shehane:
- close #447

### DIFF
--- a/source/guides/references/best-practices.md
+++ b/source/guides/references/best-practices.md
@@ -4,79 +4,6 @@ comments: false
 layout: toc-top
 ---
 
-<!-- ## Using your UI to Build Up State
-
-- use cy.request
-- seed the database
-- use stubbing
-- modify cookies
-- modify local storage
-- mock / stub methods
-
-Mostly WIP.
-
-Began {% url 'discussing here' testing-your-app#Testing-Strategies %}. -->
-
-<!-- ## Acting Too Much like a User
-
-WIP.
-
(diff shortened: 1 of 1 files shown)


brian-mann:
You're missing other key points of why someone would set a `baseUrl`.

I wrote them in my comment here: https://github.com/cypress-io/cypress-documentation/issues/447#issuecomment-371318583

Please add them :D

Issue #400, opened 2018-01-29
Typescript Support doc


jennifer-shehane:
Create a doc solely for 'TypeScript support' within the ['Guides'](https://github.com/cypress-io/cypress-documentation/tree/develop/source/guides/guides) section. We may consider having a separate section from 'Guides' called something like 'Tooling' which could include this document, Reporters, & Plugins.

If anyone wants to just write a basic markdown file of the documentation in a gist or PR, we can clean it up + add the link to the main docs after we have just the basic content.

brian-mann:
@jennifer-shehane I noticed that this had a PR merged but this issue was never closed... I don't know why not.

Anyway here is my feedback for this page: https://docs.cypress.io/guides/tooling/typescript-support.html#

I find this document very confusing for multiple reasons:

1. There is not a clear summary of what intellisense adds. I feel like this needs a list up front the benefits such as **assertion autocompletion**, amongst other features. That is one of the coolest things it can do, and we aren't even promoting it. 
2. The very first section suggests to add a directive. Then immediately below it, it suggests a `tsconfig.json` modification instead. It's not clear which one we are suggesting. It's also not clear what one does instead of another. It seems like editing `tsconfig.json` would make **all** files instantly work, without needing a directive. If this is the case, then obviously this should go first. The "one off" directive should be at the bottom as like a: "oh and you can do it as a 1-off this way, or if something isn't working in `tsconfig.json`. 
3. The animated gif is incredibly confusing. It starts strange... with the pointer not going in the direction I would think it should go - and its missing the actual cursor! Tooltips start flying in everywhere and I have **no idea** why it's doing this or even what this is communicating to me. The animated gif should be at the top - along with a better summary of what our typescript intellisense provides. We should show an actual autocomplete assertion, as this is one of the biggest and best reasons to use it.
4. Why is this issue still open? https://github.com/cypress-io/cypress/issues/1236 Have we fixed this? Does this document suffice? Are there other changes we need to make in Cypress core? If we're going to work to fix an issue, let's actually tie up all the knots surrounding it.
5. We mention Typescript in two places now - Guides - and in Plugins here: https://docs.cypress.io/plugins I feel that we should point Plugins to this guide, and then inside of this Guide link to the source code. As it stands, its fractured and makes no sense. 
6. I'm not sure this should be called "Typescript Support", maybe just "Typescript". You could argue that virtually every guide could be suffixed with Support. Reporter Support. Plugin Support. 
[...]

PR #466, opened 2018-03-16
WIP - TypeScript Support Doc

jennifer-shehane:
- addresses #400


brian-mann:
What's the rationale for plugin support being in tooling?

jennifer-shehane:
I guess I thought tooling would be things that are extensions of cypress, but not **really** specific to the core essential 'run e2e tests' - more specific to your personal workflow and involving 1 other library/third parties code, so:

- using cucumber with cypress
- how to get cypress working in my IDE
- how to have my babel config
- set up a different mocha reporter
- typescript intellisense

brian-mann:
Some of that makes sense... I guess some may have a corresponding Recipe too, which is also fine.

So Tooling is like... hey... so you wanna add stuff to Cypress to make it better? Maybe stuff for your IDE, or business processes (cucumber), or maybe how you interface with your own node code... here's a list of things.

I'm good with it 👍

source/guides/tooling/typescript-support.md
```diff
@@ -0,0 +1,34 @@
...
+## tsconfig
+
+If your project is written in TypeScript, you may find it more convenient to include Cypress in your `tsconfig.json`.
+
+```json
+{
+  "include": [
+    "integration/*.ts",
+    "support/*.ts",
+    "../node_modules/cypress"
```
NicholasBoll: I think this JSON block should be:

If you write your tests in TypeScript, you may find it more convenient to add a `tsconfig.json` inside your Cypress folder.

```json
{
  "compilerOptions": {
	"strict": true,
    "baseUrl": "../node_modules",
    "target": "es5",
    "lib": ["es5", "dom"],
    "types": ["cypress"],
  },
  "include": [
    "**/*.ts"
  ]
}
```

The `"types"` will tell the TypeScript compiler to only include type definitions from Cypress. This will address instances where the project also uses `@types/chai` or `@types/jquery`. Since chai and jQuery are namespaces (globals), incompatible versions will cause the package manager (yarn or npm) nest and include multiple definitions and cause conflicts.

source/guides/tooling/typescript-support.md
```diff
@@ -0,0 +1,34 @@
...
+comments: false
+---
+
+Cypress ships with {% url "official type declarations" https://github.com/cypress-io/cypress/tree/develop/cli/types %} for {% url "TypeScript" https://www.typescriptlang.org/ %}.
+
+# Intellisense
+
+In order to get intellisense working, you can try a few approaches:
+
+## Triple-Slash Directives
```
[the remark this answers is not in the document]
jennifer-shehane: Good point. I will have to think about a good place to mention this elsewhere in the docs.

source/guides/tooling/typescript-support.md
jennifer-shehane: omg, thank you! I was struggling 😅 I think this has helped finally push it to a shippable state.

PR #483, opened 2018-03-29
Fix a configuration option typo

davidcelis:
The "Screenshots and Videos" guide mentions setting the `trashAssetsBeforeHeadlessRun` configuration, which caused some confusion when it didn't work… Then I clicked through to the configuration guide and noticed that the option is really `trashAssetsBeforeHeadlessRuns` plural. Oops!


jennifer-shehane:

Issue #501, opened 2018-04-09
Document longpress event


jennifer-shehane:
Add as an example to `trigger` doc from https://github.com/cypress-io/cypress/issues/1487

```js
 cy.get(`.target`).trigger('mousedown')
 cy.wait(1000)
 cy.get(`.target`).trigger('mouseleave')
```

marzann:
Is this issue available to claim as a first-time contribution?

jennifer-shehane:
@marzann Yes! File to update: https://github.com/cypress-io/cypress-documentation/edit/develop/source/api/commands/trigger.md

marzann:
Great, I will work on it. Thank you.

PR #507, opened 2018-04-10
Added long-press event example to trigger.md

marzann:
Issue #501. Documented long-press event by adding an example to cypress-documentation/source/api/commands/_trigger.md.


marzann:
Thank you for this first-time opportunity. Hopefully I can help out again.

PR #530, opened 2018-04-16
Include searchable keywords 'nth', 'row'

### LINKED ISSUE #522: add more searchable keywords to .eq doc
- `nth`
- `row` - like...I want to assert the 2nd row contains this text...

ValerieThoma:
Add searchable keywords nth and row to .eq doc

### DIFF
--- a/source/api/commands/eq.md
+++ b/source/api/commands/eq.md
@@ -76,26 +76,6 @@ Option | Default | Description
 ```javascript
 cy.get('li').eq(1).should('contain', 'siamese') // true
 ```
-
-## Index Form End
-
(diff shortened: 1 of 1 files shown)


ValerieThoma:
This is just a first draft. 
I'm pretty sure the `nth child` stuff belongs in the info box up top. 
However, I'm not very confident about the `row` stuff. 
Like, does this example belong here? Should it be titled something else? Should the descriptor be different, i.e
* "Find text from the 2nd row of a table"
* "Make an assertion on the 2nd row of a table"
* "Assert the 2nd row of a table contains my text"
Does this even satisfy issue 522?

PR #534, opened 2018-04-17
add jQuery UI Sortable WIP

### LINKED ISSUE #518: Add 'drag and drop' example for jQuery UI Sortable to .trigger doc
https://github.com/cypress-io/cypress/issues/1542#issuecomment-380831700

```js
cy.get(".grip").first()
  .trigger('mousedown', { which: 1, pageX: 600, pageY: 100 })
  .trigger('mousemove', { which: 1, pageX: 600, pageY: 600 })
  .trigger('mouseup');
```

ValerieThoma:
Looking for best place to land jQuery UI sortable example

The only place I can see in the `.tirgger` docs to add this is under the existing Drag and Drop header, however, we link to a recipe and I'm worried about uniformity in the doc. 
Would it be out of place to throw in an example here or would it be okay?

### DIFF
--- a/source/api/commands/trigger.md
+++ b/source/api/commands/trigger.md
@@ -95,6 +95,8 @@ cy.get('.target').trigger('mouseleave')
 {% url 'Check out our example recipe triggering mouse and drag events to test dragging and dropping' recipes#Drag-and-Drop %}
 {% endnote %}
 
+#### jQuery UI Sortable ??
+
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
Yeah, I'd just put a header at the same level as Drag and Drop / Longpress that's just called 'jQuery UI Sortable' and put the example there.

PR #538, opened 2018-04-18
add suggested folders/files to add to .gitignore

### LINKED ISSUE #519: document what files should be added to .gitignore, or source control in general
Maybe here: https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests.html#Folder-Structure

`your screenshots and videos folders are worth ignoring`

ValerieThoma:
Played around with the language for a while. Does this sound logical and consistent?


jennifer-shehane:
Yeah, I might wrap this in an "info box", since it's kind of outside of the flow of the document, but a little side note that doesn't really belong anywhere else in the docs. How to use hexo tags for notes are...semi-documented here: https://github.com/cypress-io/cypress-documentation/blob/develop/lib/tags/note.js

But it'd look something like this:

```md
{% note info "What files should I add to my `.gitignore`?" %}
Cypress will create a...rest of text here...
{% endnote %}
```

ValerieThoma:
Absolutely, it definitely feels like an aside. Perfect!

ValerieThoma:
I overlooked the Question header. 
>"What files should I add to my `.gitignore`?"

That is an eye catcher and also checks off a, perhaps often, searched phrase. I will add this now.

PR #540, opened 2018-04-19
add global-hook example

ValerieThoma:
png example of adding a beforeEach in a cypress/support file


source/guides/core-concepts/writing-and-organizing-tests.md
```diff
@@ -77,10 +77,27 @@ By default Cypress will automatically include the plugins file `cypress/plugins/
 
 ## Support file
 
-By default Cypress will automatically include the support file `cypress/support/index.js` **before** every single spec file it runs. We do this purely as a convenience mechanism so you don't have to import this file in every single one of your spec files.
+By default Cypress will automatically include the support file `cypress/support/index.js`. This file runs **before** every single spec file . We do this purely as a convenience mechanism so you don't have to import this file in every single one of your spec files.
```
ValerieThoma: typo possibly?

source/guides/core-concepts/writing-and-organizing-tests.md
```diff
@@ -77,10 +77,27 @@ By default Cypress will automatically include the plugins file `cypress/plugins/
...
+})
+```
+![global hooks](/img/guides/global-hooks.png)
+
+{% note info %}
+**Note:** This example assumes you are already familiar with Mocha {% url 'hooks' writing-and-organizing-tests#Hooks %}. 
+{% endnote %}
+
+{% note danger%}
+{% fa fa-warning %} Keep in mind, setting something in global hook will render it less flexible for changes and for testing its behavior down the road.
```
ValerieThoma: Should we specifically address using `cy.route()` here? It's in our issue but I didn't know how to use it as an example without readers potentially feeling limited to only setting `cy.route()` there.

source/guides/core-concepts/writing-and-organizing-tests.md
jennifer-shehane: Yes. Perfect, that previous wording is definitely confusing

source/guides/core-concepts/writing-and-organizing-tests.md
jennifer-shehane: I'm gonna say no, because I've seen this question asked for various commands - they just aren't grasping the concept of where to put *anything* they want to run before every single test.

PR #543, opened 2018-04-21
Update find.md

jkassis:

### DIFF
--- a/source/api/commands/find.md
+++ b/source/api/commands/find.md
@@ -31,6 +31,14 @@ cy.find('.progress')          // Errors, cannot be chained off 'cy'
 cy.exec('node start').find()  // Errors, 'exec' does not yield DOM element
 ```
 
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
Hi, could you give some context on this addition to the doc?

jkassis:
Hi Jennifer,

It simply clarifies what what the API cannot do in this case and hopefully saves the developer some time. Sometimes anti patterns are just as useful references as regular patterns.

jennifer-shehane:
I feel that this example has lost some context with it's lack of `html`. 

- Why is it that the `header` yielded correctly and the `footer` did not?

Also, setting variables in this way is detailed at length as an anti-pattern in [this document](https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html#). I'm not sure that it's specifically relevant to `.find()` user itself.

PR #560, opened 2018-04-29
Update Installing-Cypress.md - intall local

tlemke11:
You have to mkdir node_modules before you run the npm cypress install --save-dev if you are using a non-node based project. I ran into this issue as a noob to npm.

### DIFF
--- a/source/guides/getting-started/installing-cypress.md
+++ b/source/guides/getting-started/installing-cypress.md
@@ -34,7 +34,8 @@ cd /your/project/path
 npm install cypress --save-dev
 ```
 
-This will install Cypress locally as a dev dependency for your project.
+This will install Cypress locally as a dev dependency for your project. 
+Note: Make sure that you have a node_modules folder in your project path first, otherwise cypress will install globally.
 
 {% img /img/guides/installing-cli.gif %}
 


jennifer-shehane:
I hadn't thought about this before. This...likely explains a few situations I've seen of people having it globally installed and not realizing it. Thanks for pointing this out. 

The recommended approach to set up a project to be ready to install a node_module 'locally' as opposed to 'globally' would be to run `npm init` first - to initialize the project for node_modules.

I'll update the PR to reflect this. Thanks for the contribution!

jennifer-shehane:
@tlemke11 Could you please sign our CLA?

tlemke11:
I did. I think that adding the npm init line might be better but my original commit works as well. I hope to update more "noob" insights in the future!

jennifer-shehane:
@tlemke11 So, I've actually been looking into this further, and the behavior appears to be more complex than we originally thought. Running `npm install cypress --save-dev` *should* create a `node_modules` folder within the current directory if you already do not have one **unless** you have a `node_modules` folder or `package.json` file in any parent directory - in which case cypress will install **locally** there. 

From [npm docs](https://docs.npmjs.com/files/folders#more-information):
> When installing locally, npm first tries to find an appropriate prefix folder. This is so that npm install foo@1.2.3 will install to the sensible root of your package, even if you happen to have cded into some other folder.
>
> Starting at the $PWD, npm will walk up the folder tree checking for a folder that contains either a package.json file, or a node_modules folder. If such a thing is found, then that is treated as the effective “current directory” for the purpose of running npm commands. (This behavior is inspired by and similar to git’s .git-folder seeking logic when running git commands in a working dir.)
>
> If no package root is found, then the current folder is used.

Either way, the directory where `cypress` is installed should be printed to the console in a post-install message, something like:

```
> cypress@2.1.0 postinstall /Users/jennifer/Dev/Projects/my_project/node_modules/cypress
```

So, it's likely a good idea to check the printed postinstall message if you think something funky is going on. 

I am still unaware of a situation where cypress would install globally without passing the `-g` flag directly. 


**I will update this PR again to more accurately reflect `npm`s behavior on install**

Issue #553, opened 2018-04-26
Specify lack of current support for Saucelabs / Browserstack


jennifer-shehane:
- should update last line of 'Sauce Labs' section in General Questions
- should just add a question to FAQ asking about our support and say it does not currently work right now.

ValerieThoma:
These lines already exist in the 'Sauce Labs' section of General Questions:
>Cypress’s API is written to be completely compatible with Sauce Labs, even though our API is not Selenium based at all. We will be offering better integration with Sauce Labs in the future.

Should this just be expounded on? Should this be moved to the final lines? 
I don't think we discussed in detail _what_ should be added, only that it need be explicitly understood that Sauce Labs integration was not currently available. 

I have submitted a WIP PR that requires some feedback. 
Thank you!

ValerieThoma:
🙏perfected by @brian-mann, thank you!

PR #568, opened 2018-04-30
fix typo/rewrite Sauce Labs section to be explicit/add Q to using-cypress faq

ValerieThoma:
Remove article 'a' from Capybara section 
Move statements about Sauce Labs integration to the bottom, wrap in info box, and use explicit language that integration is not currently available.

This would address one half of issue #553, leaving adding a question to Using Cypress FAQ that would answer the question definitively and potentially link to this General Question and/or issue #310 

This request requires feedback. I will leave a comment on the issue requesting it.

### DIFF
--- a/source/faq/questions/general-questions-faq.md
+++ b/source/faq/questions/general-questions-faq.md
@@ -37,7 +37,7 @@ Cypress essentially replaces Karma because it does all of this already and much 
 
 ***Capybara***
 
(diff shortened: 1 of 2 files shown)


ValerieThoma:
add *Is Cypress compatible with Sauce Labs and BrowserStack?* to Using Cypress FAQ, link to General Questions FAQ explanation and issue 310

ValerieThoma:
Reword answer to be explicit about Cypress's compatibility with Webdriver. Remove ambiguous remark about "discussions". 
Remove mention of 'Selenium' from Sauce Labs paragraph in General Questions FAQ

PR #576, opened 2018-05-03
Update plugins.yml

kentcdodds:
Just adding a section to the plugins for custom commands so I can add the cypress-testing-library 👍


jennifer-shehane:
Updated base branch - we do all PRs to `develop` in our projects.

PR #598, opened 2018-05-16
add cypress-testrail-reporter to plugins

### LINKED ISSUE #591: add cypress-testrail-reporter to docs
[`cypress-testrail-reporter`](https://github.com/Vivify-Ideas/cypress-testrail-reporter)
Where to add?  Maybe plugins, not sure. _tbd_

ValerieThoma:
Custom reporter publishes Cypress run results to TestRails
author Milutin Savovic

### DIFF
--- a/source/_data/plugins.yml
+++ b/source/_data/plugins.yml
@@ -25,12 +25,6 @@
       link: https://github.com/cypress-io/cypress-webpack-preprocessor
       keywords: [webpack]
       
-    - name: cypress-testrail-reporter
-      description: Custom reporter for publishing Cypress results to a TestRail test run. 
-      link: https://github.com/Vivify-Ideas/cypress-testrail-reporter
-      keywords: [testrail, reporter]   
-
-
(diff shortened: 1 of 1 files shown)


source/_data/plugins.yml
```diff
@@ -24,6 +24,11 @@
       description: Watches and bundles your spec files via webpack.
       link: https://github.com/cypress-io/cypress-webpack-preprocessor
       keywords: [webpack]
+      
+    - name: cypress-testrail-reporter
```
jennifer-shehane: Placing this here displays the `cypress-testrail-reporter` as under the `Preprocessors`, which isn't technically accurate. Maybe it should go under Development Tools???? It's a bit more of a category for "Reporting" than anything - but we don't have that category, should we make that category??

ValerieThoma: Okay. 🤔I guess that would/could be it's own category. Would it be okay hanging out by itself for a little while?

PR #642, opened 2018-05-30
cache home .npm folder on ci

bahmutov:
On Circle, because we use `npm ci` which blows away local `node_modules` it makes no sense to cache it. Instead we need to cache `~/.npm` folder


jennifer-shehane:
This ended up being addressed in a separate PR, I believe.

PR #650, opened 2018-05-30
move example from wait doc to route doc

### LINKED ISSUE #644: Add note in Route documentation on how to handle repeated requests
A common situation is needing to test multiple requests to the same route that have different responses. (For instance, if you get an error back on the first request, retry and then handle a successful response appropriately.)

This can be done in Cypress, but the documentation for it is under the `wait` command — it would be great to have this covered in the `route` documentation as well (a short note and a link to the example in wait would be enough I think).

Thanks!

ValerieThoma:
add note linking to wait command example of testing multiple requests from the same route

### DIFF
--- a/source/api/commands/route.md
+++ b/source/api/commands/route.md
@@ -233,6 +233,10 @@ cy.server()
 cy.route('DELETE', '**/users/*', {})
 ```
 
+{% note info Making multiple requests to the same route %}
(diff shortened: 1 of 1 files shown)


ValerieThoma:
Right, that's the example I link to. 
Okay. I will add an example comparable to it in the cy.route() doc in place of the aside.

ValerieThoma:
😂I just plagiarized you

PR #652, opened 2018-05-30
Update link in requests section

jdukelee:
Update the 'Requests' section regarding Cypress default behavior of ignoring requests that fetch static files. Change link to direct to documentation for cy.server() options.whitelist.


jdukelee:
Awesome! I'm really digging Cypress so thank you for everything you all have done!

PR #679, opened 2018-06-05
update ci docs

Bkucera:
need node version 10 on travis and circle v1 for the `npm ci` command


jennifer-shehane:
I see the `cypress/base:10` exists now too, should that be updated in the config examples from `cypress/base:8`? I haven't used it, so I'm not sure it's ready.

Bkucera:
add info about flushing cache in CI after caching `~/.cache`

PR #697, opened 2018-06-11
update link to hexo format

### LINKED ISSUE #696: fix broken link in server docs
screen shot 2018-06-11 at 10 47 01 am

>can change this option permanently 

this is ☝🏻styled as a link but this link goes nowhere

I believe it should go [here](https://docs.cypress.io/api/cypress-api/cypress-server.html#Options)

ValerieThoma:
I think this link was left over from the old docs, I updated it to use the hexo formatting, but I only _suspect_ that I linked the proper doc. Please let me know if otherwise.

### DIFF
--- a/source/api/commands/server.md
+++ b/source/api/commands/server.md
@@ -195,7 +195,7 @@ cy.server({
 })
 ```
 
-If you would like to change the default option for **ALL** `cy.server()` you [can change this option permanently](#permanently-override-default-server-options).
+If you would like to change the default option for **ALL** `cy.server()` you {%url 'can change this option permanently' cypress-server#Options %}.
 
 ***Turn off the server after you've started it***
 


jennifer-shehane:
K, so, you should always be able to test if you wrote the right link in the url helper, you can just prepend `http://on.cypress.io/` to the url passed in - so if http://on.cypress.io/cypress-server#Options link works, then you are good to go.

ValerieThoma:
@jennifer-shehane That's good to know, however, I didn't have any problems formatting this url I just wasn't positive I _linked_ it to the proper doc. The old link went nowhere, it looked left over from [the way you used to have your docs](https://www.cypress.io/blog/2017/07/05/10-ways-we-improved-our-docs/).

PR #754, opened 2018-07-02
Add cypress on rails gem to plugins list

grantspeelman:
Request to add a ruby gem that makes cypress testing easier on Ruby on Rails application


grantspeelman:
@jennifer-shehane /

grantspeelman:
Awesome, thank you too 😄

PR #804, opened 2018-07-27
Polish up some minor stuff like possessives

richdouglasevans:
Correcting `API's` to `APIs` plus a few smaller things in the same file.

### DIFF
--- a/source/guides/core-concepts/variables-and-aliases.md
+++ b/source/guides/core-concepts/variables-and-aliases.md
@@ -7,24 +7,24 @@ title: Variables and Aliases
 # {% fa fa-graduation-cap %} What You'll Learn
 
 - How to deal with async commands
-- What Aliases are, how they simplify your code
+- What Aliases are and how they simplify your code
 - Why you rarely need to use variables with Cypress
 - How to use Aliases for objects, elements and routes
 {% endnote %}
 
 # Return Values
 
-New users to Cypress may initially find it challenging to work with the asynchronous nature of our API's.
+New users to Cypress may initially find it challenging to work with the asynchronous nature of our APIs.
 
 {% note success 'Do not worry!' %}
 There are many simple and easy ways to reference, compare and utilize the objects that Cypress commands yield you.
 
-Once you get the hang of async code - you'll realize you can do everything you could do synchronously, without your code doing any backflips.
+Once you get the hang of async code you'll realize you can do everything you could do synchronously, without your code doing any backflips.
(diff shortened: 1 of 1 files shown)


source/guides/core-concepts/variables-and-aliases.md
```diff
@@ -7,24 +7,24 @@ title: Variables and Aliases
...
-- What Aliases are, how they simplify your code
+- What Aliases are and how they simplify your code
 - Why you rarely need to use variables with Cypress
 - How to use Aliases for objects, elements and routes
 {% endnote %}
 
 # Return Values
 
-New users to Cypress may initially find it challenging to work with the asynchronous nature of our API's.
+New users to Cypress may initially find it challenging to work with the asynchronous nature of our APIs.
```
richdouglasevans: Corrected a handful of such in this file: APIs is the plural of API.

Some style-guides advocate for using an apostrophe with acronyms so the style with apostrophes isn't wrong. I changed it to APIs in this file _only_ to be consistent with the _Introduction To Cypress_ which uses APIs. However, there are other files in the documentation that use API's with the apostrophe: I can add 'em to this PR to get all of the documentation consistent.

> Totes cool if you also think that the style using the apostrophe is what you want.

source/guides/core-concepts/variables-and-aliases.md
```diff
@@ -7,24 +7,24 @@ title: Variables and Aliases
...
 
 {% note success 'Do not worry!' %}
 There are many simple and easy ways to reference, compare and utilize the objects that Cypress commands yield you.
 
-Once you get the hang of async code - you'll realize you can do everything you could do synchronously, without your code doing any backflips.
+Once you get the hang of async code you'll realize you can do everything you could do synchronously, without your code doing any backflips.
```
richdouglasevans: Corrected a few sentences where ` - ` is used to indicate a clause or pause (I think). The sentences read just fine if it is removed though.... so I removed it. Style thing? No biggie.

source/guides/core-concepts/variables-and-aliases.md
```diff
@@ -49,7 +49,7 @@ button.click()
 
 ## Closures
 
-To access what each Cypress command yields you - you simply use {% url `.then()` then %}.
+To access what each Cypress command yields you use {% url `.then()` then %}.
```
richdouglasevans: The double `you - you` was a tad jarring.

Issue #32, opened 2017-09-06
Docs: Improve 'Debugging'


jennifer-shehane:
@brian-mann commented on [Tue Jul 04 2017](https://github.com/cypress-io/cypress-monorepo/issues/259)

saschwarz:
It would be very helpful if you could add to this section: https://docs.cypress.io/guides/guides/debugging.html#Using on using `debug()`. Specifically how do I access the current subject in the console and test assertions against it?

richdouglasevans:
Legit suggestion there @saschwarz : I've opened a PR that hopefully addresses your suggestion.

richdouglasevans:
@jennifer-shehane What else can be done to improve the `Debugging` guide? I was surfing backwards through the open tickets looking at the oldest ones and thus happened on this one.

If you can suggest improvements then I'll crack 'em out and we can get this ticket closed ✅

PR #817, opened 2018-08-04
Document accessing the current subject when using .debug()

richdouglasevans:
Added in response to this suggestion:

https://github.com/cypress-io/cypress-documentation/issues/32#issuecomment-405958740

screen shot 2018-08-04 at 12 27 30


source/guides/guides/debugging.md
```diff
@@ -65,7 +65,11 @@ it("let's me debug like a fiend", function() {
 })
 ```
 
-Ah, that's better! And functionally equivalent. This is just another example of how Cypress seeks elegance and expressivity for the modern web tester. Fewer keystrokes, more power, don't break standard assumptions, and we all win.
+Ah, that's better! And functionally equivalent. The current subject that was yielded by the {% url `cy.get()` get %} is exposed as `subject` so that you can interact with it in the console.
+
+{% img /img/guides/debugging-subject.png Debugging Subject %}
```
richdouglasevans: I ran the image through `optipng` with the default settings so that it's already optimized for the web.

PR #821, opened 2018-08-07
Remove `length` expectation from documentation

CheungJ:
I think that it's a bit redundant to test the length of the array because the next expectation checks the contents of the whole array - which implicitly checks that there are three elements in the array.


jennifer-shehane:
Yeah, I do agree that this would be unnecessary in a real world test, but since there is no other mention of how to test 'how many things are in an array' in the rest of the intro guide, I believe this should be left in for instructional purposes. A lot of people don't realize how to check how many elements are inside an element when they start out.

Love to hear your suggestions and ideas though. 👍

PR #822, opened 2018-08-08
Consider adding chrome to the travis-ci example

paulfalgout:
By default travis comes with Chromium v61 which has some [significant performance issues](https://github.com/cypress-io/cypress/issues/1905#issuecomment-410953582) with Cypress v3.  Users may also not expect that the tests are running in Chromium or understand that chrome is an option.

Introducing this would add extra opinion to the example that isn't necessarily needed, but I think not mentioning this in some fashion will lead to headaches that could be avoided.
https://docs.travis-ci.com/user/chrome#selecting-a-chrome-version

### DIFF
--- a/source/guides/guides/continuous-integration.md
+++ b/source/guides/guides/continuous-integration.md
@@ -61,9 +61,13 @@ As of Cypress version 3.0, Cypress downloads its binary to the global system cac
 ### Example `.travis.yml` config file
 
 ```yaml
+dist: trusty
+sudo: required
 language: node_js
 node_js:
   - 10
+addons:
+  chrome: stable  
 cache:
   directories:
     - ~/.npm


Bkucera:
Since all the other example CI configs do not have chrome installed, it might confuse users why only one example involves installing chrome. Likely we should include CI configs using the docker images here https://github.com/cypress-io/cypress-docker-images/tree/master/browsers

paulfalgout:
To my knowledge travis comes with only chromium and not chrome where as circleci has chrome by default I believe, so it isn't necessary to specify.  But are you suggesting that if you're using CI, docker would be the only documented solution?

Bkucera:
Correct. For using Chrome, the only documented solution is using a Docker image or having Chrome preinstalled somehow- like on the a Jenkins machine

Bkucera:
This will be further the case when we release Firefox support...Docker images with both browsers pre-installed is the simplest solution.

paulfalgout:
Hmm perhaps on a new project. If you already have a significantly complicated travis run, docker does not necessarily simplify things.  Where-as if I want to run firefox in travis with my current config it is a single line in the yml

The issue as I see it is there was/is issues with electron running on CI and in other issues the solution offered was to try chrome, but with `-- browser chrome` for anyone running travis by default that gives you chromium and it has similar problems to electron, and unless you are really paying attention it is hard to notice that it falls back to chromium.

Bkucera:
@paulfalgout so you can get both chrome and firefox on travis using only travis.yml? In that case, perhaps this _is_ easier for users on travis

jennifer-shehane:
@Bkucera Any conclusion to this PR? I will be closing very soon if no response.

jennifer-shehane:
Closing due to inactivity.

Issue #828, opened 2018-08-10
Be Explicit About The Scope Of Cypress.env


richdouglasevans:
The documentation for [`Cypress.env()`](https://docs.cypress.io/api/cypress-api/env.html#Notes) states:

> We chose to use Cypress because changes to your environment variables take effect for the remainder of ALL tests.

Just clarify what is meant by _"ALL tests"_ in light of the breaking change to 3.0; from [the changelog](https://docs.cypress.io/guides/references/changelog.html#3-0-0) for that release:

> We now run each spec file in isolation from one another. This shouldn’t technically break anything, but if you have set up your tests in a way that requires the state of a previous spec file, this could potentially break your tests.

Raising this issue in response to [this comment](https://gitter.im/cypress-io/cypress?at=5b6dc386b3cbd23ffc23a074) on the Cypress Gitter:

> Shouldnt Cypress.env(k,v) persist between tests?
>
> i specs file1.js and file2.js, I run them both and file2.js cannot see file1.js environment

richdouglasevans:
@jennifer-shehane I'm happy to do this one if you approve it 👍

brian-mann:
The scope of this is now just the current spec file (since we exit the browser between specs).

richdouglasevans:
Yes.

The isolation doesn't look to be in effect when you're running the `__all` spec... the isolation _is_ in effect but this requires knowing about the `_all` spec to explain the behaviour.

> I know, you're not meant to run all of the specs in interactive mode... but the button is there so it gets used.

I have distinct spec files that set and read environment variables set by `Cypress.env`.

```js
// first.spec
it("can set an env variable using Cypress.env.", function () {
    cy.log('Setting "bar" to "Bwahaha" using Cypress.env.')
    Cypress.env('bar', 'Bwahaha')
    
    cy.wrap(Cypress.env(bar)).should("eq", 'Bwahaha')
});

it("can see an env variable set in the same spec file using Cypress.env.", () => {
    cy.wrap(Cypress.env('bar')).should("eq", 'Bwahaha')
});
```
```js
// later.spec
it("cannot see env variable set in another spec using Cypress.env.", function () {
    cy.log(Cypress.env('bar')) // undefined
    cy.log(Cypress.env())      // {}
    cy.wrap(Cypress.env('bar')).should('be.undefined')
});
```

Running in headless mode demonstrates the isolation nicely.

[...]

PR #829, opened 2018-08-10
[#159] Document watching

richdouglasevans:
Addresses #159 Add docs around file watching rules.

screen shot 2018-08-10 at 19 19 14

PR #831, opened 2018-08-12
[#828] Explicitly mention the scope of Cypress.[env|config]

richdouglasevans:
Addresses #828


source/api/cypress-api/config.md
```diff
@@ -9,6 +9,12 @@ title: Cypress.config
 {% url 'Read about configuration first.' configuration %}
 {% endnote %}
 
+{% note warning Scope %}
+Configuration set using `Cypress.config` _is only in scope for the current spec file._
+
+Cypress runs each spec file in isolation: the browser is exited between specs. Configuration changed in one spec won't be visible in other specs.
+{% endnote %}
```
richdouglasevans: I expended the scope of this ticket because the same scoping applies to `Cypress.config` too.

> Somewhere else I read a comment from Brian Mann about being thorough when updating documentation... since this is so closely related I've included it under the rationale of _being thorough._

I also tested this locally to ensure that the scoping behaviour is the same as `Cypress.env`.

source/api/cypress-api/config.md
```diff
@@ -117,12 +120,13 @@ Cypress.config() // => {defaultCommandTimeout: 10000, viewportHeight: 900, ...}
 
 # Notes
 
-**Why use `Cypress.config` instead of `cy.config`?**
+**Why is it `Cypress.config` and not `cy.config`?**
 
 As a rule of thumb anything you call from `Cypress` affects global state. Anything you call from `cy` affects local state.
 
-Methods on `cy` are local and specific to a single test. Side effects from `cy` methods are restored between each test. We chose to use `Cypress` because changes to your configuration options take effect for the remainder of **ALL** tests.
+Since the configuration added or changed by `Cypress.config` is only in scope for the current spec file, you'd think that it should be `cy.config` and not `Cypress.config`&hellip;and you'd be right. The fact that `Cypress.config` affects local state is an artifact of the API evolving over time: `Cypress.config` used to affect global state&mdash;configuration added in one test spec file was available in other specs&mdash;but the Cypress team wisely made each spec run in isolation in {% url `3.0.0` changelog#3-0-0 %} and by that time `Cypress.config` was public API.
```
richdouglasevans: Not totes sure about adding such a comment, but it's legit and it's squirrelled away in the _Notes_ section at the bottom so it's not _in your face when you're reading the documentation as reference.

> Any thoughts to adding a `cy.config` command and deprecating `Cypress.config`? (Same for `cy.env` and `Cypress.env`?)

richdouglasevans:
Cheers Jennifer. Lemme have one more pass through the docs to ensure that the language around `tests` and `specs` is consistent per Chris' comment in [another PR](https://github.com/cypress-io/cypress-documentation/pull/829#pullrequestreview-145659913).

> The language might fine: just wanna check.

PR #835, opened 2018-08-14
Update sample spec filename

developering:
The videos on this page use sample_spec.js as the filename, but the doc says simple_spec.js. I updated to sample so they're consistent.


jennifer-shehane:

developering:
@jennifer-shehane Just took care of that. Thanks!

PR #840, opened 2018-08-17
add links to CI repos, close #837

### LINKED ISSUE #837: Add links to CI example repos to the continuous integration page
We list the CI providers we have tried and know we support on https://docs.cypress.io/guides/guides/continuous-integration.html#

screen shot 2018-08-16 at 10 13 18 am

But we also have all the repos (including kitchensink that we run on multiple CIs), so we should link the example repo for each one - for example Codeship Pro (which should be split from Codeship Basic)

bahmutov:
- close #837 

Adds links to example repo and direct CI config file, so it looks like this

screen shot 2018-08-17 at 11 10 39 am


jennifer-shehane:
I'd like to pretty this up a little bit. Probably put this in a table.

Issue #46, opened 2017-09-07
menuspy not working precisely


jennifer-shehane:
@lorennorman commented on [Fri May 26 2017](https://github.com/cypress-io/cypress-monorepo/issues/99)

I've noticed a few things weird about menuspy:
- click a link on the right, page jumps to the right place, but the incorrect link is highlighted on the right
- sometimes the highlighted link on the right does not match the anchor tag in the URL

Example:
![cypress io documentation - error messages 2017-05-26 11-55-15](https://cloud.githubusercontent.com/assets/17697/26502795/374edfea-420c-11e7-981a-8d9c85abbb1f.png)

---

@jennifer-shehane commented on [Thu Jul 06 2017](https://github.com/cypress-io/cypress-monorepo/issues/99#issuecomment-313131839)

This issue was difficult to track down. Sometimes menuspy was accurate and sometimes not. 

After some use, it seems to be most off when images are included on the page, so we suspect that menuspy is not properly calculating the height of images into it's scrolling algorithm. This is likely because the images have not loaded yet. 

Can we calculate after the `onload` event? Does the `onload` event call before all images are loaded? Each image fires it's own `load` event, so you could check all images `load` event. 

To test this, throttle the network in devtools.

richdouglasevans:
Can this issue be `Closed` in light of the recent merge of #844 ?

jennifer-shehane:
Oh boy - yeah - this is old.

PR #844, opened 2018-08-19
[#46] Make MenuSpy work correctly on the Changelog page

richdouglasevans:
Addresses #46 

> The only remaining weirdness with menuspy is on the Changelog page. The highlighting of the righthand nav does not match the scrolling position AT ALL and requires some investigation.

Yeah. Upon investigation I found two things:

* the heading structure of the Changelog source was inconsistent.
* MenuSpy wasn't calculating the offset correctly.

I have changed:

* the Markdown source code for the Changelog to be at a consistent level.
* overridden a MenuSpy function so that offsets are calculated correctly.

This "works" in both of the browsers I've tested against: Chrome (latest `68.0.3440.106`) and Safari (`11.1.2`).

### DIFF
--- a/source/guides/references/changelog.md
+++ b/source/guides/references/changelog.md
@@ -2,7 +2,7 @@
 title: Changelog
 ---
 
-## 3.1.0
+# 3.1.0
 
 *Released 8/13/2018*
 
@@ -45,7 +45,7 @@ title: Changelog
 - {% url 'Update `Dashboard Service` doc with new screenshots and run sections.' dashboard-service %}
 - {% url 'Update `Command Line` doc with new flags and alphabetized flags.' command-line %}
 
-## 3.0.3
+# 3.0.3
 
 *Released 7/30/2018*
 
@@ -96,7 +96,7 @@ title: Changelog
 
 - {% url 'New "Module API" doc' module-api %}
 
-## 3.0.2
+# 3.0.2
 
 *Released 6/28/2018*
 
@@ -156,7 +156,7 @@ title: Changelog
 - {% url 'Added `Cypress.spec` doc.' spec %}
 - {% url 'Added `after:screenshot` plugin event doc.' after-screenshot-api %}
 
-## 3.0.1
+# 3.0.1
 
 *Released 5/30/2018*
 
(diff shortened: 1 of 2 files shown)


source/guides/references/changelog.md
```diff
@@ -2,7 +2,7 @@
 title: Changelog
 ---
 
-## 3.1.0
+# 3.1.0
```
richdouglasevans: I changed all of the version headings to be `#` (`h1`).

Previously all of the version headings starting with `1.0.0` were `##` (`h2`) which resulted in the MenuSpy code only detecting headings >= `1.0.0`. As you can see in the following screenshot, MenuSpy is only finding the 25 versions  >= `1.0.0`.

screen shot 2018-08-18 at 11 03 23

You can see this in action by [surfing to the current Changelog](https://docs.cypress.io/guides/references/changelog.html#1-0-0)... you can scroll down past `1.0.0` to older versions but the menu won't advance beyond `1.0.0` because that's the last entry in that collection.

After making the change MenuSpy now finds all of the versions.

screen shot 2018-08-18 at 11 08 58

You can now scroll past `1.0.0` and the menu will advance correctly.

My rationale for changing all the version headings to `#` is that each version is a top-level thing. Happy to change 'em  -- or see 'em changed -- to `##`: it's the consistency that's key to getting the MenuSpy behaviour.

themes/cypress/source/js/toc.js
```diff
@@ -9,6 +9,20 @@
...
+  MenuSpy.prototype.tick = function tick () {
+    var fromTop = this.currScrollTop + headerHeight;
+    var inViewElms = this.scrollItems.filter(function (item) {
+      return item.offset < fromTop;
+    });
+    var elementInView = (inViewElms.length > 0)
+      ? inViewElms.pop()
+      : this.scrollItems[0];
+    this.activateItem(elementInView)
+  };
```
richdouglasevans: Honestly, this feels like a band-aid solution 😳

The better (?) approach would be to address this at the source in the MenuSpy library. I did start down that route but my enthusiasm was swiftly extinguished when I discovered that [there are no tests of any significance](https://github.com/lcdsantos/menuspy/blob/master/test/basic.spec.js) in the MenuSpy source code.

> I'm not code shaming there: libraries sans tests are fine... I _personally_ can't be bothered investing my time into them though.

Issue #846, opened 2018-08-19
H3 headings do not highlight within sidebar's Table of Contents


jennifer-shehane:
We wanted H3 level headings to be linkable, but not have the H3 headings display in the Table of Contents in the sidebar (as it would be too cluttered). 

<img width="943" alt="screen shot 2018-08-19 at 11 48 16 am" src="https://user-images.githubusercontent.com/1271364/44310530-e44d4280-a3a5-11e8-885e-2d1673a8cdc9.png">

But, in implementing this, now the url hashtag updates when inside an H3, which does not match the selectors on the Table of Contents, so nothing highlights. 

Ideally, we would like the H3 heading's parent H2 referenced in the table of contents to be highlighted when on an H3 url hash.

PR #858, opened 2018-08-26
[#843] TOCs h2 is highlighted when in child h3

richdouglasevans:
Addresses #846

## Behaviour Before The Change

* we're in the `<h3>` section _We recommend users_
* the parent `<h2>` section is **not** highlighted in the sidebar

before

## Behaviour After The Change

* we're in the `<h3>` section _We recommend users_
* the parent `<h2>` section **is** highlighted in the sidebar

after

### DIFF
--- a/themes/cypress/layout/page.swig
+++ b/themes/cypress/layout/page.swig
@@ -21,7 +21,7 @@
           <aside id="article-toc" role="navigation">
             <div id="article-toc-inner">
               <strong class="sidebar-title">{{ __('page.contents') }}</strong>
-              {{ toc(page.content, {list_number: false}) }}
+              {{ toc(page.content, {list_number: false, max_depth: 2}) }}
               <a href="#" id="article-toc-top"><i class="fa fa-long-arrow-up"></i> {{ __('page.back-to-top') }}</a>
             </div>
           </aside>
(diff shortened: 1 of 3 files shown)


themes/cypress/layout/page.swig
```diff
@@ -21,7 +21,7 @@
           <aside id="article-toc" role="navigation">
             <div id="article-toc-inner">
               <strong class="sidebar-title">{{ __('page.contents') }}</strong>
-              {{ toc(page.content, {list_number: false}) }}
+              {{ toc(page.content, {list_number: false, max_depth: 2}) }}
```
richdouglasevans: By specifying a maximum depth value of `2`&mdash;where the `2` refers to the maximum header level&mdash;we can control what gets generated into the TOC.  In so doing we get the desired behaviour.

https://hexo.io/docs/helpers.html#toc

themes/cypress/source/css/_partial/toc.scss
```diff
@@ -79,10 +79,6 @@
   list-style: none;
 }
 
-.toc-item.toc-level-3 {
-  display: none;
-}
-
```
richdouglasevans: Because Hexo is now generating a TOC with a maximum depth of `2`, there aren't even any TOC item links with a level of 3 or higher. As such we no longer need to hide them because they're just not there in the generated markup.

PR #873, opened 2018-08-29
Fix malformed YAML in plugin data

richdouglasevans:
Fixes a wee issue introduced in the merge of #871 

screen shot 2018-08-29 at 10 09 54

### DIFF
--- a/source/_data/plugins.yml
+++ b/source/_data/plugins.yml
@@ -70,7 +70,7 @@
       keywords: [testing-library, capybara]
     
     - name: PickleJS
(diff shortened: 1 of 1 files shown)


source/_data/plugins.yml
```diff
@@ -70,7 +70,7 @@
       keywords: [testing-library, capybara]
     
     - name: PickleJS
-      description: An addition to the Cucumber plugin, featuring a collection of phrases you can use for common actions (ex: "I click on an <Element>", "I should see an <Element>")
+      description: 'An addition to the Cucumber plugin, featuring a collection of phrases you can use for common actions (ex: "I click on an <Element>", "I should see an <Element>")'
```
richdouglasevans: The presence of `ex: ` in the `description` was b0rking the build. I've explicitly quoted the string to address this.

PR #876, opened 2018-08-29
[#564] Tighten docs around force404 behaviour

### LINKED ISSUE #564: Unclear about Server doc?
[Here](https://github.com/cypress-io/cypress-documentation/blob/68c163347e1fc407dcaac534a6c59ab160a27006/source/api/commands/server.md#no-args) it says

> Any request that does not match a cy.route() will be sent a 404 status code

Is this true? It feels like `force404` is set to true with that sentence but doc says it's not.

richdouglasevans:
closes #564

> This change isn't as trivial as I thought originally 'cos there are a few places in the docs that have "incorrect" info. I'm scare quoting there because I'm now doubting my own understanding of the behaviour in light of the docs: I'm writing some (more) tests to prove the behaviour.

I verified the behaviour locally: routes that are _not_ matched will, by default, just pass through to the server.

screen shot 2018-08-29 at 10 38 54

### DIFF
--- a/source/api/commands/route.md
+++ b/source/api/commands/route.md
@@ -191,38 +191,14 @@ cy.route('POST', '**/comments', commentsResponse)
 
 Any request that matches the `method` and `url` of a route will be responded to based on the configuration of that route.
 
-If a request doesn't match any route, {% urlHash "it will automatically receive a 404" Notes %}. For example, given we have the following routes:
+{% note info %}
+If a request doesn't match any route then the behavior depends on the value of the `force404` option on the {% url `cy.server()` server %}:
 
-```javascript
-cy.server()
-cy.route('**/users', [
-  {id: 19, name: 'Laura'},
-  {id: 20, name: 'Jamie'}
-])
-cy.route('POST', '**/messages', {id: 123, message: 'Hi There!'})
-cy.get('form').submit()
-```
+* if `force404` is `false` (the default) then the request will {% url 'pass through to the server' network-requests#Don’t-Stub-Responses %}.
+* if `force404` is `true` then the response {% urlHash "will be a 404" Notes %}.
 
-```javascript
-// Application Code
-// when our form is submitted
-$('form').submit(() => {
-  // send an AJAX to: GET /users
-  $.get('https://localhost:7777/users' )
-
-  // send an AJAX to: POST /messages
-  $.post('https://localhost:7777/messages', {some: 'data'})
-
-  // send an AJAX to: GET /updates
-  $.get('https://localhost:7777/updates')
-})
-```
-
-**The above application code will issue 3 AJAX requests:**
-
-1. The `GET /users` will match our 1st route and respond with a `200` status code and the array of users.
-2. The `POST /messages` will match our 2nd route and respond with a `200` status code with the message object.
-3. The `GET /updates` did not match any routes and its response automatically sent back a `404` status code with an empty response body.
+You can {% url 'read more about this behavior here.' server#Options %}
+{% endnote %}
 
 ***Specify the method***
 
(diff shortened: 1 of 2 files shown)


source/api/commands/server.md
```diff
@@ -1,15 +1,13 @@
 ---
 title: server
-
 ---
 
-Start a server to begin routing responses to `cy.route()` and `cy.request()`.
+Start a server to begin routing responses to {% url `cy.route()` route %} and {% url `cy.request()` request %}.
```
richdouglasevans: Hyperlinking the first mention of these other commands in order to be helpful.

source/api/commands/server.md
```diff
@@ -67,7 +65,7 @@ Option | Default | Description
 
 ***After starting a server:***
 
-- Any request that does not match a {% url `cy.route()` route %} will be sent a `404` status code.
+- Any request that does **NOT** match a {% url `cy.route()` route %} will {% url 'pass through to the server' network-requests#Don’t-Stub-Responses %}.
```
richdouglasevans: The meat of this PR:

* document the correct behaviour
* hyperlink to the relevant section of the _Network Requests_ section to be helpful.

source/api/commands/server.md
```diff
@@ -53,7 +51,7 @@ Option | Default | Description
 Option | Default | Description
 --- | --- | ---
 `enable` | `true` | pass `false` to disable existing route stubs
-`force404` | `false` | forcibly send XHR's a 404 status when the XHR's do not match any existing
+`force404` | `false` | forcibly send XHR's a 404 status when the XHR's do not match any existing route
```
richdouglasevans: I thought the sentence ended abruptly so I added a trailing _route_ to make it read smoother.

source/api/commands/route.md
```diff
@@ -191,38 +191,14 @@ cy.route('POST', '**/comments', commentsResponse)
...
 
-**The above application code will issue 3 AJAX requests:**
+* if `force404` is `false` (the default) then the request will {% url 'pass through to the server' network-requests#Don’t-Stub-Responses %}.
+* if `force404` is `true` then the response {% urlHash "will be a 404" Notes %}.
 
-1. The `GET /users` will match our 1st route and respond with a `200` status code and the array of users.
-2. The `POST /messages` will match our 2nd route and respond with a `200` status code with the message object.
-3. The `GET /updates` did not match any routes and its response automatically sent back a `404` status code with an empty response body.
+You can {% url 'read more about this behavior here.' server#Options %}
+{% endnote %}
```
richdouglasevans: > Right, so this is rather a large edit, possibly the largest I've made so far. Do steer me Jennifer.

I've removed this example in favour of a short block that directs the reader to the documentation about `force404` on the `server` page. It's a good example but it essentially duplicates a much shorter example on the `server` page. I've attempted to preserve the thrust of this section while also directing the reader to the `server` page because the `force404` option is applied at the level of the `server` rather than the `route`.

Yeah? 🤔

source/api/commands/route.md
```diff
@@ -461,16 +437,10 @@ The following XHR's which were `xhr.open(...)` with these URLs would:
...
-Status | Body | Headers
---- | --- | ---
-`404` | "" | `null`
-
-If you'd like to enable this behavior you need to pass:
+You can force requests that do _not_ match a route to return a `404` status and an empty body by passing an option to the `cy.server()` like so:
 
 ```javascript
-cy.server({force404: true})
+cy.server({ force404: true })
```
richdouglasevans: Again, I'm directing the reader to the "canonical" documentation about the `force404` option on the `server` page.

source/api/commands/server.md
```diff
@@ -103,9 +101,9 @@ Adding delay can help simulate real world network latency. Normally stubbed resp
 cy.server({delay: 1500})
 ```
 
-***Prevent sending 404's to unmatched requests***
+***Send 404s on unmatched requests***
```
richdouglasevans: This header logic was the wrong way round: the `force404` option doesn't _prevent_; rather, it enables the sending of a `404` on unmatched requests so I've changed the header to reflect that.

source/api/commands/route.md
jennifer-shehane: I feel like we have lost the info about the `headers` being `null` - I haven't verified this is what truly does happen though, so I'm trusting that it was documented correctly to begin with. 😅

source/api/commands/route.md
jennifer-shehane: This is fine and a good use of a 'note' as it is outside of the flow of the content.

source/api/commands/server.md
jennifer-shehane: 👍

PR #877, opened 2018-08-29
[#217] Emphasize that GET is the default method for cy.route

### LINKED ISSUE #217: Emphasize default method for cy.route() is GET
https://github.com/cypress-io/cypress/issues/567
> Just a side-mention that the failure to match POST should probably be included in the intro-level documentation as a note/gotcha. As a Cypress newbie I got bitten by the same problem--my XHR was using POST--and only after stumbling upon this GitHub issue did I learn the default was GET. (Better, perhaps, would be that if you don't specify, it would match either).

richdouglasevans:
closes #217 

screen shot 2018-08-29 at 11 04 04
screen shot 2018-08-29 at 11 03 50

### DIFF
--- a/source/api/commands/route.md
+++ b/source/api/commands/route.md
@@ -39,7 +39,11 @@ Supply a response `body` to *stub* in the matching route.
 
 **{% fa fa-angle-right %} method** ***(String)***
 
-Match the route to a specific method (`GET`, `POST`, `PUT`, etc). If no method is defined, Cypress will match `GET` requests by default.
+Match the route to a specific method (`GET`, `POST`, `PUT`, etc).
+
+{% note bolt %}
+If no method is defined Cypress will match `GET` requests by default.
+{% endnote %}
 
 **{% fa fa-angle-right %} callbackFn** ***(Function)***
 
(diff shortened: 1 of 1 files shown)


source/api/commands/route.md
```diff
@@ -39,7 +39,11 @@ Supply a response `body` to *stub* in the matching route.
 
 **{% fa fa-angle-right %} method** ***(String)***
 
-Match the route to a specific method (`GET`, `POST`, `PUT`, etc). If no method is defined, Cypress will match `GET` requests by default.
+Match the route to a specific method (`GET`, `POST`, `PUT`, etc).
+
+{% note bolt %}
```
richdouglasevans: 😆 I just learned about `bolt` yesterday... using it plenty now 😄

source/api/commands/route.md
```diff
@@ -191,6 +195,10 @@ cy.route('POST', '**/comments', commentsResponse)
 
 Any request that matches the `method` and `url` of a route will be responded to based on the configuration of that route.
 
+{% note bolt %}
+`GET` is the default HTTP method used to match routes. If you want to stub a route with another HTTP method such as `POST` then you {% urlHash 'must be explicit about the method' Arguments %}.
+{% endnote %}
+
 If a request doesn't match any route, {% urlHash "it will automatically receive a 404" Notes %}. For example, given we have the following routes:
```
richdouglasevans: Well... darn 😢 

Only just spotted this line about _"automatically [receiving] a 404"_... I'd better go back and amend my earlier PR #876

source/api/commands/route.md
jennifer-shehane: Yeah, it's a newer addition to notes, but pretty useful here.

Issue #861, opened 2018-08-27
Convert ***headings*** into ### headings


jennifer-shehane:
Should only be converted if ***heading*** is within a ## heading like:

```
## Heading

### Heading 
```

Otherwise the Table of Contents does not work properly when:

```
# Heading

### Heading
```

jennifer-shehane:
Closing as stale

PR #886, opened 2018-08-31
Display sections that were hidden because of missing quote

richdouglasevans:
A missing quote was causing a couple of sections to not be generated.

## Before

There's supposed to be a couple of sections here before the _Test Failures_ section... they're missing.

before

## After

The sections are now present in the generated documentation.

after

### DIFF
--- a/source/guides/core-concepts/dashboard-service.md
+++ b/source/guides/core-concepts/dashboard-service.md
@@ -218,7 +218,7 @@ Standard output includes details and summaries of your tests for each spec file 
 
 You will also see a summary at the bottom indicating the screenshots, or videos that were uploaded during the recording.
 
-{% img /img/dashboard/standard-output-of-recorded-test-run.png "standard output %}
+{% img /img/dashboard/standard-output-of-recorded-test-run.png "standard output" %}
 
 ***{% fa fa-picture-o fa-fw %} Screenshots***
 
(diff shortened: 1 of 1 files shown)


source/guides/core-concepts/dashboard-service.md
```diff
@@ -218,7 +218,7 @@ Standard output includes details and summaries of your tests for each spec file
 
 You will also see a summary at the bottom indicating the screenshots, or videos that were uploaded during the recording.
 
-{% img /img/dashboard/standard-output-of-recorded-test-run.png "standard output %}
+{% img /img/dashboard/standard-output-of-recorded-test-run.png "standard output" %}
```
richdouglasevans: So yeah, the absence of that trailing `"` caused these sections to not be generated into the HTML.

I don't know why and I haven't investigated to determine the root cause: I know, I'm being lazy... yeah, I just gotta own that 😳Sorz... kinda; the sun is shining outside and I'm about to head out for a run so I'm not really that sorry 😄

> I only stumbled onto this by accident because I was idly messing about with Pandoc filters to see if they'd help with [this issue](https://github.com/cypress-io/cypress-documentation/issues/861) about converting header sections -- I don't think a Pandoc filter will help with that issue -- and my filter caused the Hexo build to fail because these sections started being picked up.

source/guides/core-concepts/dashboard-service.md
```diff
@@ -228,7 +228,7 @@ All screenshots taken during the test run can be found in the **Screenshots** of
 
 The video recorded during the test run can be found under the **Video** of the spec. You can also download the video.
 
-{% url /img/dashboard/videos-of-recorded-test-run.png "Video of test runs" %}
+{% img /img/dashboard/videos-of-recorded-test-run.png "Video of test runs" %}
```
richdouglasevans: Once the trailing quite was added the sections started being processed and this highlighted this wee issue with the wrong tag being used.

Simple to fix and then voila, all that sweet documentation started appearing 🎉 👍

source/guides/core-concepts/dashboard-service.md
jennifer-shehane: Yes, I've seen this before. Thanks for noticing. 👍

Issue #872, opened 2018-08-29
Feature Request: High Level Architecture


RichCrook:
I gave a well received talk about Cypress and have worked with it a bit, read the docs but are there any good sources for the architecture of Cypress? I know it's an Electron application but how specifically replace Selenium? It does not use Puppet I think...

If I'm asking folks to move away from Selenium, how is it actually implemented? I'd love to be able to give specifics with diagrams too!

Gleb Bahmutov not really, I think, but open an issue in cypress-documentation repo, that would be useful, even for high level architecture explanation

=========

Thanks,
Rich Crook

richdouglasevans:
Spot on Rich!

I've been wanting this very same thing and have been messily doodling the emerging architecture as I've been reading the Cypress source. I don't have anything close to a polished set of diagrams yet but when I do I'll tag this ticket in the PR.

jennifer-shehane:
Yeah there was this issue for this: https://github.com/cypress-io/cypress-documentation/issues/520

But, will close that one in favor of this.

jennifer-shehane:
I made a really sad wireframe diagram that I never finalized back then 😢 

![screen shot 2018-08-29 at 11 33 40 am](https://user-images.githubusercontent.com/1271364/44798432-795af300-ab7f-11e8-916b-aa77ba6d602b.png)

richdouglasevans:
I'll pick this one up then 🚀

> It'll be a couple of weeks at least since we've got a lot of work on at the mo'... so no PRs from me for a bit.
>
> Also, if Cypress is about to drop a release with a whole new-and-improved architecture lemme know before I scribble my way through this box of wax crayons 😆

![](https://media.giphy.com/media/zaezT79s3Ng7C/giphy.gif)

jennifer-shehane:
No, no large architecture changes for a while.

tlqiao:
@richdouglasevans , Is there any new update for high level architecture of cypress?

richdouglasevans:
Yep, it's a work in progress... been doing it off and on as time permits. (Yeah, been busy with work.) I plan to complete it and post it up for review over the Xmas holidays when I'll have some free-er time.

So, this year for sure 😄

TechGuyTy:
[...]

Issue #353, opened 2018-01-02
Linking to FAQ 'href' does not work within {% url %} tag


jennifer-shehane:
The link below throws an error during the build that the id cannot be found:

```
{% url "Open Source" general-questions-faq#Is-Cypress-free-and-open-source %} 
```

If you write the link like below (thinking, maybe the icon is needed) it does not throw an error, but also the link does not go to the correct href

```
{% url "Open Source" general-questions-faq#fa-fa-angle-right-Is-Cypress-free-and-open-source %} 
```

richdouglasevans:
Hi @jennifer-shehane 

I looked into this because I wanted to link to something in the FAQ and encountered this very issue.

The root cause is a sequencing one. The build does some eager validation of links to ensure that they are valid: in the case of validating _local_ links -- i.e. links to the Cypress documentation itself -- the validation check eagerly renders the source markdown and tries to find the link in the rendered HTML.

Unfortunately, this "simple" rendering of the source markdown doesn't account for the fact that the actual final rendered Cypress documentation will itself be transformed during the build process: so, it's looking at an unprocessed version of the HTML _but not the final HTML that's ultimately generated._ One of the helper tags -- `icon` -- transforms the HTML and this helper tag is not fired when the "simple" rendering happens.

As such, the validation fails. If the validation happened right at the end of the process using the final output of the build process then the validation would pass. (Assuming nothing was actually invalid.)

IMO the validation is happening at the wrong time. It's cool to fail fast but the validation should not be happening at the same time that the HTML is being generated 'cos that HTML may still be (and is) in the process of being transformed (mutated). Better to defer validation until the final HTML has been generated (and is immutable).

Since this is documentation about Cypress, I played around with doing the validation of the links using Cypress itself. It's pretty easy to do but the validation process does significantly increase the runtime of the Cypress test suite for the documentation. This is balanced by the fact that the build process is significantly faster because there's no validation happening at that time: the inline validation process means rendering Markdown to HTML repeatedly so being able to skip that is nice when you're repeatedly iterating on the documentation as I have been doing lately.

Certainly I could submit a PR to move the validation to the Cypress test suite and remove the validation during the build. There's no loss in the quality of the Cypress documentation 'cos links would still be being checked for validity (assuming one runs the Cypress test suite). 

Thoughts?

jennifer-shehane:
[...]

Issue #197, opened 2017-11-01
404 errors in build


jennifer-shehane:
- Find and fix these

<img width="1138" alt="screen shot 2017-11-01 at 12 30 48 pm" src="https://user-images.githubusercontent.com/1271364/32285471-c0178da6-bf00-11e7-8463-cbeb0a751b5f.png">

jennifer-shehane:
I get these socket hang ups sometimes too

<img width="803" alt="screen shot 2017-11-01 at 12 38 19 pm" src="https://user-images.githubusercontent.com/1271364/32285735-9bdfbf98-bf01-11e7-8dac-70857993be97.png">

jennifer-shehane:
Via @brian-mann explanation:

> those are failing because in node when you make gazillions of network requests it has to do with socket pooling and it can time out

> i can’t really explain it but i’ve read baotu it

> we make thousands of network requests

PR #889, opened 2018-08-31
[#663] Generate links even when validation fails

### LINKED ISSUE #663: errant punctuation breaking hexo url links
misplaced period breaks hexo links to npm modules

faq

richdouglasevans:
Closes #663 

Right, so this is a _pragmatic_ tactical fix for an issue that's manifest on the Cypress documentation site right now: "blank" hyperlinks leading to malformed sentences.

## Examples

* https://docs.cypress.io/guides/getting-started/installing-cypress.html#Continuous-Integration

Notice how the link to the `npx` module on NPM is missing from the generated text.

npx-missing

* https://docs.cypress.io/faq/questions/using-cypress-faq.html#How-do-I-wait-for-my-application-to-load

The following example is the one originally reported by Valerie in #663 

Notice how the link to the `wait-on` module on NPM is missing from the generated text.

wait-on-missing

## After The Fix

With the change from this PR in place the above docs are generated correctly.

screen shot 2018-08-31 at 15 40 11
screen shot 2018-08-31 at 15 40 24

## Root Cause

This was spotted in #663 and attributed to errant punctuation: however, the root cause is actually that the site generation process can (and does) result in "blank" content being generated during the validation of external links.

If the validation of the URL results in an error -- HTTP `500`, `404`, etc. -- then... nothing is generated which manifests as "blank" hyperlinks leading to malformed sentences.

## Tactical Fix

This PR contains a tactical fix: generate the hyperlink even in the face of an error. (The error continues to be logged to the error stream.)

This addresses the issue that's currently live on the published Cypress.io site: namely, "blank" hyperlinks leading to malformed sentences. The current approach to _external_ link validation has issues -- #197 #353 -- which'll only be addressed properly in #197.

> I do plan to address those issues with a strategic fix after I've done the much more interesting work of contributing some architecture diagrams for #872 

In the interim this tactical fix addresses a real issue on the website. Totes open to this being bounced back but I hope that my explanation communicates _why_ I've submitted this PR.

🤗

### DIFF
--- a/source/faq/questions/using-cypress-faq.md
+++ b/source/faq/questions/using-cypress-faq.md
@@ -147,7 +147,10 @@ When you load your application using `cy.visit()`, Cypress will wait for the `lo
 
 **_In CI, how do I make sure my server has started?_**
 
-There are a couple really great modules that we recommend using for this, {% url '`wait-on`' https://www.npmjs.com/package/wait-on %} and {% url '`start-server-and-test`' https://github.com/bahmutov/start-server-and-test %}.
+We recommend these great modules for this use case:
+
+* {% url '`wait-on`' https://www.npmjs.com/package/wait-on %}
+* {% url '`start-server-and-test`' https://github.com/bahmutov/start-server-and-test %}
 
 **_How can I wait for my requests to be complete?_**
 
--- a/lib/tags/url.js
+++ b/lib/tags/url.js
@@ -91,6 +91,15 @@ function url (hexo, args) {
 
       return util.htmlTag('a', attrs, markdown)
     })
+    .catch((err) => {
+      /* eslint-disable no-console */
+      console.error(`Failed to validate "${props.url}".`, err)
+      /* eslint-enable no-console */
+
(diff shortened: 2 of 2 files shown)


source/faq/questions/using-cypress-faq.md
```diff
@@ -147,7 +147,10 @@ When you load your application using `cy.visit()`, Cypress will wait for the `lo
 
 **_In CI, how do I make sure my server has started?_**
 
-There are a couple really great modules that we recommend using for this, {% url '`wait-on`' https://www.npmjs.com/package/wait-on %} and {% url '`start-server-and-test`' https://github.com/bahmutov/start-server-and-test %}.
+We recommend these great modules for this use case:
+
+* {% url '`wait-on`' https://www.npmjs.com/package/wait-on %}
+* {% url '`start-server-and-test`' https://github.com/bahmutov/start-server-and-test %}
```
richdouglasevans: Reworded this because the original had a grammatical error and lists of hyperlinked modules "pops" better for me.

> I appreciate that last point is subjective 🤗

PR #893, opened 2018-09-01
[#470] Link to assertions from API page

### LINKED ISSUE #470: Link Assertions page to API
Users who are looking at API page have no idea where to find our assertions. Add "assertions" link to the left tab between "events" and "commands" to make it simple

richdouglasevans:
Closes #470 

I've added linkage per @bahmutov's instruction.

assertions


_config.yml
```diff
@@ -49,6 +49,7 @@ alias:
   why-cypress/: guides/overview/why-cypress.html
 
   api/index.html: api/introduction/api.html
+  api/assertions/assertions.html: guides/references/assertions.html
```
richdouglasevans: 🤔 It feels a bit awkward to have to use an explicit redirect. If there's a less awkward way to do this then lemme know.

jennifer-shehane: I think this is the way to do it, but yes, it is weird.

PR #895, opened 2018-09-01
[#766] Ensure every terminal command has the $ prefix

### LINKED ISSUE #766: Can we please add '$' before every terminal command
The docs look so confusing and bad without the `$` being prefixed before every terminal command.

screen shot 2018-07-09 at 12 17 36 pm

The `$` needs to be on every line. I'm not sure how you can do this in CSS, maybe we need to update the doc build process or something to give you a `<div>` or some other wrapper for each line number that you can target.

richdouglasevans:
Closes #766 

> The `$` needs to be on every line.

I've gone through all the `shell` blocks in the Cypress documentation source code and all commands are prefixed by `$`. However, achieving this effect requires knowing how to author such shell blocks.

To wit, in order to achieve the desired effect the author has to put each command _in its own block._

each

The following block is something that feels more intuitive to author but won't achieve the desired effect.

nope

Supporting the latter authoring style will require a change to the way that the code blocks are generated 'cos right now it's just one big `<code>` block with no internal structure to hang a `:before` off of. Just bounce this PR if the bigger change is the one you actually want.

> I'd like to check [if some of the bash code blocks are actual printout instead of command prompts.]

I checked: I didn't find any that were explicitly just printouts.

> Yeah, I'm on a looooong and tiring bus journey back home so this sort of task is fine for a captive audience 😫

I found a handful that were commands followed by the output of the command: IMO they look fine; here's an example of that kind of mixed content.

with-output


source/api/plugins/configuration-api.md
```diff
@@ -70,8 +70,14 @@ You could now swap out configuration + environment variables like so:
...
+```
+```shell
 cypress run --env configFile=qa
+```
+```shell
 cypress run --env configFile=staging
+```
+```shell
 cypress run --env configFile=production
 ```
```
richdouglasevans: "Fixing" the sole example of a series of shell commands within a single code block.

source/guides/core-concepts/dashboard-service.md
```diff
@@ -114,7 +114,7 @@ This helps us uniquely identify your project. If you manually alter this, **Cypr
 
 We recommend that you check your `cypress.json` including the `projectId` into source control. If you don't want your `projectId` visible in your source code you can set it as an environment variable from your CI provider using the name `CYPRESS_PROJECT_ID`. The exact mechanism for doing so depends on your CI provider but could be as simple as:
 
-```bash
+```shell
 $ export CYPRESS_PROJECT_ID={projectId}   # replace {projectId} with your actual projectId
```
richdouglasevans: Oops 😳

Fixing a code block that I added last week. I'm just used to fencing such blocks with `bash` rather than `shell`. One for the style guide.

source/guides/guides/command-line.md
```diff
@@ -351,9 +351,8 @@ DEBUG=cypress:* cypress run
 ```shell
 set DEBUG=cypress:*
 ```
-
 ```shell
-set DEBUG=cypress:*
+cypress run
```
richdouglasevans: Command was duplicated so I changed it to reflect what the earlier Linux/Mac example was doing.

richdouglasevans:
> Is there a reason we can't modify the internal hexo code generator to markdown and wrap every line in its own <span> so it has the required HTML structure to then use a :before on?

No; I think that's the right approach :+1:

I've only done this _tactical_ PR because:

* Achieving the desired effect was quick to do.
* Adding our own plugin to Hexo [looks straightforward](https://github.com/ele828/hexo-prism-plugin/blob/master/src/index.js#L101)... but it's more work than I personally wanted to do on that bus journey.

Quite happy for this PR to be rejected 'cos it is unashamedly tactical. I'll address it properly once I've got [this ticket for a High Level Architecture diagram](https://github.com/cypress-io/cypress-documentation/issues/872) sorted.

> I _suggest_ merging this PR _but leaving the issue open_ 'cos it'll make the code blocks in the docs look better in the interim until I sort the issue properly. But, I'm aware that I am just _Random Guy On The Internet_ so that's your call.

PR #929, opened 2018-09-11
add guidance for installing type definitions

blumk:
Reason: installing cypress won't automatically add the type definitions to package.json.


bahmutov:
Nope, this is not necessary to do - when you install Cypress it brings its types (including Mocha, Chai, etc) and you don't need any additional commands.

PR #933, opened 2018-09-11
encourage the usage of baseUrl

phoet:
hard-coding the url is probably not the best way to do it. is this the way it should be?

### DIFF
--- a/source/api/commands/url.md
+++ b/source/api/commands/url.md
@@ -50,7 +50,7 @@ Option | Default | Description
 ```javascript
 // clicking the anchor causes the browser to follow the link
 cy.get('#user-edit a').click()
-cy.url().should('eq', 'http://localhost:8000/users/1/edit') // => true
+cy.url().should('eq', Cypress.config().baseUrl + '/users/1/edit') // => true
 ```
 
 # Notes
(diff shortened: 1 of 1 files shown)


phoet:
@jennifer-shehane i think your way of writing assertions is better, i will change the example and provide a hint for how to use the baseUrl if required.

source/api/commands/url.md
```diff
@@ -86,6 +87,23 @@ But you may be wondering where the `url` property comes from.  Per the `window.l
...
 
+***Hardcoded versus using the configuration object***
+
+Instead of hardcoding the URL you can use the `baseUrl` of the {% url 'Cypress configuration' configuration %}.
+
+Given the remote URL, `http://localhost:8000/index.html`, these assertions are the same.
+
+```javascript
+cy.url().should('eq', 'http://localhost:8000//index.html')
+cy.url().should('eq', Cypress.config().baseUrl + '/index.html') // tests won't fail in case the port changes
```
brian-mann: `Cypress.config('baseUrl')` is better than `Cypress.config().baseUrl`

brian-mann:
I would write `Cypress.config().baseUrl` as `Cypress.config('baseUrl')`

PR #984, opened 2018-09-25
Fixes

### LINKED ISSUE #974: Error docs related to parallelization should have link to parallelization docs
Every error related to parallelization should have a reference to parallelization docs

I don't think the `--parallel` should be linked, because that simply leads users to the command line reference, which gives very little info

image

steakfrye:
Closes #974 
Changes made in source/guides/references/error-messages.md

Removed some of the --parallel flag links in the error pages to keep with the style of the --group flag and ease confusion. Added some links to guides regarding parallelization.


jennifer-shehane:
Hi @steakfrye, I think we would prefer to keep the links for the parallel flags, so that readers can reference the flag in the command line docs if necessary. What point of confusion were you seeing with the links on the parallel flag?

steakfrye:
To be honest, I'm not sure. I was trying to resolve issue 974 as it was requested, and I wondered about that myself.

PR #986, opened 2018-09-26
Split changelog up into smaller files to reduce build time

### LINKED ISSUE #965: Improve build time of documentation
Many of our contributors are not happy with the build time of the documentation. We'd like it to build faster 🏃 

Cause is inception of dependencies with checking internal links.

lilaconlee:
- fixes #965


bahmutov:
also think where `curl https://on.cypress.io/changelog` will redirect

brian-mann:
In development we only build the changelog for the last 5 versions (to avoid it taking so long). It's possible we could/should insert some text as the "6th" version which specifies that. 

When in staging or production, it will build the entire changelog.

lib/tags/changelog.js
```diff
@@ -0,0 +1,30 @@
...
+    return rawRender.call(this, hexo, str, { engine: 'markdown' })
+  }
+
+  const env = hexo.env.NODE_ENV || hexo.env.env || process.env.NODE_ENV || 'development'
+  const pathToChangelogs = path.resolve('source', '_changelogs', '*')
+
+  return globAsync(pathToChangelogs)
+  .call('reverse')
+  .then((paths) => {
+    const shouldSlice = env !== 'production' | 'staging'
```
brian-mann: should this be `||` and not `|` ?

lib/url_generator.js
```diff
@@ -230,6 +230,10 @@ function getLocalFile (sidebar, href, source, fullUrl) {
 function validateLocalFile (sidebar, href, source, render) {
   const hash = extractHash(href)
 
+  if (href.indexOf('changelog') !== -1) {
```
brian-mann: It's safe to use `String.prototype.includes` nowadays.

https://node.green/#ES2015-built-in-extensions-String-prototype-methods-String-prototype-includes

`if (href.includes('changelog'))`

cypress/integration/main_spec.coffee
```diff
@@ -141,14 +141,22 @@ describe "Main", ->
...
-    it "has table of contents and goes to 0.19.0", ->
-      cy.get("aside#article-toc")
-        .should("be.visible")
-        .wait(2000) # allows menuspy to load and set the menu links
-        .contains("0.19.0")
-        .click()
-      cy.url()
-        .should('include', '#0-19-0')
+    it "has a populated table of contents", ->
+      if Cypress.config('baseUrl').includes('localhost')
```
brian-mann: I'm not sure if we do this on the docs... but in other projects we set a environment variable in Cypress to be that of `process.env.NODE_ENV` so that you can go off of that.

For instance by using `cypress`plugins/index.js` you can use the configuration API...

```js
module.exports = (on, config) => {
  // set the env to be that of NODE_ENV
  config.env.NODE_ENV = process.env.NODE_ENV || 'development'
}
```

Now in your tests you can do...

```js
if Cypress.env('NODE_ENV') ...
```

cypress/integration/main_spec.coffee
```diff
@@ -141,14 +141,22 @@ describe "Main", ->
...
+        cy.get("aside#article-toc")
+          .should("be.visible")
+          .get(".toc-item")
+          .should("have.length", 5)
+        cy.url()
+          .should("match", /.+#\d+-\d+-\d+/)
+      else
+        cy.get("aside#article-toc")
+          .should("be.visible")
+          .wait(2000) # allows menuspy to load and set the menu links
```
brian-mann: There'a always a better way to handle this rather than use `wait(2000)` and without seeing what's happening, I'm not sure why this is necessary.

Cypress will automatically retry for 4 seconds in the first place.

You could move this into an assertion or as part of the `contains(...)`

```js
cy.get("aside#article-toc").contains("0.19.0", { timeout: 10000 })

// or

// wait up to 10 seconds for both assertions to pass
cy.get("aside#article-toc", { timeout: 10000 }).should("be.visible").and("contain", "0.19.0")
```

jennifer-shehane:
k, cool, will check it out in staging when it's rdy to merge then.

richdouglasevans:
Woot, the docs build so fast now 🚀 This really _is_ the real deal, nice one Lila 💯

cypress/integration/main_spec.coffee
lilaconlee: Wasn't able to get this wired up correctly, but seems like it should've been an easy fix. Think I'd vote for merging as is and I can dig into adding this to the docs when I'm blocked on other stuff.

lib/tags/changelog.js
```diff
@@ -18,7 +18,7 @@ module.exports = function changelog (hexo) {
   return globAsync(pathToChangelogs)
   .call('reverse')
   .then((paths) => {
-    const shouldSlice = env !== 'production' | 'staging'
+    const shouldSlice = env !== 'production' || 'staging'
```
brian-mann: I didn't catch this the first time around either - but you have two expressions here...

```js
env !== 'production'
// ...and...
'staging'
```

This wil result in 2nd `staging` expression always to be truthy.

Needs to be either... `env === 'production' || env === 'staging'` or something like `['production', 'staging'].includes(env)`

Also cannot use `!==` when comparing two things since one will always be true.

cypress/integration/main_spec.coffee
brian-mann: No worries - I think i'll hack on it so it doesn't get lost in the shuffle and you'll see what I mean.

PR #995, opened 2018-09-28
Fix baseUrl in Cypress plugin config

lilaconlee:
Should set env correctly in staging.


brian-mann:
Ooomph, thanks for fixing @lilaconlee this was my bad! Glad the tests caught this once it was merged into `develop` lol.

PR #1024, opened 2018-10-06
adding reference to fake emails to avoid conversion

### LINKED ISSUE #1002: fake emails are converting to `mailto` links
These should not be mailto links. 

screen shot 2018-10-02 at 1 26 39 pm

eperedo:
To avoid fake emails converted to mailto links I am using a fake solution adding a fake id reference 😶 
Closes #1002

### DIFF
--- a/source/guides/getting-started/writing-your-first-test.md
+++ b/source/guides/getting-started/writing-your-first-test.md
@@ -278,9 +278,9 @@ Or in the Given, When, Then syntax:
 
 > 1. Given a user visits `https://example.cypress.io`
 > 2. When they click the link labeled `type`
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
I thought that setting `autolink` to `false` within our [`hexo-renderer-marked` plugin](https://github.com/hexojs/hexo-renderer-marked#options) would prevent making links like this, but it looks like we already have it set to false.

And looking into their code it doesn't look like they are the one transforming this into a link, so now I'm not sure where this comes from. :(

I don't particularly like the solution of adding a broken link to this, although it's slightly better than open the mail app on click.

jennifer-shehane:
Closing - don't particularly like this solution. Thanks for your contribution though!

PR #1031, opened 2018-10-10
Refresh faq answers

bahmutov:
Added a few more questions and answers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3647,9 +3647,9 @@
       "integrity": "sha1-QGXiATz5+5Ft39gu+1Bq1MZ2kGI="
     },
     "debug": {
-      "version": "3.2.5",
-      "resolved": "https://registry.npmjs.org/debug/-/debug-3.2.5.tgz",
-      "integrity": "sha512-D61LaDQPQkxJ5AUM2mbSJRbPkNs/TmdmOeLAi1hgDkpDfIfetSrjmWhccwtuResSwMbACjx/xXQofvM9CE/aeg==",
+      "version": "3.2.6",
+      "resolved": "https://registry.npmjs.org/debug/-/debug-3.2.6.tgz",
+      "integrity": "sha512-mel+jf7nrtEl5Pn1Qx46zARXKDpBbvzezse7p7LqINmdoIk8PYP5SySaxEmYv6TZ0JyEKA1hsCId6DIhgITtWQ==",
       "requires": {
         "ms": "^2.1.1"
       },
@@ -16191,6 +16191,15 @@
           "integrity": "sha512-vsGdkwSCDpWmP80ncATX7iea5DWQemg1UgCW5J8tqjU3lYw4FBYuj89J0CTVomA7BEfvSZd84GmHko+MxFQU2A==",
           "dev": true
         },
+        "debug": {
+          "version": "3.2.5",
+          "resolved": "https://registry.npmjs.org/debug/-/debug-3.2.5.tgz",
+          "integrity": "sha512-D61LaDQPQkxJ5AUM2mbSJRbPkNs/TmdmOeLAi1hgDkpDfIfetSrjmWhccwtuResSwMbACjx/xXQofvM9CE/aeg==",
+          "dev": true,
+          "requires": {
+            "ms": "^2.1.1"
+          }
+        },
         "folktale": {
           "version": "2.3.1",
           "resolved": "https://registry.npmjs.org/folktale/-/folktale-2.3.1.tgz",
@@ -16226,6 +16235,12 @@
           "requires": {
             "minimist": "0.0.8"
           }
+        },
+        "ms": {
+          "version": "2.1.1",
+          "resolved": "https://registry.npmjs.org/ms/-/ms-2.1.1.tgz",
+          "integrity": "sha512-tgp+dl5cGk28utYktBsrFqA7HKgrhgPsg6Z/EfhWI4gl1Hwq8B/GmY/0oXZ6nF8hDVesS/FpnYaD/kOWhYQvyg==",
+          "dev": true
         }
       }
     },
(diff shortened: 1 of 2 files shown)


source/faq/questions/using-cypress-faq.md
```diff
@@ -542,3 +541,18 @@ That being said, we use Cypress to test our own Desktop app's front end - by stu
 - Search existing {% url "open issues" https://github.com/cypress-io/cypress/issues %}, it may already be reported!
 - Update Cypress. Your issue may have {% url "already been fixed" changelog %}.
 - {% open_an_issue %}. Your best chance of getting a bug looked at quickly is to provide a repository with a reproducible bug that can be cloned and run.
+
+## {% fa fa-angle-right %} What are your best practices for organizing tests?
+
+We see organizations starting with Cypress by placing end-to-end tests in a separate repo. This is a great practice that allows someone on the team to prototype a few tests and evaluate Cypress without "scaring" everyone else with a new tool. As the time passes and the number of tests grows, we suggest moving end-to-end tests to live right alongside your front-end code. This brings two benefits:
```
jennifer-shehane: I'm not sure I agree with this. This is necessary for some projects, but I would outright recommend this. 

I also don't like equating using Cypress with "scaring everyone else" - def does not give a good impression.

bahmutov: yeah, scaring is not the word we want to use. But for general test organization, we have to say something. Starting with a separate project for just tests is ok I think, but migrating to have tests closer is much better - developers write tests, etc - same as our dashboard tests.

jennifer-shehane: None of our projects have a separate repo for just our tests though, which is why I think it's strange to have that as the first answer to this question.

bahmutov: yup, but we do a lot of things differently from a person who is trying to add Cypress to an existing project. Installing it right away on a large project would
- add a huge dependency which is not cached correctly on CI would absolutely cause howls of indignation
- could clash with TypeScript configs for Jest for example
- would immediately start argument: why use Cypress if it does not have cross-browser support

Using it in a separate project at first is the simplest way to try Cypress before jumping head first

jennifer-shehane: Maybe you should pull this out of the PR if you want this merged in quickly, cause I think there's general disagreement over this being a 'best practice'.

source/faq/questions/dashboard-faq.md
```diff
@@ -68,4 +68,20 @@ We have already begun the implementation for capturing even more things from you
...
 
+Cypress {% url "test parallelization" parallelization %} is indeed based on specs. For each spec the test runner scaffolds the new running context, in a sense isolating each spec file from any previous spec files, and ensuring a clean slate for the next spec. Doing this for each _individual_ test would be very very expensive and would slow down the test runs significantly.
+
+Spec file durations are also more meaningful and consistent than timings of individual tests, thus we can order specs by the moving average of the previously recorded durations. This would be much less useful when load balancing quickly finishing individual tests.
+
+To better load balance the specs, thus you would want more spec files with approximately the same running duration. Otherwise, a single very long running test might be the limit how fast all your tests finish, and the run completes. Due to starting a new test execution context before each spec file and encoding and uploading video after, making spec files to run shorter than approximately 10 seconds would also be fruitless - because Cypress overhead would eat any time savings.
```
lilaconlee: Think the thus in the first sentence can be removed

lilaconlee: might be the limit -> might limit

source/faq/questions/using-cypress-faq.md
brian-mann: I would actually argue that keeping the tests as close as possible, even modeling the tests to the code as closely as possible is absolutely **essential**.

Best practices are about what's ideal - not workarounds or "starters" for organizations wanting to dip their toes in.

For instance, to get a react application up and running, it's often much "easier and faster" to drop in a `<script src"https://some.cdn/react.js"></script>` right? But that's most certainly not a **best practice**.

Keeping the tests and the code aligned is really the only ideal way to manage a large suite of tests, because you...

1) want to automatically run the tests when the code changes
2) so that when tests fail, you can associate the commit sha which points at the exact snapshot of code
3) and allow developers of the project to continually keep up with and maintain the tests

Pulling the tests out separately would fail all 3 of those things, and brings us back to the antithesis of modern development, which is having QA as a separate department and separate function where the feedback loop is far too long and prone to internal communication breakdowns and errors.

Baking the code, quality and tests directly into the codebase so that engineers have the responsibility of writing and maintaining them is exactly what **shift left** is - and it's the blueprint for modern, healthy, fast, iterative development.

lilaconlee: This might be good for a "Is Cypress right for me?" section or something emphasizing that writing tests in a separate repo is good for the discovery process/introduction to a large project.

lilaconlee: I think if it's just a straight up best practice, I'd think that Cypress was a bit pushy thinking I'd be happy to add a new repo to my org just for them.

source/faq/questions/using-cypress-faq.md
```diff
@@ -542,3 +541,18 @@ That being said, we use Cypress to test our own Desktop app's front end - by stu
...
+
+- keeps tests and the features they tests in sync
+- allows code sharing between the application code and the tests (like selectors)
+
+## {% fa fa-angle-right %} What is the right balance between custom commands and utility functions?
+
+There is already a great section in {% url "Custom Commands" custom-commands#Best-Practices %} guide that talks about trade-offs between custom commands and utility functions. We feel reusable functions in general are a way to go. Plus they do not confuse {% issue 1065 'IntelliSense like custom commands do' %}
```
lilaconlee: Think this needs some ending punctuation

source/faq/questions/using-cypress-faq.md
bahmutov: I think we are making a big assumption about people willing to bring big tools and changes to their codebase right away. Starting with tests in a separate repo to show how little is needed to start testing is good practice to start with. Then moving tests to be close to code IS better of course, for all the reasons above, and IS better practice, so whoever wants to phrase it better, be my guest. 

Of course dropping `<script src="react ...>` is NOT the way to production React code. But trying out React like that IS good. And then using create-react-app for example when you know React is for you.

bahmutov: Also, why I say "start in a separate repo, then bring it in". When I demo Cypress I don't use local TodoMVC usually, I just point at `todomvc.com`. When I wrote about testing Hacker News for Vue I did not run local fork of the app - instead just point at the url https://vuejsdevelopers.com/2018/01/29/vue-js-e2e-test-hacker-news/ The barrier to trying Cypress should be low, and then a natural progression is to include it with code. The beauty of Cypress is that it does NOT need your code to be successful.

source/faq/questions/dashboard-faq.md
bahmutov: done

source/faq/questions/using-cypress-faq.md
brian-mann: I think your approach is valid Gleb - I just don't think its the answer to the question of:

**What are your best practices for organizing tests?**

This isn't a question that says: "How do you recommend first trying out Cypress?" or "How should I go about bringing Cypress into my organization?".

I feel like your answer is fine given that question- but when it comes to **best practices** which is the ideal way in which we believe Cypress should be utilized, suggesting anything short of what we **want** a successful user/organization/adopter of Cypress to look like is a disservice to all of our users.

brian-mann: Sorry my comment was out of date - your updated answer is great.

PR #1041, opened 2018-10-11
Convert browser gifs to mp4s

### LINKED ISSUE #1040: Some images getting 403s
From https://docs.cypress.io/guides/guides/launching-browsers.html#Browser-Icon: 
screen shot 2018-10-11 at 10 38 12 am

lilaconlee:
- Fixes #1040

### DIFF
--- a/source/guides/guides/launching-browsers.md
+++ b/source/guides/guides/launching-browsers.md
@@ -97,13 +97,13 @@ Cypress automatically disables certain functionality in the Cypress launched bro
 
 You might notice that if you already have the browser open you will see two of the same browser icons in your Dock.
 
-![switching browsers](/img/guides/switching-between-cypress-and-other-chrome-browser.gif)
+{% video local /img/snippets/switching-between-cypress-and-other-chrome-browser.mp4 %}
 
 We understand that when Cypress is running in its own profile it can be difficult to tell the difference between your normal browser and Cypress.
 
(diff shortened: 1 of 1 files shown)


lilaconlee:
Quality is a bit low, would be great to confirm if that's expected.

screen shot 2018-10-11 at 11 15 47 am

jennifer-shehane:
Did you use the conversion script for this? https://github.com/cypress-io/cypress-documentation/blob/develop/cy_scripts/snippet-creator/convert.js#L6

@Bkucera I actually can't track down where this is documented now.

brian-mann:
I don't think he wrote any docs on this but the CLI has some args and info here: https://github.com/cypress-io/cypress-documentation/blob/develop/cy_scripts/snippet-creator/cli.js

This needs to be added to the readme for sure. I thought all you do is add them to a folder and run the script and it'll auto convert them and you don't even need to pass any arguments, etc.

Can't quite remember but we need it all documented.

lilaconlee:
The mp4s already existed, but they weren't being used.

Bkucera:
Looks like i just missed changing these references from `.gif` to `.mp4` after the conversion :sweat_smile:

Bkucera:
to address @brian-mann , yes I should better document the usage.
Basically you call the script with a path to a `.gif` or `mp4`, and it will place it in snippets alongside a generated thumbnail

PR #1042, opened 2018-10-11
Add terminology rules for markdown linting

### LINKED ISSUE #952: Add automated terminology checks
The [style guide](https://github.com/cypress-io/cypress-documentation/issues/853) mentions a good bit of product terminology that we could give a go automating with tools like [textlint-rule-terminology](https://github.com/sapegin/textlint-rule-terminology).

Imagining this running as part of `lint`, but could see wanting to pull it out as part of a separate content-linting script.

lilaconlee:
Waiting for review of rules before actually running `--fix` and applying them.

- Fixes #952


lilaconlee:
Yeah I had Node.js in there initially, but was concerned about other uses of "node". Might make a manual pass converting to "Node.js" and add it to the style guide but not add it as a rule.

jennifer-shehane:
@lilaconlee Looks like there are some broken url links in the build, would it have replaced text within a url link?

lilaconlee:
Probably! That's why I pulled out the `url` checks hah. I'll dig into it.

lilaconlee:
So that was due to updating "Typescript" to "TypeScript" in a header, which was an easy internal fix. My other concern is if anyone has that hash saved, it'll just take them to the top of the page instead of the hash now.

I think that's fine, but would be down to change it back if we think it's not worth it.

PR #1103, opened 2018-11-01
Add note about Electron not accepting flags in browser launch plugin

### LINKED ISSUE #1099: Document that sending flags to Electron on Browser Launch does not work.
Document: https://docs.cypress.io/api/plugins/browser-launch-api.html#Usage

Need to add note that sending flags to Electron is not currently supported. See issue: https://github.com/cypress-io/cypress/issues/1519

jennifer-shehane:
- close #1099


brian-mann:
I'm super confused... this is completely incorrect - you can **absolute** pass flags to electron. 

You're getting confused between what actual CLI flags there are that affect chromium (before the process is launched) and actual browser launching flags.

You can absolutely without a doubt **do both**.

PR #1131, opened 2018-11-09
Add history tag

### LINKED ISSUE #716: Include the versions commands changed / were added / updated, etc
Now that we have had several breaking changes and upgrades, we need to probably start versioning the API commands so people can see when they were added, etc

lilaconlee:
- Fixes #716 

Added a history block to `writeFile` as an example usage, I'd update it if we wanted to add content along with the tag logic as part of this PR (or remove it if we rather separate content/functionality PRs).

history


lilaconlee:
Got the outline to this, which seems better than not having one at all but still not my fave. 
screen shot 2018-11-09 at 11 04 07 am

jennifer-shehane:
Cool, I can't figure out where that extra 30 px is coming from either 😞

PR #1143, opened 2018-11-13
Add Josh J TDD talks

CodingItWrong:
Added two talks I did using Cypress for test-driven development, in React and in Vue.

Also removed an older Vue meetup talk that was the same topic; the newer version is much more refined, so there's no reason to link to the old one any longer.

screen shot 2018-11-13 at 5 02 39 pm


CodingItWrong:
That's what I hoped to link to at first, but when I go to that URL I'm taken to the home page of their site. I've asked them about fixing it, but in the meantime linking to the conference page seems to be the closest I can get

CodingItWrong:
Gregg is looking into it now, so you can hold on merging this until maybe we get the direct link working. I'll push up another commit at that time

CodingItWrong:
@jennifer-shehane Link is now working and pointing directly to the talk!

PR #1145, opened 2018-11-14
Added parallelization and load balancing, and grouping diagrams 

### LINKED ISSUE #1142: Add new diagrams from public website
- [x] parallelization diagram 
- [x] load balancing chart 
- [x] grouping diagram

anjanab96:
closes #1142


source/guides/guides/parallelization.md
```diff
@@ -151,18 +153,26 @@ For multiple runs to be grouped into a single run, it is required for CI machine
...
+
+  ```shell
+  cypress run --record --group Windows/Chrome-69
+  ```
+
+- The second group can be called `Mac/Chrome 70`. 
 
   ```shell
-  cypress run --record --group electron
+  cypress run --record --group chrome --browser Mac/Chrome-70
```
jennifer-shehane: Let's make sure we don't miss up the arguments here. The arguments passed to the flag will be the one immediately following the flag. So, how this is written above actually says:

- group = 'chrome'
- browser = 'Mac/Chrome-70'

These arguments need to be switched around. The `--browser` flag is what's used to define the browser we want to run Cypress in and the `--group` flag is where we pass what we want to name our group.

source/guides/guides/parallelization.md
```diff
@@ -151,18 +153,26 @@ For multiple runs to be grouped into a single run, it is required for CI machine
...
   ```
 
-- The second group can be called `chrome`.
+- The third group can be called `Linux/Electron`. *Electron is the default browser used in Cypress runs*.
 
   ```shell
-  cypress run --record --group chrome --browser chrome
+  cypress run --record --group chrome --browser Linux/Electron
```
jennifer-shehane: See above

source/guides/guides/parallelization.md
```diff
@@ -68,7 +70,7 @@ As more and more tests are recorded to the Cypress Dashboard, Cypress can better
 
 ## Spec duration history analysis
 
-{% img 'no-border' /img/guides/parallelization/spec-forecast.png "Spec duration forecasting" %}
+{% img 'no-border' /img/guides/parallelization/load-balancing.png "Spec duration forecasting" %}
```
jennifer-shehane: Should be able to delete this 'spec-forecast' image if it's not being used elsewhere.

source/guides/guides/parallelization.md
jennifer-shehane: Actually - since Electron is the default browser, this one would not use the browser flag. Instead the `--browser chrome` flag should be added to the Windows example above.

PR #1159, opened 2018-11-20
changes to width size and adding image-rendering css

GeraAlcantara:
# [1032] UI - Cypress logo scaling issue
I have change the width to 102px and add some code to improve image rendering


GeraAlcantara:
Done!

PR #1161, opened 2018-11-22
Update visit.md Documentation

### LINKED ISSUE #1154: Explain in cy.visit() that the test will reload each time called - should set baseUrl instead to not reload.
>https://docs.cypress.io/api/commands/visit.html mentions setting baseUrl as a convenience feature, but not that it will cause a test restart if it is not start. Could that be explained there, please?

Should probably just link to thee best practices section.

JasonObeid:
Closes #1154: cy.visit() documentation update

Added an anti-pattern warning with a link to the best practices page.

### DIFF
--- a/source/api/commands/visit.md
+++ b/source/api/commands/visit.md
@@ -5,6 +5,11 @@ title: visit
 
 Visit a remote URL.
 
+{% note warning 'Anti-Pattern' %}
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
Thanks for adding this! I edited a little bit of the warning messaging. 👍

PR #1165, opened 2018-11-23
Add docs for requestTimeout and responseTimeout in cy.wait

jtymes:
Adding docs for new options introduced in cypress-io/cypress#2829.


jennifer-shehane:
@jtymes Will this PR also not remove the generic 'timeout' option?

jtymes:
@jennifer-shehane I did not remove the `timeout` option. I went the safest route to prevent breaking existing tests that may be relying on that timeout option today. Do you have a guideline for deprecations?

jennifer-shehane:
Ok, we can update this after the PR is formally reviewed if needed.

We are beginning to incorporate a 'history' section into each command's doc, so that we can track changes like this. It has not been incorporated into every command yet, but is being added as we go forward. writefile has an example of `history` table. https://github.com/cypress-io/cypress-documentation/blob/develop/source/api/commands/writefile.md#L182

PR #1179, opened 2018-11-28
Add instructions to collect git related info using environment variables

### LINKED ISSUE #1176: Add instructions to collect git related info with docker-compose
>It took me time to know that `Cypress` is expecting `.git` folder to be there to get all git related info, because when I tried it for the first time (without `docker-compose`), I was able to see commit author and message in the dashboard, but once I integrated it with `docker-compose` I wasn't able.
>
>I would suggest to add this information clearly as part of the documentation, and I'd be happy to be able to pass them as environment variables (for security reasons)

mshaaban088:
This PR tackles few modifications to the [Continuous-Integration](https://docs.cypress.io/guides/guides/continuous-integration.html) documentation

- Update environment variables in 
   - Renamed `Other Configuration Values` to `Configuration Values`
   - Placed `Record Key` under `Configuration Values` (does that make sense?)
- Add section for git info in continuous-integration.md, which closes #1176

### DIFF
--- a/source/guides/guides/continuous-integration.md
+++ b/source/guides/guides/continuous-integration.md
@@ -301,6 +301,21 @@ Typically you'd set this inside of your CI provider.
 
 ![Travis key environment variable](/img/guides/cypress-record-key-as-env-var-travis.png)
 
+
+## Git Information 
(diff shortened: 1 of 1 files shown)


mshaaban088:
@jennifer-shehane How do you usually update PR to make it up-to-date with `develop`, do you recommend `rebase`, or just `Update branch` option from GitHub?

PR #1184, opened 2018-11-30
Updates to assertions page

lilaconlee:
Minor changes adjacent to #967 and 4.0/Chai upgrade, but none of these changes are related to changes themselves and are relevant in the current version. 

Biggest change is the addition of the aliases tag and removal of aliases in the chai TDD table. Totally down to change that back to the way it was if it seems preferable, but thought I'd give a go reducing the number of things in the table. 

screen shot 2018-11-30 at 4 30 03 pm


jennifer-shehane:
I love it! 😍

PR #1197, opened 2018-12-07
added example of chained wait calls

facetcounter:
Maybe this is obvious to others (or too much detail for most)  but I had to dig back to a much earlier conversation in gitter and derive it indirectly from a conversation between Brian Mann and Joe Hanink in 10/2016.

### DIFF
--- a/source/guides/guides/network-requests.md
+++ b/source/guides/guides/network-requests.md
@@ -213,6 +213,24 @@ cy.wait(['@getActivities', '@getMessages'])
 
 // these commands will not run until the wait command resolves above
 cy.get('h1').should('contain', 'Dashboard')
+
+//if your route is more generic, and you would like to check several return values you can chain .wait() calls
+
+  cy.server()
+  cy.route({
+    method: 'POST',
+    url: '/myApi',
+  }).as('apiCheck')
+  cy.visit('/')
+  cy.wait('@apiCheck').then(xhr => {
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
`cy.wait()` doesn't really get any benefit from being chained together, as `cy.wait()` will always be called on the global `cy` context and never use the arguments from the previous command - as noted in it's [requirements](https://on.cypress.io/wait#Requirements).

So, this is the equivalent of writing:

```
  cy.server()
  cy.route({
    method: 'POST',
    url: '/myApi',
  }).as('apiCheck')
  cy.visit('/')
  cy.wait('@apiCheck').then(xhr => {
    assert.isNotNull(xhr.response.body.data,"First API call has data");
  })
  cy.wait('@apiCheck').then(xhr => {
    assert.isNotNull(xhr.response.body.data,"Second API call has data");
  })
  cy.wait('@apiCheck').then(xhr => {
    assert.isNotNull(xhr.response.body.data,"Third API call has data");
  });
```

We advice chaining it off of `cy` for clarity, so that there is no confusion - what is yielded from the previous command is *not* being used in `cy.wait()`.

If you can update the code to match the chaining, I'd appreciate it. 

There are also a few of the changes that don't match our documentation linting rules - you can see these errors by running `npm run lint`. Could you fix these too?

facetcounter:
Jennifer, thanks for the feedback, I've incorporated it into my PR!

----->Nathan

PR #1213, opened 2018-12-12
add a few new plugins to the list

bahmutov:
- cypress-pipe
- cypress-plugin-snapshots
- cypress-graphql-mock
- cypress-xpath

screen shot 2018-12-12 at 8 03 47 am


bahmutov:
I didn’t even notice that we are doing it alphabetically

PR #1239, opened 2018-12-20
Fix Crypress typo

lilaconlee:


bahmutov:
thank you @lilaconlee  - this was error in the original video caption, and they have fixed it

PR #1270, opened 2018-12-29
add should(cb) examples for #1267

### LINKED ISSUE #1267: Assertion example for dynamically scoped class names
@LaurieLinz commented on [Fri Dec 28 2018](https://github.com/cypress-io/cypress/issues/3028)

### Is this a Feature or Bug?

Posted the question in gitter and @bahmutov asked me to open an issue.

### Current behavior:

Querying for a class name in a dynamic list is quite challenging with Cypress. Sub-string matches do not seem to pick it up. `Using indexOf` seems to be a good option and an assertion example would be a great help.

### Desired behavior:

A nice example of how to do this would be great!

### Steps to reproduce: (app code and test code)
Please see this convo: 
_cypress-io_cypress_-_gitter
_cypress-io_cypress_-_gitter2

### Versions

Version 3.1.3, Mac 10.13.6, Chrome 71.0.3578.98

bahmutov:
- closes #1267 
- [ ] clarify the `Assert explicitly within .should() - Any errors raised by failed assertions will immediately bubble up and cause the test to fail.` text because it is confusing, isn't the `should(cb)` retried until it times out or passes, but not immediately fail?
- [ ] add `cypress-pipe` example where `cy.pipe(cb1).should(cb2)` is used

### DIFF
--- a/source/api/commands/should.md
+++ b/source/api/commands/should.md
@@ -127,11 +127,11 @@ cy.get('#header a').should('have.attr', 'href', '/users')
 
 ## Function
 
-### Verify length, content, and classes from multiple `<p>`
-
 Passing a function to `.should()` enables you to make multiple assertions on the yielded subject. This also gives you the opportunity to *massage* what you'd like to assert on.
 
 Just be sure *not* to include any code that has side effects in your callback function. The callback function will be retried over and over again until no assertions within it throw.
+
+### Verify length, content, and classes from multiple `<p>`
 
 ```html
 <div>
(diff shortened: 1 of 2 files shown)


source/api/commands/should.md
```diff
@@ -166,9 +166,54 @@ cy
...
+```js
+cy.get('.docs-header')
+  .find('div')
+  // .should(cb) callback function will be retried
+  .should(($div) => {
+    expect($div).to.have.length(1)
+    
+    const className = $div[0].className
+
+    expect(className).to.match(/heading-/)
```
jennifer-shehane: If you want to assert that the className *starts* with 'heading-' as the title of this section says, this should be `/$heading-/` yes?

bahmutov: yes, makes sense, I would need to rewrite either the assertion to split the names or change what I call this

source/api/commands/should.md
```diff
@@ -166,9 +166,71 @@ cy
   })
 ```
 
-### Assert explicitly within `.should()`
+{% note warning %}
+Any value returned from `should(cb)` callback function will be ignored. The original subject will passed to the next command automatically.
+
+```
```
bahmutov: I have added the warning right next to `should(cb)` even if there is another warning way down the page. Think a little warning is good, even if repeated

PR #1308, opened 2019-01-14
add example of expect with string messages

bahmutov:
- a user asked this question, and I have noticed that there is no good answer. So I have expanded our `should(cb)` example with multiple assertions that can benefit from more context

screen shot 2019-01-14 at 2 45 08 pm


source/api/commands/should.md
```diff
@@ -253,6 +253,21 @@ cy.get('#todos li').should(($lis) => {
...
+  expect($lis.eq(0), 'first item').to.contain('Walk the dog')
+  expect($lis.eq(1), 'second item').to.contain('Feed the cat')
+  expect($lis.eq(2), 'third item').to.contain('Write JavaScript')
+})
+```
+
+These string messages will be shown in the Command Log giving each assertion more context.
+
+![Expect assertions with messages](/img/api/should/expect-with-message.png)
+
```
jennifer-shehane: Oh, I learned something new. Is there a larger library documentation this can link to? To show where these description arguments come from?

bahmutov: I will add link to https://www.chaijs.com/guide/styles/#expect

PR #1321, opened 2019-01-18
Add another overwrite example

lukeapage:
Provide info that would have saved me an hour: https://github.com/cypress-io/cypress/issues/3166

I could also include waitForImages or put that somewhere if you'd like it, but I think this covers a couple of things the visit example doesn't.


source/api/cypress-api/custom-commands.md
```diff
@@ -268,6 +268,28 @@ This is usually unnecessary because Cypress is already configured to swap out ba
...
+```javascript
+Cypress.Commands.overwrite('screenshot', (originalFn, subject, name, options) => {
+
+  // call another command, no need to return as it is managed
+  cy.get('.app')
+    .should('be.visible')
+
+    // overwrite the default timeout, because screenshot does that internally
+    // otherwise the `then` is limited to the default command timeout
+    .then({ timeout: 45000 },
```
jennifer-shehane: I believe this should correspond to the `responseTimeout` defined in the user's configuration. You could access this as `Cypress.config('responseTimeout')`

jennifer-shehane:
We run eslint on our coding examples within our documentation to ensure a cohesive style. There are some errors from the build in CI, could you fix these and commit again? Line number and char number are shown for each error.

screen shot 2019-01-19 at 12 12 51 am

lukeapage:
@jennifer-shehane fixed the comments and eslint

PR #1326, opened 2019-01-20
Adds cypress drag drop plugin.

bierik:
As requested in https://github.com/cypress-io/cypress/issues/857#issuecomment-454432997


rocketsurgeonboic:
how do you use options with this? Example force: true in case the element is covered?

bierik:
@rocketsurgeonboic This PR is closed, so could you please open an issue in https://github.com/4teamwork/cypress-drag-drop, which is the maintained repo. Thank you.

PR #1334, opened 2019-01-23
Update server.md

streamnsight:
note about fetch not being supported and requiring stubbing / polyfill


bahmutov:
@streamnsight thank you 2nd time for taking time to make a PR. I will take over this, since I am doing this in https://github.com/cypress-io/cypress-documentation/pull/1338 and I feel pointing at the relevant issue is better than suggesting a recipe.

PR #1350, opened 2019-01-25
Typo "can not" -> "cannot"

calvinballing:
Corrected several instances in the docs where "can not" should have instead been "cannot".


jennifer-shehane:
We should add rule `["can[- ]not", "cannot$1"]` to the `.textlintrc` so these are caught in our linter next time.

calvinballing:
I've added this rule to the linter

jennifer-shehane:
@calvinballing Thanks! Yeah, there does seem to be exceptions, but probably better to have it be the default for now. 🤔

PR #1406, opened 2019-02-07
fix: wrong sinon stubbing API

juristr:
The docs seem to be out of date as there is a missing `.getLog()...` in the call to the Sinon stub.


chrisbreiding:
I tried this out and while using `.log()` works, it also works without, so I don't see any need to make it more verbose. It also looks to be [documented to work without `.log()`](https://sinonjs.org/releases/v7.2.3/spy-call/).

PR #1447, opened 2019-02-23
Update Error Handling documentation

### LINKED ISSUE #789: Remove confusing use of '.catch' in Conditional docs
- Either remove or put big warning that "THIS CODE DOES NOT WORK"

https://on.cypress.io/conditional-testing#Error-Recovery

AlexK777:
This doesn't remove `.catch()` from the code examples in the Error Handling section, but it does make it clear the code is not valid and is for demonstration purposes only.

closes #789


jennifer-shehane:
Thanks so much for the contribution! 💯

PR #1468, opened 2019-03-06
add better getCookie example

### LINKED ISSUE #1456: Add example to getCookie API page
to answer this question https://github.com/cypress-io/cypress/issues/3377#issuecomment-468651690

Our command log screenshot kind makes it look synchronous, so an example would be helpful

bahmutov:
- closes #1456


source/api/commands/getcookie.md
```diff
@@ -63,8 +63,43 @@ In this example, on first login, our server sends us back a session cookie.
...
+cy.getCookie('session_id')
+  .should('exist')
+  .then((c) => {
+    // save cookie until we need it
+    cookie = c
+  })
+// some time later, force the "cy.request"
+// to run ONLY after the cookie has been set
+// by placing it inside ".then"
+cy.wrap(null).then(() => {
```
jennifer-shehane: Is this really the best advice we have? cy.wrap(null)? 😥

bahmutov: actually you are right - this is no longer necessary, I can simply do `cy.then`

jennifer-shehane: I don't really think you're supposed to do `cy.then()` either. 😬We have an issue open to convert `.then()` to a child command - rendering this broken https://github.com/cypress-io/cypress/issues/466

Can't you just add in the example something arbitrary to chain it off of? Like... `cy.get('#submit').click().then(...`

bahmutov: reworked it one more time

PR #1472, opened 2019-03-07
tags: sort by dotted-decimal version in changelog

### LINKED ISSUE #1471: javascript sort ftw changelog fail
It sorts the older versions wrong

screenshot 2019-03-07 17 11 57

flotwig:
Fixes #1471 

image

### DIFF
--- a/lib/tags/changelog.js
+++ b/lib/tags/changelog.js
@@ -8,8 +8,9 @@ const rawRender = require('../raw_render')
 const globAsync = Promise.promisify(glob)
 
 const versionNumberSort = (b, a) => {
-  a = path.basename(a).split('.').map((n) => parseInt(n))
-  b = path.basename(b).split('.').map((n) => parseInt(n))
+  a = path.basename(a).split('.').map((n) => Number(n))
+  b = path.basename(b).split('.').map((n) => Number(n))
+
   if (a[0] !== b[0]) {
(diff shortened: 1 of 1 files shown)


lib/tags/changelog.js
```diff
@@ -7,6 +7,20 @@ const rawRender = require('../raw_render')
 
 const globAsync = Promise.promisify(glob)
 
+const versionNumberSort = (b, a) => {
+  a = path.basename(a).split('.').map((n) => parseInt(n))
```
brian-mann: generally best to just use `Number` rather than `parseInt`

```js
path.basename(a).split('.').map(Number)
```

lib/tags/changelog.js
```diff
@@ -18,8 +32,9 @@ module.exports = function changelog (hexo) {
   const pathToChangelogs = path.resolve('source', '_changelogs', '*')
 
   return globAsync(pathToChangelogs)
-  .call('reverse')
   .then((paths) => {
+    paths.sort(versionNumberSort)
```
brian-mann: javascript's mutation sort is usually frowned on... prob better to use `_.sortBy`

flotwig: `_.sortBy` doesn't let you use an arbitrary function to sort by, but I did update the PR to do `Array.sort` into a new array

PR #1486, opened 2019-03-13
Add Proxy Configuration to References

flotwig:
Documents changes from cypress-io/cypress#3531

Is the Configuration page a good spot for this?


flotwig:
@jennifer-shehane Thanks for the review. I moved the proxy configuration to it's own page in References, added some explanations for why Cypress needs a proxy to work and what that looks like, clarified the environment variables thing, and added some of those keywords that people might search for. I also added a section on how to check out their proxy config through the UI.

It might still need some work, so if you want to make some changes, please feel free to pick it up.

Re: adding info on how they can check if they're behind a proxy, I'm not sure about that. There are tools like http://amibehindaproxy.com/ that do some simple HTTP header checks, but it's easy to get a false positive/false negative and it might just confuse users more. I thought about saying "Unsure if you're behind a network proxy? Contact your administrator." in an infobox but I felt like that might be too obvious to bother writing. I'm open to suggestions.

jennifer-shehane:
@flotwig Right, I thought there may be an easier way to determine your proxy server, but I guess you're right - they likely don't have permissions to even check this.

I will review sometime today.

PR #1487, opened 2019-03-13
Fix table borders by making last th, td fill the remaining table width

### LINKED ISSUE #1183: Table headers not expanding for full width of table
Played around a bit with border-collapse and 100% width but couldn't get it quite right.
screen shot 2018-11-30 at 4 00 28 pm

flotwig:
Fixes #1183 

This visual bug has been bothering me more every time I notice it :sweat_smile: Is this a sufficient fix?

## Before

image

image

## After

image

image

### DIFF
--- a/themes/cypress/source/css/_partial/page.scss
+++ b/themes/cypress/source/css/_partial/page.scss
@@ -207,6 +207,10 @@ a.article-edit-link {
 
     td:last-child, th:last-child {
       width: 100%;
+
+      code {
+        white-space: normal;
+      }
     }
 
     tr:nth-child(2n) {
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
Some of the tables seem to end up cramming the lefthand column to be smaller than it reasonably should be. Maybe it's still better? I dunno.

Tried to fix with `white-space: nowrap` but that made some of the tables with long text unreadable as it scrolled to the right forever.

Screen Shot 2019-03-14 at 10 12 42 AM
Screen Shot 2019-03-14 at 10 12 23 AM
Screen Shot 2019-03-14 at 10 11 43 AM
Screen Shot 2019-03-14 at 10 11 06 AM

flotwig:
@jennifer-shehane Looks like setting `white-space: nowrap;` on all cells except for the last cells in the row has the desired effect, wrapping still works fine where needed and doesn't unnecessarily break the first cells:

image

image

image


I had to also make code blocks in the last cell wrap, for Assertions not to look weird:

image

PR #1511, opened 2019-03-22
Add select plugin

### LINKED ISSUE #1508: Explain how commands are chained better in "cy.wait" examples
Content update:

Based on a conversation with a user who was chaining `cy.wait(alias).then(...)` I have noticed that our `cy.wait` page only shows `cy.wait(alias).then((request object) => ...)` examples, but says nothing how this is not necessary if you just want to wait for the network call to happen.

Commands following `cy.wait` wait for it to finish automatically, we need to add an example like the one in slide https://gitpitch.com/cypress-io/testing-workshop-cypress?p=slides/05-xhr#/0/7 and link to section of the docs explaining that commands run serially https://on.cypress.io/introduction-to-cypress#Commands-Run-Serially

```js
it('starts with zero items', () => {
  // start Cypress network server
  // spy on route `GET /todos`
  // THEN visit the page
  cy.server()
  cy.route('GET', '/todos').as('todos')
  cy.visit('/')
  cy.wait('@todos') // wait for `GET /todos` response
    // inspect the server's response
    .its('response.body')
    .should('have.length', 0)
  // then check the DOM
  // note that we don't have to use "cy.wait(...).then(...)"
  // because all Cypress commands are flattened into a single chain
  // automatically. Thus just write "cy.wait(); cy.get();" naturally
  cy.get('li.todo').should('have.length', 0)
})
```

bahmutov:
- add cypress-select-tests
- adds note to `cy.wait alias`, closes #1508

### DIFF
--- a/source/_data/plugins.yml
+++ b/source/_data/plugins.yml
@@ -30,11 +30,6 @@
       link: https://github.com/cypress-io/cypress-webpack-preprocessor
       keywords: [webpack]
 
-    - name: cypress-select-tests
-      description: User space solution for grepping Cypress tests to run
-      link: https://github.com/bahmutov/cypress-select-tests
-      keywords: [browserify]
-
 - name: Development Tools
   plugins:
     - name: CircleCI Cypress Orb
@@ -46,7 +41,7 @@
       description: Several color themes for Cypress test runner
       link: https://github.com/bahmutov/cypress-dark
       keywords: [theme]
-      
+
     - name: cypress-protobuf
       description: Encode a fixture with Protocol Buffers.
       link: https://github.com/NoriSte/cypress-protobuf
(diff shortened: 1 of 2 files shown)


source/_data/plugins.yml
```diff
@@ -30,6 +30,11 @@
       link: https://github.com/cypress-io/cypress-webpack-preprocessor
       keywords: [webpack]
 
+    - name: cypress-select-tests
+      description: User space solution for grepping Cypress tests to run
+      link: https://github.com/bahmutov/cypress-select-tests
+      keywords: [browserify]
+
```
jennifer-shehane: Do you think this is best in preprocessors? I mean, I know it technically is - but wouldn't this be best under development tools?

bahmutov: good point :)

PR #1512, opened 2019-03-22
Add doc for window events

### LINKED ISSUE #1465: Add docs for `window:*` events
Document the [new `window:*` events](https://github.com/cypress-io/cypress/issues/2886) and add change to `window:before:unload` to the migration guide.

chrisbreiding:
TODO

- [ ] Do not merge until https://github.com/cypress-io/cypress/issues/3529 is released

Closes #1465

### DIFF
--- a/source/api/events/window-event.md
+++ b/source/api/events/window-event.md
@@ -1,0 +1,61 @@
+---
+title: window:&lt;event&gt;
+---
+
+You can use the `window:<event>` event to listen for arbitrary events on the page's `window`, where `<event>` is any event name that can be listened for on a page's `window`. This is the same as listening to the events with `window.addEventListener` with the following advantages:
+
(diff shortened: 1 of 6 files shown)


source/api/events/window-event.md
```diff
@@ -0,0 +1,61 @@
+---
+title: window:&lt;event&gt;
+---
+
```
jennifer-shehane: What if their page has several `window`s, like if they have iframes on the page, can they listen to the iframe's window events? Can we show an example of this if so?

chrisbreiding: No, it's the status quo as far as iframe support is concerned. It only binds to the AUT's top window. We'll have to do extra work when we add iframe support, but it shouldn't change the API. The window is available as `event.target`, so the user can know which iframe window it is based on that.

PR #1514, opened 2019-03-24
Add the cypress-protobuf plugin

NoriSte:
I created a Cypress plugin that allows encoding the fixtures with [Protobuf](https://developers.google.com/protocol-buffers/) before sending them to the web app. It allows to easily manage the fixtures in case your of Protobuf-based client-server communication.

### DIFF
--- a/source/_data/plugins.yml
+++ b/source/_data/plugins.yml
@@ -275,3 +275,10 @@
       description: A simple tool which integrates Cypress with Slack to report failing tests.
       link: https://github.com/bdimitrovski/cypress-healthcheck
       keywords: [reporter, slack, healthcheck]
+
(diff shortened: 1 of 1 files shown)


NoriSte:
Thank you so much @jennifer-shehane 😊
I'd like to know even if I can avoid the user from adding the plugin itself to the `plugins/index.js` file.

Instead of making (as I suggest to do in my readme)
```javascript
module.exports = on => {
  on("task", {
    protobufEncode: require("cypress-protobuf"),
  });
};
```
I'd like to let the user just import my plugin
```javascript
require("cypress-protobuf");
```
and manage myself the `on("task",` assignment.

Is it possible?

PR #1520, opened 2019-03-26
fix typo in example url. two slash should be one

wind0r:
example url is http://localhost:8000/index.html but code snipped used http://localhost:8000//index.html


wind0r:
No Problem :)
Thanks for the fast response and merge!

PR #1551, opened 2019-04-03
Update dashboard-service.md

mguery:
Re: [Issue 395](https://github.com/cypress-io/cypress-internal/issues/395) - Added "A user is anyone who logs in to the Dashboard service." under Manage Users.


jennifer-shehane:
@mguery We have some linting set up on the project - so that the tests fail if some of the text content doesn't match these rules. 

>Incorrect usage of the term: “Dashboard service”, use “Dashboard Service” instead  terminology

Do you have access to circle CI? Where you can see these errors? If not, I will invite you. Let me know.

PR #1558, opened 2019-04-04
command line option --spec should be --specs

staunsholm:
Changed documentation for commandline option `--spec` to be `--specs` as that is what actually works in Cypress :-)

### DIFF
--- a/source/guides/guides/command-line.md
+++ b/source/guides/guides/command-line.md
@@ -56,13 +56,13 @@ When calling a command using `npm run`, you need to pass the command's arguments
 ...and want to run tests from a single spec file and record the results on the Dashboard, the command should be:
 
 ```shell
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
@staunsholm This is not the case. From the Cypress code - it is `spec`, there is not even an alias for `specs` to allow this to work. It's possible, if you are not adding double quotes to the arguments after `--spec` that it is reading `--specs` as `--spec` somehow.

**Source code:**

https://github.com/cypress-io/cypress/blob/master/cli/lib/exec/run.js#L34

staunsholm:
Turns out cypress was wrapped inconsistently in the solution I was using. Sorry for wasting your time.

jennifer-shehane:
@staunsholm No worries! Documentation contributions are always welcome 😄

PR #1578, opened 2019-04-13
Add Azure DevOps guide - pipeline variables

### LINKED ISSUE #1579: Passing pipeline variables on Azure DevOps NPM task
Currently the documentation is lacking regarding passing environment variables when adding a new NPM task in an Azure DevOps build phase.

Since Azure DevOps is such a common CI provider the docs should be updated with a solution enabling users to read pipeline variables in Cypress from within NPM tasks.

nicolasdelfino:
Add a guide explaining how to pass pipeline variables in Azure DevOps through a custom NPM task and how to read the variables in Cypress.

Fixes #1579


bahmutov:
Hi @nicolasdelfino 

Thank you for taking time to add this section to the documentation. I am not sure if this makes sense though. 

Take a look at the linked kitchensink repo - https://github.com/cypress-io/cypress-example-kitchensink#ci-status I have both the basic and the parallel CI examples set up on Azure CI 

If there is more to do there - like to pass additional environment variables, you could show how to do there as a comment in https://github.com/cypress-io/cypress-example-kitchensink/blob/master/basic/azure-ci.yml or we could link a section in the README. But in this document that shows just the basic things for multiple CIs your new section immediately goes into details for a very Azure-specific detail that is of no use to most users. 

I am not even sure why you need to pass additional environment variables like that - doesn't Node Task on Azure grab the existing environment variables by default?

nicolasdelfino:
Hi @bahmutov !

> But in this document that shows just the basic things for multiple CIs your new section immediately goes into details for a very Azure-specific detail that is of no use to most users.

Yeah you're right about it not belonging there, had a feeling it didn't but went for it anyway

> I am not even sure why you need to pass additional environment variables like that - doesn't Node Task on Azure grab the existing environment variables by default?

Using a Node task instead would mean bypassing the package.json if I understand you correctly, meaning they would do something like this directly in an Azure task:

```javascript
./pathToCypress run cypress --env $(envVar1),$(envVar2)
```
Think this would work just fine but imo I would opt for keeping the execute call to Cypress in the repo instead of the CI Provider, also I think that most new users would try to use an npm task to connect the dots instead of something like a Node task, which is the reason my guide assumes it.

> also I think that most new users would try to use an npm task to connect the dots instead of something like a Node task, which is the reason my guide assumes it.

Quoting myself now but if you feel my is assumption is wrong then let us burry this PR, if nothing else a closed PR is also searchable and works as "documentation" :)

nicolasdelfino:
@bahmutov @jennifer-shehane closing this PR due to the conversation above.

PR #1581, opened 2019-04-14
Fix typo in writing-your-first-test.md

chaficnajjar:
Replace "Now that we've got a page loading" with "Now that we've got a page loaded".


jennifer-shehane:
Thanks so much!

PR #1615, opened 2019-04-24
fix wrong variable name in should.md

ChrisRose:


bahmutov:
Thanks @ChrisRose for noticing this!

PR #1636, opened 2019-04-29
Update docs for type - tab support

zhammer:
1. According to https://github.com/cypress-io/cypress/issues/299#issuecomment-469792368, "the actual tab implementation will not be a separate command", so the current documentation "Tabbing will be implemented as a separate command as `.tab()`" is out of date.
2. Users can use cypress-plugin-tab as a temporary fix.
3. Users can follow along with the Native Browser Events issue.


zhammer:
great flow for submitting / testing a documentation PR by they way!

PR #1647, opened 2019-05-01
Update Travis example re: Travis env. update

alexkb0009:
In response to this issue in the primary repo - https://github.com/cypress-io/cypress/issues/4069

Travis build environment(s) have been updated from running Ubuntu 14 to Ubuntu 16 which requires `libgconf-2-4` (Electron dependency which might be removed in future) to be installed manually.

Also xvfb must be installed/started manually as well it seems -- though unsure if Cypress itself does this when ran headlessly so this could be moot.

### DIFF
--- a/source/guides/guides/continuous-integration.md
+++ b/source/guides/guides/continuous-integration.md
@@ -162,12 +162,11 @@ CI Provider | Example Project | Example Config
 language: node_js
 node_js:
   - 10
-services:
-- xvfb
 addons:
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


bahmutov:
@alexkb0009 could you add a comment to your addition saying that ubuntu 16 does not install this library by default, so we need to install it ourselves. Then we can land your change.

alexkb0009:
All set re: comment. Removed the ~ `services: [xvfb]` line as well after testing a Travis run without it.

PR #1649, opened 2019-05-01
Document retryOnStatusCodeFailure, retryOnNetworkFailure

### LINKED ISSUE #1648: Document changes to network retries
Document changes from cypress-io/cypress#4015

flotwig:
Closes #1648 

I wonder if we should document the nature of the request retries under the hood. Probably not, since it should be transparent to the user?


brian-mann:
Yeah I don't think we need to update docs anywhere related to automatically retrying failed network requests at the proxy level. Should be completely transparent.

PR #1665, opened 2019-05-08
Improved phrasing and fixed typo

benjamincrozat:

### DIFF
--- a/source/guides/getting-started/installing-cypress.md
+++ b/source/guides/getting-started/installing-cypress.md
@@ -15,9 +15,9 @@ title: Installing Cypress
 
 Cypress is a desktop application that is installed on your computer. The desktop application supports these operating systems:
 
-- **macOS** 10.9 and earlier *(64-bit only)*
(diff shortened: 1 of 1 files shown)


source/guides/getting-started/installing-cypress.md
```diff
@@ -15,9 +15,9 @@ title: Installing Cypress
 
 Cypress is a desktop application that is installed on your computer. The desktop application supports these operating systems:
 
-- **Mac OS** 10.9+ (Mavericks+), only 64bit binaries are provided for macOS.
-- **Linux** Ubuntu 12.04+, Fedora 21, Debian 8, 64-bit binaries
-- **Windows** 7+, only 32bit binaries are provided for Windows.
+- **macOS** 10.9 and earlier *(64-bit only)*
```
jennifer-shehane: All of these should actually indicate "and above", since it should be 10.9 +

benjamincrozat: Sure, it's done. :)

PR #1669, opened 2019-05-09
Document `Cypress.isCy`

### LINKED ISSUE #1668: Document `Cypress.isCy()` interface
Not too complicated, just tells the user if the object is an instance of `cy` or a `cy` `Chainable` instance.

For cypress-io/cypress#3822

flotwig:
Fixes #1668


jennifer-shehane:
Needs to be added to on.cypress.io https://github.com/cypress-io/cypress-monorepo/issues/641

flotwig:
Yeah, the only examples I can think of are for plug-ins which want to change their behavior depending on if they've been passed a cy chainable or not. The only example I know of is in `cypress-pipe`: https://github.com/NicholasBoll/cypress-pipe/blob/d42577fdece4290d66a58b4277c86f2ad463860a/index.js#L47

PR #1685, opened 2019-05-16
Update chinese translation test

bahmutov:
Docker image `cypress/base:8.16.0` created in https://github.com/cypress-io/cypress-docker-images/pull/111


bahmutov:
Seems to be working fine now

Guides -- Chinese -- loads why Cypress

PR #1704, opened 2019-05-21
add a few more notes to visual testing page

bahmutov:
Just added a missing section in the beginning and general advice on recognizing when to start doing visual testing


source/guides/tooling/visual-testing.md
```diff
@@ -2,6 +2,8 @@
 title: Visual Testing
 ---
 
+# Functional vs visual testing
```
jennifer-shehane: So, I think the double header at the beginning of the article looks strange. We always start into a paragraph of text or have the 'What you'll learn' block. I didn't think the header was really necessary. 

Screen Shot 2019-05-22 at 11 42 03 AM

bahmutov:
kk, I can add "what you will learn" section in the beginning good idea. I just wanted a title for that really long introduction, otherwise the "Tooling" subheader was referring to something so far away on the page

PR #1792, opened 2019-06-19
3.3.2 release

jennifer-shehane:
Changelog and new documentation for 3.3.2 release


source/guides/getting-started/installing-cypress.md
```diff
@@ -288,3 +288,39 @@ CYPRESS_DOWNLOAD_MIRROR="https://www.example.com" cypress install
 ```
 
 Cypress will then attempt to download a binary with this format: `https://www.example.com/desktop/:version?platform=p`
+
+## Opt out of sending exception data to Cypress
```
brian-mann: You should also mention that users can just set the environment variable when running Cypress - rather than exporting it.

```
CYPRESS_CRASH_REPORTS=0 cypress open
CYPRESS_CRASH_REPORTS=0 cypress run
```

jennifer-shehane: Yah, I doubt they would want to do this one-off though, right? Plus this is Linux specific.

brian-mann: They might want to do that in CI. It's not linux specific - it's linux + OSX specific. Only windows is different.

PR #1813, opened 2019-06-25
dont return from a custom command

phoet:
see comment

### Translations updated

- [ ] Not applicable (this is not a change to an `en` doc content).


phoet:
not an expert on the custom commands, but i ran into this error because i was following the example.

if i understand the linked document correctly

> For convenience, you can also omit any return value or return undefined and Cypress will not error.

not returning should be the default. if this is so, i would just add that sentence instead of the link

jennifer-shehane:
Closing, do not think this is necessary in this example.

PR #1821, opened 2019-06-27
Add more code coverage information

bahmutov:
- added new section to code coverage guide explaining fullstack coverage
- added all existing code coverage repos as links to examples section

For now only did the English doc


.textlintrc
```diff
@@ -26,8 +26,10 @@
         ["(Chrome ?)DevTools", "Developer Tools"],
         ["(?!^Chrome $)DevTools", "Developer Tools"],
         "ESLint",
+        ["Express.js", "Express"],
```
bahmutov: nice!

.textlintrc
```diff
@@ -92,7 +97,9 @@
         ["can[- ]not", "cannot$1"],
 
         // Multiple words
+        ["back-?end(s?)", "back end$1"],
```
bahmutov: I never know which version to use

jennifer-shehane: I just picked having them as multiple words. I'm not really sure there is a right answer - but we want consistency anyways.

PR #1860, opened 2019-07-10
Translation for Chinese

icelemon1314:
Translation for Chinese. Thanks to all the people involved in the translation.

### Translations updated

- [ ] Not applicable (this is not a change to an `en` doc content).


icelemon1314:
In the circleci check, I got the error, but I think this may not be a problem, I think we should modify the circleci program.

/root/cypress-documentation/source/zh-cn/guides/references/assertions.md
   53:10  ✓ error  Incorrect usage of the term: “regexp”, use “RegExp” instead  terminology
  135:9   ✓ error  Incorrect usage of the term: “html”, use “HTML” instead      terminology

/root/cypress-documentation/source/zh-cn/guides/tooling/reporters.md
  164:49  ✓ error  Incorrect usage of the term: “Cypress'”, use “Cypress's” instead  terminology

jennifer-shehane:
@icelemon1314 We have text linting to ensure English terminology has the correct and consistent spelling and capitalization. These checks are correct. `regexp` should be changed to `RegExp`, `html` should be change to `HTML` and `Cypress'` should be changed to `Cypress's`

icelemon1314:
image

So there will be one more s on the page.

jennifer-shehane:
No, this should be 'Cypress', there may be a linting error where it's reading `Cypress'` within the url as being an apostrophe instead of escaping the url value. I can take a look at it.

PR #1907, opened 2019-07-25
write new copy and example to explain 'then' yield

### LINKED ISSUE #1902: Fix `.then` documentation dealing with yielded value
In our `cy.then` Yields section we talk about special meaning of `undefined` returned value, which is incorrect if there is a Cy command.

Situation: https://github.com/cypress-io/cypress-example-kitchensink/pull/275

Also fix examples (`input` vs `button`) in https://docs.cypress.io/api/commands/then.html#DOM-element

jennifer-shehane:
- close #1902 
- fix input -> button

### DIFF
--- a/source/api/commands/then.md
+++ b/source/api/commands/then.md
@@ -70,11 +70,11 @@ cy.get('button').then(($btn) => {
 })
 ```
 
-### The number is yielded
+### The number is yielded from previous command
 
 ```js
 cy.wrap(1).then((num) => {
-  cy.wrap(num)).should('equal', 1) // true
+  cy.wrap(num).should('equal', 1) // true
 }).should('equal', 1) // true
 ```
 
--- a/source/api/commands/then.md
+++ b/source/api/commands/then.md
@@ -46,7 +46,7 @@ Pass a function that takes the previously yielded subject as its first argument.
 
 `.then()` is modeled identically to the way Promises work in JavaScript.  Whatever is returned from the callback function becomes the new subject and will flow into the next command (with the exception of `undefined`).
 
-Additionally, the last Cypress command in the callback function will be yielded as the new subject and flow into the next command if there is no `return`.
+Additionally, the result of the last Cypress command in the callback function will be yielded as the new subject and flow into the next command if there is no `return`.
 
 When `undefined` is returned by the callback function, the subject will not be modified and will instead carry over to the next command.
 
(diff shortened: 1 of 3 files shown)


source/api/commands/then.md
```diff
@@ -46,6 +46,8 @@ Pass a function that takes the previously yielded subject as its first argument.
 
 `.then()` is modeled identically to the way Promises work in JavaScript.  Whatever is returned from the callback function becomes the new subject and will flow into the next command (with the exception of `undefined`).
 
+Additionally, the last Cypress command in the callback function will be yielded as the new subject and flow into the next command if there is no `return`.
```
bahmutov: `the command will be yielded` or the `result of the last Cypress command in the callback function will be yielded`?

source/api/commands/then.md
```diff
@@ -70,6 +72,17 @@ cy.get('button').then(($btn) => {
...
 
+### The subject is changed with another command
+
+```javascript
+cy.get('button').then(($btn) => {
+  const cls = $btn.class()
+
+  cy.wrap($btn).click().should('not.have.class', cls)
+    .find('i')
+}).should('have.class', 'spin') // assert on i element
```
bahmutov: I feel this needs an explicit comment after `.find(i)` explaining that this is the last Cypress command, there is no `return` so the found `<i>` element will be yielded from `.then` callback

source/api/commands/then.md
```diff
@@ -70,6 +72,17 @@ cy.get('button').then(($btn) => {
...
+```
+
```
bahmutov: I would also include "simple" `cy.wrap(1).then(() => cy.wrap(2)).should('equal', 2)` examples for each (return, no return, return Cypress command), I think they make it pretty clear what is going on

bahmutov:
Let me also make a pass on this PR adding examples

bahmutov:
Adding more examples to the kitchensink page https://github.com/cypress-io/cypress-example-kitchensink/pull/280

source/api/commands/then.md
jennifer-shehane: yeah, I knew I needed some review on this language. Was hard to write. Thanks!

source/api/commands/then.md
```diff
@@ -68,18 +70,45 @@ cy.get('button').then(($btn) => {
 })
 ```
 
+### The number is yielded
+
+```js
+cy.wrap(1).then((num) => {
+  cy.wrap(num)).should('equal', 1) // true
```
bahmutov: syntax error `cy.wrap(num))` and this case really is also "subject is changed with another command", right

bahmutov:
@jennifer-shehane I approve but only if you fix the syntax error, and maybe add comments to the example

PR #1912, opened 2019-07-27
Update continuous-integration.md

lots0logs:
Update dependencies (electron now depends on gtk3 instead of gtk2)

### Translations updated

- [x] Not applicable (this is not a change to an `en` doc content).


jennifer-shehane:
Thanks for the contribution!

PR #1928, opened 2019-07-31
Develop

jennifer-shehane:


jennifer-shehane:
K, so I'm realizing there is an issue with the new banner - that the tests on docs-staging caught (not sure why it wasn't caught before), but the banner now pushed down the content of the page so that when you click on the rightside table of contents - it no longer correctly scrolls to the h1, h2, etc - the h1, h2 slides undernearth the nav as shown.

**Clicked on "Who Uses Cypress"**

Screen Shot 2019-07-31 at 5 59 13 PM

Looking into fixing this now.

jennifer-shehane:
omg, look at these beautiful cypress GitHub checks. 😍

jennifer-shehane:
Honestly, I don't know. Something is going wrong and it's not just about the header placement. The main navigation does not highlight correctly with the `.active` CSS style when visiting a page. The logic for this is here: https://github.com/cypress-io/cypress-documentation/blob/develop/scripts/helpers.js#L157 so this code is just not running?? 

I can't replicate this locally AT ALL. Ran it with all kinds of env vars.

jennifer-shehane:
So, now the populated changelog is the only failing test - which makes sense because this test is not run in development. 

Actually...I think there may be a bug in Cypress that is causing this failure. So...that's fun. 😬 	

Screen Shot 2019-08-01 at 2 02 54 PM

jennifer-shehane:
When clicking on table of contents - the article content slides under the header. We should probably write a test for this (click toc - make sure h2 `is.visible`)

Screen Shot 2019-08-01 at 3 38 45 PM

PR #1937, opened 2019-07-31
Header partial cache removal

TonySynenko:


KateAbramova:
@jennifer-shehane 

In this PR partial cache was removed, it was used according to  https://hexo.io/docs/templates.html#Optimization
After that we’ll check if that helps on stage

If that doesn’t help - than all stage files need to be deleted on S3 and build restarted.

Otherwise, we'll revert all the changes and look for another solution
https://github.com/cypress-io/cypress-documentation/pull/1938

PR #1945, opened 2019-08-01
Add hash scroll offset

TonySynenko:
Add temporary workaround for hash scroll


amirrustam:
@TonySynenko resolve that last conflict, and then feel free to squash & merge this in.

PR #1961, opened 2019-08-06
Copy untranslated english pages to other languages

bahmutov:
- [x] removes untranslated Japanese pages
- [x] removes untranslated Chinese pages

Instead looks at the Git information to find English pages without corresponding checked in fully translated file, and copies the English file as a pre-build step (called from `index.js` that runs Hexo)


bahmutov:
yeah, I don't think titles in each Markdown document matter - I usually know which document I am editing from the filename itself displayed on top of the document. I would not count just titles there as translations. Continuing with Chinese docs.

bahmutov:
I have removed the rest of the duplicates, and have updated the test to not look at the translated doc titles.

jennifer-shehane:
Yeah, I am suggesting just removing the metadata at the top of the pages - and using the defined titles in the relevant en.yml, ja.yml doc, which are all translated.

bahmutov:
@jennifer-shehane I think removing the meta-data title on each page and fetching it from the `language.yml` is good idea - just needs to be a separate work, since it is a little bit tricky

jennifer-shehane:
I *really* wish I had been given time to review this before merging in.

I tried writing some partial new info to a translated file (as is common), and am running into an issue with lint because the entire `/zh-cn` and `/ja` files are in `.gitignore`?

How does someone commit any translations to these actual translated docs if they are ignored? 

My use case is that I am adding some new wording to the `installing-cypress` doc in english - which is translated into chinese, so I just add the new note in English to the chinese doc so that it still stays up to date.

PR #1974, opened 2019-08-07
Fix: Translation fix in Chinese in chapter Permanent-trade-offs

MakDon:
> 你无法使用Cypress同时驱动两个浏览器在同一时间   

means 'you could not concurrently driver two browser at the same time", in which 'concurrently' is duplicated with 'at the same time'.
So I changed it to a more native expression.

### Translations updated

- [x] Not applicable (this is not a change to an `en` doc content).


jennifer-shehane:
Thanks for the fix!

MakDon:
@jennifer-shehane Hi jennifer, could you have a look at this PR and determind if it could be merged?

PR #2018, opened 2019-08-15
Added waitFor

dominikwilkowski:

### DIFF
--- a/source/_data/plugins.yml
+++ b/source/_data/plugins.yml
@@ -184,6 +184,11 @@
       link: https://github.com/NoriSte/cypress-wait-until
       keywords: [commands, wait, wait-until, recursive-promise, check-async-value, check-value, open-source-saturday]
 
+    - name: cypress-waitFor
+      description: A zero-dependency plugin for Cypress to wait for an element or a function.
+      link: https://github.com/dominikwilkowski/cypress-waitFor
+      keywords: [commands, wait, waitFor, loading, check-value, wait-for-element]
+
 - name: Extending other testing frameworks
   plugins:
     - name: cyphell


dominikwilkowski:
That should probably not fail the tests... :(

dominikwilkowski:
Hey @Bkucera 

From the readme:
> You sometimes have to wait for an animation, data loading or whatever before we can find an element on the page. This plugin allows you to wait for an element using Cypress own select engine and Promise ecosystem. Cypress comes with a get method which will wait for a timeout before failing which works most of the time as long as you use that method. If you need to use the jQuery selector or non-cypress methods you need to wait manually. This is where this plugin comes in.

My use-case is that I have a dynamic loop that I need to go through and read from the DOM first to know what to select. Think a game where you have a question and need to select the answer. This will end up in callback hell very quickly with Cypress Promises.

I'd be very happy if you could educate me how you'd do that with vanilla cypress:

```js
it('should carry through a round', function() {
	const questionID = Cypress.$('[data-question="true"] title').attr('data-id');
	const answerID = SIGNALS[ questionID ].id;
	const correct = new RegExp(`^(${ answerID })$`, 'g');
	const wrongs = new RegExp(`^(?!${ answerID }$).*$`, 'gm');

	// loop this for SIGNALS.length times
	cy
		.get('[data-answer=""]').contains( wrongs ).click()
		.get('[data-score]').should('contain', '-1')
		.get('[data-answer=""]').contains( wrongs ).click()
		.get('[data-score]').should('contain', '-2')
		.get('[data-answer]').contains( correct ).click()
		.get('[data-score]').should('contain', '-1')
		.getAllByText('Next question ⇢').filter(':visible').click()
		// now do it all again from the beginning for SIGNALS.length time
		cy.wrap( null ).should( ()=> {
			// get new questionID
			// get new answerID
			// get new correct
			// get new wrongs
	});
	// end loop
});
```

Bkucera:
@dominikwilkowski Sorry, I'm still not sure I understand. Can you show a side-by-side example of the Cypress code with callback hell vs using your plugin which alleviates the callback hell?

dominikwilkowski:
Hey @Bkucera 

Thanks for the feedback. I actually think you're right about the plugin not adding much value right now. I'm struggling through the implementation of promises in cypress so I'll close this PR until I have a real plugin to contribute ;)

For someone who found this issue, this is how I "solved" my issue described above. Note though that this is an [anti-pattern as described in the cypress docs](https://docs.cypress.io/api/commands/wait.html#Nesting). I just can't find a better way to do this.

```js
it('should carry through a round', function() {
	let questionID;
	let correct;
	let wrongs;
	let score = 0;

	// loop this for SIGNALS.length times
	cy
		.wrap( new Array( SIGNALLENGTH ) )
		.each( ( $li, index ) => {
			cy
				.wrap( null ).then( () => {
					const questionID = Cypress.$('[data-question="true"] title').attr('data-id');
					const answerID = SIGNALS[ questionID ].id;
					const correct = new RegExp(`^(${ answerID })$`, 'g');
					const wrongs = new RegExp(`^(?!${ answerID }$).*$`, 'gm');
				})
				.wrap( null ).then( () => {
					cy.get('[data-answer=""]').contains( wrongs ).click();
				})
				.wrap( null ).then( () => {
					score --;
					cy.get('[data-score]').should( 'contain', score );
				})
				.wrap( null ).then( () => {
					cy.get('[data-answer=""]').contains( wrongs ).click();
				})
				.wrap( null ).then( () => {
					score --;
					cy.get('[data-score]').should( 'contain', score );
				})
				.wrap( null ).then( () => {
					cy.get('[data-answer]').contains( correct ).click();
				})
				.wrap( null ).then( () => {
					score ++;
					cy.get('[data-score]').should( 'contain', score );
			});
	});
});
```

jennifer-shehane:
@dominikwilkowski This may not work completely since I didn't run this code, but this is how I would approach running the tests above just by my brief look at it. 

```js
it('should carry through a round', function() {
  // loop this for SIGNALS.length times
  cy
    .wrap( new Array( SIGNALLENGTH ) )
    .each(function () {
      let correct;
      let wrongs;
      let score = 0;

      cy.get('[data-question="true"] title').invoke('data', 'id').then((questionID) => {
        const answerID = SIGNALS[ questionID ].id;

        correct = new RegExp(`^(${ answerID })$`, 'g');
        wrongs = new RegExp(`^(?!${ answerID }$).*$`, 'gm');
      })

      cy.get('[data-answer=""]').contains( wrongs ).click()
        .then( () => {
					score --;
				})
      cy.get('[data-score]').should( 'contain', score );
			cy.get('[data-answer=""]').contains( wrongs ).click()
        .then(() => {
					score --;
				})
      cy.get('[data-score]').should( 'contain', score );
      cy.get('[data-answer]').contains( correct ).click()
        .then( () => {
					score ++;
        });
      cy.get('[data-score]').should( 'contain', score );
	});
});
```

I would suggest reading through some more of the docs, checking out what API commands are available and looking at our examples. I'm not sure we have one good place to look at to learn better how to approach what you're trying to do - since it would require a comprehensive understanding of how chaining in Cypress works and also of the other API commands that are available to work with. 

Let me know if you have some suggestions on what we could add to our docs that could help better understand how to write these tests though. I'd love to hear them!

PR #2045, opened 2019-08-27
add section explaining how to patch Cypress CLI module

bahmutov:
added another advanced debugging section showing how to patch Cypress CLI module


bahmutov:
Thank you

PR #2064, opened 2019-09-09
add question to faq about tests failing in CI

bahmutov:
It is pretty common question / complain from users and deserves an answer. I have built upon https://github.com/cypress-io/cypress/issues/5098#issuecomment-529594760

### DIFF
--- a/source/faq/questions/using-cypress-faq.md
+++ b/source/faq/questions/using-cypress-faq.md
@@ -588,3 +588,14 @@ Usually your end-to-end tests interact with the application through public brows
 
 - see {% url "Testing Redux Store" https://www.cypress.io/blog/2018/11/14/testing-redux-store/ %} blog post and {% url "Redux Testing" recipes#Blogs %} recipe.
 - see {% url "Testing Vue web applications with Vuex data store & REST back end" https://www.cypress.io/blog/2017/11/28/testing-vue-web-application-with-vuex-data-store-and-rest-backend/ %} blog post and {% url 'Vue + Vuex + REST Testing' recipes#Blogs %} recipe.
+
+## {% fa fa-angle-right %} Help, my tests pass locally, but fail in CI!
+
+There are a lot of reasons tests could fail in CI and pass locally.
+
+Some reasons this could happen:
+
+- Old Electron version. If you are executing tests using `cypress run`, it uses quite an old version of the Electron browser. While we are working on upgrading Electron, try running CI tests using Chrome browser.
+- Headless Electron mode. Even when using the default Electron browser, there are differences in its behavior between headless and headed modes. You can execute tests in CI in headed mode with `cypress run --headed`, which might avoid the failures.
+- Application build process. If there is a build process in between running tests locally and running in CI - sometimes Cypress test failures are highlighting a bug in the build process.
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
Will have to take some time to review this later, but my saved responses:

### 1

>There are a lot of reasons tests could fail in CI and pass locally. On top of that - running in CI is a bit of a black box. So, the combination of these 2 facts make it especially frustrating to debug failing tests that solely fail in CI. 
>
>Some reasons this could happen:
>
>1. There is a problem isolated to the Chrome version of our Electron browser. We're working on updating it every day. https://github.com/cypress-io/cypress/pull/4720 Installing a later Chrome version in Travis would highlight if this is your problem.
>2. If there is a build process in between running tests locally and running in CI - sometimes Cypress test failures are highlighting a bug in the build process.
>3. Variability in timings when running your application in CI. If you have a test that will only pass when a network request finishes in 10ms (which happens every time on your local machine) and then that request takes 20ms on your CI machine (for whatever reason from that machine - location, CPU, etc slows it down) - this can cause some unexpected failures. You'll want to make sure to write assertions around the important steps that need to be reached before moving on to the next action in your tests. for example - ensuring a network request has finished before looking for the DOM element that relies on the data from that network request. I would suggest reviewing the video of the test failures, looking at the timings of commands and actions in the Command Log, and comparing them to your local run's Command Log to highlight if this is the issue.

### 2

>Tests run in CI run within the Electron browser. So, the first suggestion is to run the tests locally within the Electron browser with DevTools open to see if there are any issues there. 
>
>Since you already mentioned running the tests locally with other browser combinations, next I would make sure that video recording and/or screenshots are turned on during the CI run. Review the video to see if anything can be detected there that is causing the failure. 
>
>If it's still not obvious, compare the [Command Log](https://on.cypress.io/test-runner#Command-Log) from within the video to the Command Log of the same test passing locally when run in Cypress. Take a screenshot of each at the same moment and compare each command side by side. Seeing differences of the execution time or missing commands, etc can sometimes illuminate differences in the timings of requests or other methods that may have caused an issue.

### 3 (more about flaky tests)

>Most often in cases of flaky tests, we see that there is not enough assertions surrounding the actions or XHR requests/responses necessary before moving on to the next assertion. So if there is any variation in the speed of the XHR requests response in interactive mode versus run mode, then there will start to be failures in one over the other - or flaky tests. 
>
>So, for example if you had a test to check that when you click a 'Save' button that the users page refreshes with the correct content, you may write a test like so:
>
>cy.get('.name-input').type('Jennifer')
>cy.get('form').contains('Save').click()
>cy.get('#user- name).should('contain', 'Jennifer')
>
>Each command by default will wait 4 seconds to meet it's requirements, but if the POST for the form request *sometimes* takeds longer to respond than it takes for the "cy.get('#user- name).should('contain', 'Jennifer')" to run, then it will fail sometimes. Because of this, we recommend asserting on as many required steps as possible - this also helps later to isolate where the exact failure is when debugging on an actual bug. 
>
>So, the example above would be less flaky written like this:
>
>cy.server()
>cy.route('POST', 'users/*').as('updateUser')
>cy.route('GET', 'users').as('getUsers')
>
>cy.get('.name-input').type('Jennifer')
>cy.get('form').contains('Save').click()
>cy.wait('@updateUser').then((xhr) => {
>expect(xhr.requestBody).to.have.property('name', 'Jennifer')
>}
>cy.url('contains', 'users/123')
>cy.wait('@getUsers')
>cy.get('#user- name).should('contain', 'Jennifer') 
>
>Now when the last line runs, we don't have to worry about the POST or that the POST sent the right information, that the url changed and the GET's or whatever requests needed to respond - we know it's in the right state to find the content here.

jsrhodes15:
Any update on this? We are having the same issue, and this is the only "documentation" we could find. BTW, thanks for this @jennifer-shehane

jennifer-shehane:
I opened a new issue for this because I no longer have time to address this fully. https://github.com/cypress-io/cypress-documentation/issues/2343

PR #2066, opened 2019-09-11
Update retry-ability.md

jnnnnn:
Add a warning against `cy.get().contains()` because it confused us for a while and seems to be a common anti-pattern (see #3745)


bahmutov:
yeah, I have re-read that section and the warning at the top seems really out of place, it kind of muzzles the explanation of the problem and the two solutions

jennifer-shehane:
closing due to inactivity - never responded to requested changes.

Issue #2067, opened 2019-09-11
Create a dedicated content type for the docs top banner


KateAbramova:
## I'm submitting a...

```
[ ] Bug report
[ ] Content update
[X] Process update (build, deployment, ... )
```

## Type of bug / changes

Create separate entity in Contentful for the Docs top banner.

PR #2068, opened 2019-09-11
Issue-2067

TonySynenko:
close: #2067 

change banner content_type from topBanner to docsTopBanner


amirrustam:
@magnolia-triplett is going to make sure we have new/appropriate content in place and then we can merge and ensure all is well in staging.

PR #2090, opened 2019-09-19
describe solution to xvfb problem

bahmutov:
I wanted an answer / help to answer Xvfb questions like
- https://github.com/cypress-io/cypress/issues/5168
- https://github.com/cypress-io/cypress/issues/2698


bahmutov:
good point @jennifer-shehane - we probably can rename it to something else, almost like "tips & tricks"?

PR #2103, opened 2019-09-26
Revert "issue 2071"

TonySynenko:
Reverts cypress-io/cypress-documentation#2072

Needs to be re-worked due to the latest changes


amirrustam:
@TonySynenko and @KateAbramova I understand the change here, but can you articulate why this change needs to be made before we merge in?

KateAbramova:
@amirrustam 

This logic doesn't work right for dev, there have to be different keys for stage environment. That will be done later and PR will be made with appropriate changes.

amirrustam:
@KateAbramova, ok merging in now since the content in staging is irrelevant. Please create dedicated issue for the fix and link me.

PR #2112, opened 2019-09-27
[WIP] Add typescript type info from `cypress` info, add tags for rendering type info

### LINKED ISSUE #2111: Proposal: Use typescript types to fill in API docs
We could use the typescript docs from the `cypress` repo to potentially autofill some of our API docs for us.

[`typedoc`](https://github.com/TypeStrong/typedoc) can generate machine-readable and HTML documentation from our `.ts` files.

Things that would have to be figured out:

- How does docs release work? How is this versioned against `cypress` repo?
	- Deploy docs after latest types are available on `cypress` `master` branch?
- What level of docs could be replaced by this?
	- Probably shouldn't put our entire API docs in Typescript
	- Would be nice to pull call signatures, property descriptions, default values
	- Could also pull examples from the docs - `@example` tags
	- Could add partials to existing `.md` docs: `{% call-signatures "cy.request" %}`, `{% properties "Cypress.RequestOptions" %}`
- Should we host `typedoc`'s HTML output and link to it whenever types are mentioned in the docs?

I have typedoc generating type info and some basic tags here: #2112

flotwig:
Closes #2111

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ cypress/videos
 # Dynamically parsed data
 
 source/_data/banners.yml
+source/_data/cypress-typedoc.json
 
 # Translations
 # By default we ignore each translation folder like "source/ja",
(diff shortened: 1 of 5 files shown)


jennifer-shehane:
Closing due to inactivity

PR #2128, opened 2019-10-03
Document @alias.all and @alias.2

### LINKED ISSUE #1573: Document @alias.all and @alias.2
## I'm submitting a...

```
[ ] Bug report 
[x] Content update
```

## Type of bug / changes

We should document the use of @alias.all and @alias.2 within a couple of documents:
- https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html
- https://docs.cypress.io/api/commands/get.html#Alias

When referencing aliases, you can reference @alias.all or @alias.2 and it will return respectively all of the aliases of that name or the 2nd request matching the name.

Examples:

```js
it("does not poll for list again", function() {
  cy.wait("@getList");
  cy.tick(10000);
  // does not poll for another request
  cy.get("@getList.all").should("have.length", 1);
});
```

```js
it("first request has correct data", function() {
  cy.route({
    url: 'users',
    response: 'fx:users'
  }).as("getUsers");
  cy.visit("/users");
  cy.get("@getUsers.1").then((xhr) => {
    expect(xhr.url).to.include("page=1"))
  })
});
```

clarmso:
The `.all` and `.2` syntax can be used on aliases.  They indicates all requests and second request respectively if they are used on xhr requests.

See https://github.com/cypress-io/cypress/pull/4248 and https://github.com/cypress-io/cypress/issues/3512 for the motivation for such a documentation update.

Close #1573

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -40,9 +40,10 @@ alias:
   index.html: guides/overview/why-cypress.html
 
   guides/index.html: guides/overview/why-cypress.html
-  guides.html/: guides/overview/why-cypress.html
+  guides.html: guides/overview/why-cypress.html
   guides/references/type/: api/commands/type.html
-  guides/integrating-cypress/features.html/: guides/guides/continuous-integration.html
+  guides/references/known-issues.html: 404.html
+  guides/integrating-cypress/features.html: guides/guides/continuous-integration.html
   guides/network-requests-xhr/: guides/guides/network-requests.html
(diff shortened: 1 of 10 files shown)


bahmutov:
also relevant https://github.com/cypress-io/cypress-example-recipes/pull/386

jennifer-shehane:
Closing as we decided against publicly documented this before the network changes.

PR #2136, opened 2019-10-05
Add note about potential IDE TS server issues

### LINKED ISSUE #1110: In TypeScript doc, mentioning restarting TS server
> we should note restarting the Typescript server (Cmd+Shift+P and then type "restart TS server) or restart VS Code (or whatever editor is used).

https://github.com/cypress-io/cypress/issues/1236#issuecomment-423824058

kylerob:
closes #1110

I added a warning note about potential IDE TS server issues that may throw off TS setups.

### DIFF
--- a/source/guides/tooling/typescript-support.md
+++ b/source/guides/tooling/typescript-support.md
@@ -43,6 +43,16 @@ The `"types"` will tell the TypeScript compiler to only include type definitions
 You can find an example of Jest and Cypress installed in the same project using a separate `tsconfig.json` file in the {% url cypress-io/cypress-and-jest-typescript-example https://github.com/cypress-io/cypress-and-jest-typescript-example %} repo.
 {% endnote %}
 
+{% note warning %}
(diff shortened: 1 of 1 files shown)


source/guides/tooling/typescript-support.md
```diff
@@ -43,6 +43,16 @@ The `"types"` will tell the TypeScript compiler to only include type definitions
 You can find an example of Jest and Cypress installed in the same project using a separate `tsconfig.json` file in the {% url cypress-io/cypress-and-jest-typescript-example https://github.com/cypress-io/cypress-and-jest-typescript-example %} repo.
 {% endnote %}
 
+{% note warning %}
+You may have to restart your IDE's TypeScript server if the setup above does not appear to work. For example:
+
+VS Code (within a .ts file):
```
bahmutov: I think it would be better to mention that restart TS server can help with intellisense while in a `.ts` or `.js` file

kylerob: @bahmutov Thanks for the review, I'll add js as well.

PR #2157, opened 2019-10-14
Update intelligent-code-completion.md

n4bb12:
Documents a simpler and shareable alternative to configuring the JSON schema in vscode user settings.

### DIFF
--- a/source/guides/tooling/intelligent-code-completion.md
+++ b/source/guides/tooling/intelligent-code-completion.md
@@ -117,6 +117,12 @@ To set up in {% url "Visual Studio Code" https://code.visualstudio.com/ %} you c
 }
 ```
 
+Or you can directly add a `$schema` key to the cypress config, which is also a way of sharing this setting with collaborators. 
+
+```json
+"$schema": "https://on.cypress.io/cypress.schema.json",
+```
+
 # See also
 
 - {% url 'Adding custom properties to the global `window` with the right TypeScript type' https://github.com/bahmutov/test-todomvc-using-app-actions#intellisense %}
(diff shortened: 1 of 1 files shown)


source/guides/tooling/intelligent-code-completion.md
```diff
@@ -117,6 +117,12 @@ To set up in {% url "Visual Studio Code" https://code.visualstudio.com/ %} you c
 }
 ```
 
+Or you can directly add a `$schema` key to the cypress config, which is also a way of sharing this setting with collaborators.
```
jennifer-shehane: When you say 'cypress config', I'm not sure where you are saying to put this - is this in the `cypress.json` file? Can you clarify - link to an external doc if possible explaining this `$schema` key. Thanks!

n4bb12: Hi @jennifer-shehane 

yes, directly in the cypress.json. Example:

```json
{
  "$schema": "https://on.cypress.io/cypress.schema.json",
  ...other cypress options
}
```

If you need more detail, this is described here and generally works for all JSON files:
https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings

bahmutov: Yup this works per project https://glebbahmutov.com/blog/json-schema-for-the-win/ you can also set it up for all Cypress json files like we already advise in the docs. I am all for this addition.

jennifer-shehane: Ok, great. This is fun because we are making the configuration file configurable in our next release, so I would just mention the 'configuration file'.

```suggestion
Or you can directly add a `$schema` key to your Cypress configuration file, which is a great way to share the schema with all collaborators of the project.
```

PR #2169, opened 2019-10-18
add faq: how to spy on console.log

gabbersepp:
dependent on https://github.com/cypress-io/cypress-example-recipes/pull/346


gabbersepp:
> I would link to the example recipe - so we have one source of truth where this lies. https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/stubbing-spying__console
> 
> The example page also needs to be updated to include the new stub example here: https://docs.cypress.io/examples/examples/recipes.html#Stubbing-and-spying

I fellowed the existing documentation that adresses "Stubbing Google Analytics | Use cy.stub() to test Google Analytics calls"

PR #2174, opened 2019-10-21
.rightclick command

jennifer-shehane:
- Document new .rightclick command
- Update contributing docs on proper way to add new docs (considering new auto-generated translation files)


source/api/commands/rightclick.md
```diff
@@ -0,0 +1,199 @@
+---
+title: rightclick
+---
+
+Right click a DOM element.
+
+{% note warning %}
+`.rightclick()` will not open context menus native to the browser. `.rightclick()` should be used to test the behavior of an application where a user would right click a mouse.
```
Bkucera:
```suggestion
`.rightclick()` will not open context menus native to the browser. `.rightclick()` should be used to test your app's handling of right click related events such as `contextmenu`.
```

source/api/commands/rightclick.md
```diff
@@ -0,0 +1,199 @@
...
+{% yields same_subject .rightclick %}
+
+# Examples
+
+## No Args
+
+### Right click the menu
+
+```javascript
+cy.get('#context-menu').rightclick()
```
Bkucera: usually a context-menu is spawned as a result of a right click, so this example might be confusing (I guess this was in the driver tests?)

PR #2175, opened 2019-10-21
Run tests against examples in Markdown files

bahmutov:
Here is a first second stab at directly testing examples in our API commands Markdown files. I have added `@cypress/fiddle` and creates a command that points at the Markdown source pages. I have surrounded a few examples with fiddle block comments, so they are invisible in "normal" processing, but Cypress can run these files as test files

Screen Shot 2019-10-21 at 2 46 15 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
     "test-watch": "npm run test --silent -- --watch",
     "test": "jest",
     "warn-only": "stop-only --warn --folder cypress",
-    "make-empty-commit": "node node_modules/make-empty-github-commit/bin/empty.js --repo cypress-io/cypress-documentation"
+    "make-empty-commit": "node node_modules/make-empty-github-commit/bin/empty.js --repo cypress-io/cypress-documentation",
+    "test-markdown": "cypress open --config='integrationFolder=source/api/commands'"
   },
   "husky": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -90,6 +92,7 @@
     "chalk": "2.4.2",
     "cheerio": "1.0.0-rc.3",
     "coffee-react": "5.0.1",
+    "coffeescript": "2.4.1",
```
bahmutov: for some reason, `coffeescript` dependency was not found by default, even if it was a dependency of `@cypress/browserify-preprocessor`

bahmutov:
@jennifer-shehane here is an example of testing command examples

package.json
jennifer-shehane: I'm confused why we need this. There is no coffeescript in this repo at all. This a dependency of @cypress/browserify-preprocessor??

jennifer-shehane:
This does require maintenance. 

- I assume these tests will run in CI - running each time we push code - which requires supporting and documenting.
- Every js and html snippet now has these comments wrapping them - which requires supporting and documenting.

I understand - this may be nice, but this is extremely low priority compared to any other feature we could spend time developing in the Test Runner.

PR #2182, opened 2019-10-22
Update check.md

djaromir:
the example didn't work
you need to chain .check() on .get(), doesn't work with .should()

### DIFF
--- a/source/api/commands/check.md
+++ b/source/api/commands/check.md
@@ -105,7 +105,7 @@ You can ignore Cypress' default behavior of checking that the element is visible
 
 ```javascript
 cy.get('.action-checkboxes').should('not.be.visible') // Passes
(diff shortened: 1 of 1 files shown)


djaromir:
@jennifer-shehane thanks for looking at this
it says this to me:
image

`cy.get('#companyprofilecb').should('not.be.checked')
.check().should('be.checked');`

when changed to below it works

`cy.get('#companyprofilecb').should('not.be.checked');
cy.get('#companyprofilecb').check().should('be.checked');`

djaromir:
@jennifer-shehane thanks for looking at this
it says this to me:

image


`cy.get('#eandsdatacb').should('not.be.checked')
.check().should('be.checked');`

when changed to below it works

`cy.get('#eandsdatacb').should('not.be.checked');
cy.get('#eandsdatacb').check().should('be.checked');`

PR #2194, opened 2019-10-24
Use eslint plugin to catch exclusive tests

bz2:
On platforms without a decent grep, it seems stop-only does a poor
job of correctly detecting focused tests. There's an eslint plugin
which is much more suitable, use that instead.

While changing eslint config, tidy up the lint commands and fix a
couple of minor issues in files that were not being linted.

Exclude the themes/ folder from lint, cover everything else.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+/public/*
+/themes/*
--- a/circle.yml
+++ b/circle.yml
@@ -66,9 +66,8 @@ jobs:
       #
       - run: ls -la
       - run: npm run deps
-      - run: npm run lint
+      - run: npm run lint -- --max-warnings=0
       - run: npm run md-lint
-      - run: npm run stop-only
       - run: npm test
       - run: npm run build
 
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "plugins": [
-    "cypress"
+    "cypress",
+    "no-only-tests"
   ],
   "extends": [
     "plugin:cypress-dev/general",
@@ -10,5 +11,8 @@
     "es6": true,
     "node": true,
     "cypress/globals": true
+  },
+  "rules": {
+    "no-only-tests/no-only-tests": "warn"
   }
 }
--- a/lib/tags/assertions.js
+++ b/lib/tags/assertions.js
@@ -11,16 +11,12 @@ module.exports = function yields (hexo, args) {
   const type = args[0]
   const cmd = `<code>${args[1]}()</code>`
 
-  /* eslint-disable commas */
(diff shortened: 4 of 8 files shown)


bz2:
Example false positive from `stop-only` command:
```
> stop-only --folder cypress

Found .only here 👎
cypress/integration/contentful_banners_spec.js:55:  const onlyText = `
```
What this looks like with a test using `.only` using this branch:
```
husky > pre-commit (node v10.16.3)
  ↓ Stashing changes... [skipped]
    → No partially staged files found...
  ❯ Running linters...
    ↓ Running tasks for source/**/*.md [skipped]
      → No staged files match source/**/*.md
    ❯ Running tasks for *.js
      ✖ eslint --fix --max-warnings=0
        git add



✖ eslint --fix --max-warnings=0 found some errors. Please fix them and try committing again.

/home/martin/branches/cypress-documentation/cypress/integration/page_header_spec.js
9:6  warning  it.only not permitted  no-only-tests/no-only-tests

✖ 1 problem (0 errors, 1 warning)


ESLint found too many warnings (maximum: 0).
husky > pre-commit hook failed (add --no-verify to bypass)
```

bz2:
@bahmutov @jennifer-shehane Is there anything I can do to help you review this? I can rebase to resolve `package-lock.json` conflict when the change is approved.

circle.yml
```diff
@@ -66,9 +66,8 @@ jobs:
       #
       - run: ls -la
       - run: npm run deps
-      - run: npm run lint
+      - run: npm run lint -- --max-warnings=0
```
bahmutov: why `--max-warning=0`?

bz2: This reproduces the current flow when removing `npm run stop-only` - any `.only` causes the lint check to fail.

The question inside is why make `no-only-tests` a warning rather than an error. This is from experience with `npm start` on other projects, where lint errors cause a live reload to abort, but warnings allow it to complete. So, may be less relevant here but have found it helps combining TDD with in-browser development.

.eslintignore
```diff
@@ -0,0 +1 @@
+/themes/*
```
bahmutov:
```suggestion
/themes/*
/public/*
```
When docs build, they are placed into the "public" folder that can confuse the linter.

bz2: Done.

bz2:
@bahmutov Okay, all done.

Note that this time the `package-lock.json` change has ended up much larger, despite being the same method, I think because some intermediate change removed a bunch of `^` from transitive dependencies? Probably just from a npm version discrepancy? I am using node 10 and npm 6.12.1 at present.

Issue #2030, opened 2019-08-20
Switch npx vs $(npm bin)/cypress open commands in the install docs.


jennifer-shehane:
## I'm submitting a...

```
[ ] Bug report
[x] Content update
[ ] Process update (build, deployment, ... )
```

## Type of bug / changes

Do people use npm < 5.2? Should probably look up and see how popular this is somehow and maybe switch this recommendataion of how to run cypress open to just suggest npx cypress open first. 

https://docs.cypress.io/guides/getting-started/installing-cypress.html#Opening-Cypress

<img width="947" alt="Screen Shot 2019-08-20 at 4 24 17 PM" src="https://user-images.githubusercontent.com/1271364/63381628-143b7700-c367-11e9-8cb6-34acc4b12efd.png">

clarmso:
Yes, some people do still use an older version of Node/npm. I just took a quick look at the page. It's not wrong to leave the content as is.  The latest npm version is 6.11.3.

jennifer-shehane:
This was done recently

MikeMcC399:
- This was resolved by https://github.com/cypress-io/cypress-documentation/pull/5388 in Jul 2023

PR #2228, opened 2019-11-06
WIP - Expand global options docs w/ example snippets

### LINKED ISSUE #2227: Update global config docs to provide more examples
## Why

Page https://on.cypress.io/configuration is confusing and does not effectively explain to users how to change the Test Runner configuration values. I have seen people struggle to understand where to place run-time values when trying to run Cypress during workshops

## Some examples

- `baseUrl` is the very first option we list at the top of the page. Yet, we don't give a single example overriding it, and that is the most common option to override when running Cypress against a deployed system
- a few settings in the global table are both `cypress run` parameters AND valid values in `cypress.json`
- `env` can be set via `--env` AND via `--config env` AND via `CYPRESS_<name>` variables (aside from `cypress.env.json`). The number and boolean values are also automatically parsed into the right type, yet all we say is `{}`.
- we don't say the resolution order
- we should really give at least justiciation for `--config-file` (if you override multiple options)
- options dealing with files, like `testFiles` really would win from examples. For example, we could show how we can specific subset of all files inside `cypress.json` and then limit the specs to run using `--spec` CLI argument, and also takes `ignoreTestFiles` option. 
- we cannot link to individual settings by name, only to a table of settings

## Personal notes

I really think users do NOT need to spend a lot of time to read through the configuration page in order to be successful, they just need examples to copy and paste and learn from.

## Update global config page

Because we should provide individual addressable examples, we should split short tables into a hierarchy of options, each with its short example. We should alternate examples how we pass the option (CLI, cypress.json, environment variable, `--config`) so that users learn the possibilities, even if each configuration value does not show EVERY possible way

```
[ ] Bug report
[X] Content update
```

oak-wildwood:
Closes #2227

### DIFF
--- a/source/guides/references/configuration.md
+++ b/source/guides/references/configuration.md
@@ -14,27 +14,19 @@ The default behavior of Cypress can be modified by supplying any of the followin
 
 ## Global
 
-Option | Default | Description
------ | ---- | ----
-`baseUrl` | `null` | URL used as prefix for {% url `cy.visit()` visit %} or {% url `cy.request()` request %} command's URL
-`env` | `{}` | Any values to be set as {% url 'environment variables' environment-variables %}
-`numTestsKeptInMemory` | `50` | The number of tests for which snapshots and command data are kept in memory. Reduce this number if you are experiencing high memory consumption in your browser during a test run.
-`port` | `null` | Port used to host Cypress. Normally this is a randomly generated port
-`reporter` | `spec` | The {% url 'reporter' reporters %} used during `cypress run`
-`reporterOptions` | `null` | The {% url 'reporter options' reporters#Reporter-Options %} used. Supported options depend on the reporter.
-`watchForFileChanges` | `true` | Whether Cypress will watch and restart tests on test file changes
+The following options affect the running Cypresss instance globally. Refer to the examples below for different ways to apply these options. 
 
 ## `baseUrl`
 URL used as prefix for {% url `cy.visit()` visit %} or {% url `cy.request()` request %} command's URL
 
-**Default value:** null
+**Default value:** `null`
 
-**Example:** pass staging url during tests
+**Example:** pass staging url during tests by overriding ENV variable
 ```shell
 CYPRESS_baseUrl=https://staging.company.com npx cypress run
 ```
 
-**Example:** override local url using config option
+**Example:** override local url using `--config` option
 ```shell
 npx cypress run --config baseUrl=http://localhost:444
 ```
@@ -42,9 +34,14 @@ npx cypress run --config baseUrl=http://localhost:444
 ## `env`
 Any values to be set as {% url 'environment variables' environment-variables %}
 
-**Default value:** {}
+**Default value:** `{}`
 
-**Example:** override any environment variable using env option
+**Example:** override an environment variable using `--env` option
+```shell
+npx cypress run --env=host=local.dev
+```
+
+**Example:** override multiple environment variables using `--env` optiom
 ```shell
 npx cypress run --env=host=kevin.dev.local,api_server=http://localhost:8888/api/v1
 ```
@@ -52,11 +49,52 @@ npx cypress run --env=host=kevin.dev.local,api_server=http://localhost:8888/api/
 ## `numTestsKeptInMemory`
 The number of tests for which snapshots and command data are kept in memory. Reduce this number if you are experiencing high memory consumption in your browser during a test run.
 
-**Default value:** 50
+**Default value:** `50`
 
-**Example:** reduce this value to reduce memory consumption
+**Example:** reduce this value to reduce memory consumption using `--config` option
 ```shell
 npx cypress run --config --numTestsKeptInMemory=25
+```
+
+## `port`
+Port used to host Cypress. Normally this is a randomly generated port.
+
+**Default value:** `null`
+
+**Example:** provide a static port to host Cypress using `--port` option
+```shell
+npx cypress run --port=9000
+```
+
+## `reporter`
+The {% url 'reporter' reporters %} used during `cypress run`
+
+**Default value:** `spec`
+
+**Example:** provide a different reporter  using `--reporter` option
+```shell
+npx cypress run --reporter=junit
+```
+
+## `reporterOptions`
+The {% url 'reporter options' reporters#Reporter-Options %} used. Supported options depend on the reporter.
+
+**Default value:** `null`
+
+**Example:** provide options for a specific reporter using `--reporter-options` option
+```shell
+npx cypress run --reporter=junit \
+ --reporter-options=mochaFile=results/my-test-output.xml,toConsole=true
+```
+
+## `watchForFileChanges`
+Whether Cypress will watch and restart tests on test file changes
+
+**Default value:** `true`
+
+**Example:** disable file watching using `--config` option
+```shell
+npx cypress run --config watchForFileChanges=false
 ```
 
 ## Timeouts
--- a/source/guides/references/configuration.md
+++ b/source/guides/references/configuration.md
@@ -14,7 +14,7 @@ The default behavior of Cypress can be modified by supplying any of the followin
 
 ## Global
 
-The following options affect the running Cypresss instance globally. Refer to the examples below for different ways to apply these options. 
+The following options affect running Cypress when installed via {% url "direct download" https://on.cypress.io/installing-cypress#Direct-download %}. Refer to the examples below for different ways to apply these options.
 
 ## `baseUrl`
 URL used as prefix for {% url `cy.visit()` visit %} or {% url `cy.request()` request %} command's URL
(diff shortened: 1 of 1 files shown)


oak-wildwood:
Two screenshots below. @bahmutov - like to get your feedback

image

image

bahmutov:
I also wonder if we need to link this part of the testing workshop at the bottom of the guide, something like

Configuration and environment variables are also covered in [this section](https://github.com/cypress-io/testing-workshop-cypress/blob/master/slides/10-configuration/PITCHME.md), [slides](https://gitpitch.com/cypress-io/testing-workshop-cypress?p=slides/10-configuration) of the [Cypress Testing Workshop](https://github.com/cypress-io/testing-workshop-cypress/)

oak-wildwood:
Updated numTestsKeptInMemory as noted. @bahmutov 

image

oak-wildwood:
> I also wonder if we need to link this part of the testing workshop at the bottom of the guide, something like
> 
> Configuration and environment variables are also covered in [this section](https://github.com/cypress-io/testing-workshop-cypress/blob/master/slides/10-configuration/PITCHME.md), [slides](https://gitpitch.com/cypress-io/testing-workshop-cypress?p=slides/10-configuration) of the [Cypress Testing Workshop](https://github.com/cypress-io/testing-workshop-cypress/)

I like it, will add.

oak-wildwood:
Where is this in the code? I can't seem to find this anywhere. 

> We need better visual hierarchy on the right now:
> 
> Screen Shot 2019-11-06 at 5 33 18 PM

oak-wildwood:
> We need better visual hierarchy on the right now:

Without shrinking the options themselves down to ### (level 3) (currently ##) I think this looks the best. Kinda weird Options and Global are both at # (level 1). 

image

source/guides/references/configuration.md
```diff
@@ -14,15 +14,88 @@ The default behavior of Cypress can be modified by supplying any of the followin
...
+Any values to be set as {% url 'environment variables' environment-variables %}
+
+**Default value:** `{}`
+
+**Example:** override an environment variable using `--env` option
+```shell
+npx cypress run --env=host=local.dev
+```
+
+**Example:** override multiple environment variables using `--env` optiom
```
bahmutov: can you add one more variable value like `rate:20,selected=true` and add a sentence that boolean and numeric variables are automatically parsed from strings

source/guides/references/configuration.md
```diff
@@ -14,15 +14,88 @@ The default behavior of Cypress can be modified by supplying any of the followin
...
+```
+
+## `watchForFileChanges`
+Whether Cypress will watch and restart tests on test file changes
+
+**Default value:** `true`
+
+**Example:** disable file watching using `--config` option
+```shell
+npx cypress run --config watchForFileChanges=false
```
bahmutov: this option only affects the interactive mode `cypress open`

bahmutov:
@oak-wildwood I just realized that the global table is even worse and more confusing - some of the options are "normal" command line arguments to `cypress run` and `cypress open` command, while other options are `--config` only options.

For example `--port`, `--reporter` and `--reporterOptions` are command line arguments, which you can see from `npx cypress run --help`, the rest need `--config <name>=<value>` first, aside from `--env`

hehe, defnitely need to clarify

oak-wildwood:
Yeah I wondered about that. That's why I added the extra clarity at the end of each description, but maybe it needs to be reworked. Any suggestions? 
image

oak-wildwood:
There's also this page that explains in more detail. https://docs.cypress.io/guides/guides/environment-variables.html#Setting

source/guides/references/configuration.md
```diff
@@ -14,15 +14,88 @@ The default behavior of Cypress can be modified by supplying any of the followin
...
+**Default value:** `null`
+
+**Example:** pass staging url during tests by overriding ENV variable
+```shell
+CYPRESS_baseUrl=https://staging.company.com npx cypress run
+```
+
+**Example:** override local url using `--config` option
+```shell
+npx cypress run --config baseUrl=http://localhost:444
```
jennifer-shehane: We do not use `npx` in our documentation - there is an issue to change this convention that still requires considering. These examples need to match our other documentation and consistency of how we call npm commands. https://github.com/cypress-io/cypress-documentation/issues/2030

source/guides/references/configuration.md
```diff
@@ -14,15 +14,88 @@ The default behavior of Cypress can be modified by supplying any of the followin
...
-Option | Default | Description
------ | ---- | ----
-`baseUrl` | `null` | URL used as prefix for {% url `cy.visit()` visit %} or {% url `cy.request()` request %} command's URL
-`env` | `{}` | Any values to be set as {% url 'environment variables' environment-variables %}
-`numTestsKeptInMemory` | `50` | The number of tests for which snapshots and command data are kept in memory. Reduce this number if you are experiencing high memory consumption in your browser during a test run.
-`port` | `null` | Port used to host Cypress. Normally this is a randomly generated port
-`reporter` | `spec` | The {% url 'reporter' reporters %} used during `cypress run`
-`reporterOptions` | `null` | The {% url 'reporter options' reporters#Reporter-Options %} used. Supported options depend on the reporter.
-`watchForFileChanges` | `true` | Whether Cypress will watch and restart tests on test file changes
+The following options affect the running Cypresss instance globally. Refer to the examples below for different ways to apply these options.
```
jennifer-shehane:
```suggestion
The following options affect running Cypress when installed via {% url "direct download" https://on.cypress.io/installing-cypress#Direct-download %}. Refer to the examples below for different ways to apply these options.
```

source/guides/references/configuration.md
```diff
@@ -14,15 +14,88 @@ The default behavior of Cypress can be modified by supplying any of the followin
...
-`reporterOptions` | `null` | The {% url 'reporter options' reporters#Reporter-Options %} used. Supported options depend on the reporter.
-`watchForFileChanges` | `true` | Whether Cypress will watch and restart tests on test file changes
+The following options affect the running Cypresss instance globally. Refer to the examples below for different ways to apply these options. 
+
+## `baseUrl`
+URL used as prefix for {% url `cy.visit()` visit %} or {% url `cy.request()` request %} command's URL
+
+**Default value:** `null`
+
+**Example:** pass staging url during tests by overriding ENV variable
```
jennifer-shehane: These headings aren't consistent with the rest of our documentation - there should be a level `###` heading so as to enable linking. Showing the default value like this is also unconventional. I'd have to look at other examples of default values and edit this to be consistent.

source/guides/references/configuration.md
```diff
@@ -14,15 +14,88 @@ The default behavior of Cypress can be modified by supplying any of the followin
...
+```shell
+npx cypress run --env=host=local.dev
+```
+
+**Example:** override multiple environment variables using `--env` optiom
+```shell
+npx cypress run --env=host=kevin.dev.local,api_server=http://localhost:8888/api/v1
+```
+
+## `numTestsKeptInMemory`
```
jennifer-shehane: Should be spacing between headings and content - install markdownlint in VSCode for these recommendations.

source/guides/references/configuration.md
jennifer-shehane: Typo

```suggestion
**Example:** override multiple environment variables using `--env` option
```

source/guides/references/configuration.md
bahmutov: I think we are fine switching to `npx cypress run` for these examples - it has been in use for longer than 1 year, and looks much better and works cross-platform

jennifer-shehane:
Sorry, I'm looking over this again and I'm really confused what the intent of this PR originally was - the issue gives no detail as well as the PR comment on what and why this change is necessary. 

I was thinking this had to do with 'global install' from this content, 

>The following options affect the running Cypresss instance globally. Refer to the examples below for different ways to apply these options. 

And even recommended a change to that effect, but this page is just removing the original tables of configuration and showing examples of each use - so the change I mentioned earlier not necessary.

This change converts the table into what I consider to be a less readable configuration. There are several ways to set config - and I don't think this is the place to document every single way to set configuration for every single configuration available - this just serves as an index of all available config values.

**Before**

Screen Shot 2019-11-15 at 12 07 24 PM

**After**

Screen Shot 2019-11-15 at 3 12 58 PM

- cannot quickly scan available config, default value and definition
- the table of contents shows the config names for half the page - and then switches to the name of the headings of the tables below (seems because only 1 of the 8 tables was converted to a completely new format)
- this example is inaccurate `npx cypress run --config --numTestsKeptInMemory=25` will not work


**Please open an issue *clearly* describing the problem that needs to be solved so that we can work towards finding a solution as a whole - there are multiple ways that the issue could be resolved and I don't think this was the correct move forward, but again - I cannot find where the actual problem was described anywhere in GitHub.**

bahmutov:
our users are literally confused about options, see https://github.com/cypress-io/cypress/issues/5706#issuecomment-554519115 and we have definitely outgrew config table.

PR #2230, opened 2019-11-07
add cypress-skip-test plugin

bahmutov:
for doing things like `cy.skipOn('mac')` and `cy.onlyOn('chrome')`

### DIFF
--- a/source/_data/plugins.yml
+++ b/source/_data/plugins.yml
@@ -204,6 +204,11 @@
       link: https://github.com/FlorianGoussin/cypress-cy-select
       keywords: [commands, shorthand]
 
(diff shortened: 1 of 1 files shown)


source/_data/plugins.yml
```diff
@@ -204,6 +204,11 @@
       link: https://github.com/FlorianGoussin/cypress-cy-select
       keywords: [commands, shorthand]
 
+    - name: '@cypress/skip-test'
+      description: Simple commands to skip a test based on platform, browser or an url
```
kevinold:
```suggestion
      description: Simple commands to skip a test based on platform, browser or a url
```

PR #2244, opened 2019-11-13
docs: add examples to code-coverage

timdeschryver:
Hi, I was looking for an Angular code coverage example and didn't find one on the docs website. Later I found that there were [examples](https://github.com/cypress-io/code-coverage#examples) in the [cypress-io/code-coverage repo](https://github.com/cypress-io/code-coverage), and these docs included an Angular example.

This PR adds the examples from cypress-io/code-coverage to the docs page.


bahmutov:
Thank you @timdeschryver - we need to come up with some better way to avoid duplication and out of sync lists of resources

PR #2275, opened 2019-12-02
normalize and clean up reporter documentation

SavePointSam:
I was running into difficulties understanding the Reporters documentation between missing information, formatting, and structure. I went through the document and removed duplicate information and normalized the structure of the document to be consistent across sections.


source/guides/tooling/reporters.md
```diff
@@ -61,19 +93,25 @@ Some reporters accept options that customize their behavior. These can be specif
...
 cypress run --reporter junit \
   --reporter-options "mochaFile=results/my-test-output.xml,toConsole=true"
 ```
 
+{% note %}
 The above configuration will output the JUnit report to `STDOUT` and save it into an XML file. Reporter options differ depending on the reporter (and may not be supported at all). Refer to the documentation for the reporter you are using for details on which options are supported.
+{% endnote %}
 
-## Report per spec
+## Report Per Spec
```
jennifer-shehane: We use sentence case for our headings, so these should be left as is.

PR #2277, opened 2019-12-03
Add example of triggering mousedown for specific mouse buttons

jennifer-shehane:


yGuy:
IMHO this does not go far enough. The documentation still speaks about the "correct" usage for triggering a mouse down and fails to mention the button option. Without it, the event will not be compatible with a lot of code out there that actually uses this property.
IMHO the code should be improved to either set a default or issue a warning, but at the very least the "correct usage" example should be made "more correct".
I glanced through the code and [there](https://github.com/cypress-io/cypress/blob/2c88f0c58ec3c1bd8421a05259fc2ab893191f99/packages/driver/src/cy/mouse.ts#L423) does seem to be some default handling, but in my tests there is no "button" property on the synthetic event if you omit it from the "trigger" call.

PR #2295, opened 2019-12-09
add Electron switches documentation

### LINKED ISSUE #2294: Document passing additional Electron app command line switches
Add to https://on.cypress.io/browser-launch-api page section on how to pass via Cypress application command-line switches for Electron browser

- implementation https://github.com/cypress-io/cypress/pull/5891

bahmutov:
- closes #2294


source/api/plugins/browser-launch-api.md
```diff
@@ -56,6 +56,25 @@ module.exports = (on, config) => {
 }
 ```
 
+## Modify Electron app switches
+
+Cypress Test Runner is an Electron application, and its behavior (and the behavior of the bundled-in Electron browser) can be customized using command line switches. The supported switches depend on the Electron version, see {% url "Electron documentation" https://electronjs.org/docs/api/chrome-command-line-switches/history %}.
+
+Unfortunately, because one needs to pass these switches before the Cypress itself loads, you cannot use the plugins file. Instead you need to use an environment variable `ELECTRON_EXTRA_LAUNCH_ARGS`. For example, to disable HTTP browser cache and ignore certificate errors, open or run the Cypress Test Runner like this:
```
flotwig: It's not that unfortunate :) 

```suggestion
You can pass Electron-specific launch arguments using the `ELECTRON_EXTRA_LAUNCH_ARGS` environment variable. For example, to disable HTTP browser cache and ignore certificate errors, open or run the Cypress Test Runner like this:
```

source/api/plugins/browser-launch-api.md
```diff
@@ -56,6 +56,25 @@ module.exports = (on, config) => {
...
+
+```shell
+export ELECTRON_EXTRA_LAUNCH_ARGS=--disable-http-cache --ignore-certificate-errors
+npx cypress open|run
+```
+
+Cypress code already sets some of the Electron command line switches, see file {% url "packages/server/lib/environment.coffee" https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/environment.coffee %}. Unfortunately, there is no way to see all currently set switches after Electron's launch.
```
flotwig:
```suggestion
Cypress already sets some the Electron command line switches internally. See file {% url "packages/server/lib/environment.coffee" https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/environment.coffee %}. There is no way to see all currently set switches after Electron's launch.
```

PR #2311, opened 2019-12-16
add more diverse css selector examples to click

### LINKED ISSUE #2293: Add a few links to cy.click API page
I was looking at Gitter and messages like this one always 💔

Screen Shot 2019-12-09 at 9 31 01 AM

Why can't the users find the docs and examples for `click`?
- each command needs to have a link to its API page, probably when you click on it in the test runner, this is part of our general command and error UI improvement, whenever that gets shipped
- if the users have not scaffolded a new project and have not looked at the examples, then THEY HAVE NO IDEA ABOUT https://example.cypress.io/commands/actions even if they find https://on.cypress.io/click

Ughh.

I think we need to include link to the example page from `cy.click` and other relevant commands in our docs.

Also, a good idea is to include a few blog posts there, like https://www.cypress.io/blog/2019/01/22/when-can-the-test-click/

jennifer-shehane:
I think the key to this original issue is that I honestly can't tell where they were originally looking and what failed. Did they find the click doc and they just didn't know how to write a css selector? Or did they never make it to the click doc? 

I expanded the css selectors cause I think this is helpful anyways - added the other examples to See Also section.

- close #2293

### DIFF
--- a/source/api/commands/click.md
+++ b/source/api/commands/click.md
@@ -20,16 +20,16 @@ Click a DOM element.
 **{% fa fa-check-circle green %} Correct Usage**
 
 ```javascript
-cy.get('button').click()          // Click on button
-cy.focused().click()              // Click on el with focus
-cy.contains('Welcome').click()    // Click on first el containing 'Welcome'
+cy.get('.btn').click()          // Click on button
+cy.focused().click()            // Click on el with focus
+cy.contains('Welcome').click()  // Click on first el containing 'Welcome'
 ```
 
 **{% fa fa-exclamation-triangle red %} Incorrect Usage**
 
 ```javascript
-cy.click('button')          // Errors, cannot be chained off 'cy'
-cy.window().click()         // Errors, 'window' does not yield DOM element
+cy.click('.btn')          // Errors, cannot be chained off 'cy'
+cy.window().click()       // Errors, 'window' does not yield DOM element
 ```
 
 ## Arguments
@@ -67,10 +67,10 @@ Option | Default | Description
 
 ## No Args
 
-### Click the button
+### Click a link in a nav
 
 ```javascript
-cy.get('button').click()
+cy.get('.nav > a').click()
 ```
 
 ## Position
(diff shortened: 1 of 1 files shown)


source/api/commands/click.md
```diff
@@ -20,16 +20,16 @@ Click a DOM element.
 **{% fa fa-check-circle green %} Correct Usage**
 
 ```javascript
-cy.get('button').click()          // Click on button
-cy.focused().click()              // Click on el with focus
-cy.contains('Welcome').click()    // Click on first el containing 'Welcome'
+cy.get('.btn').click()          // Click on button
```
bahmutov: I would keep the original line too
```javascript
cy.get('button').click() // click on <button> element
cy.get('.btn').click()   // click on element with class "btn"
```

source/api/commands/click.md
```diff
@@ -100,27 +100,28 @@ cy.get('button').click(15, 40)
...
 ```javascript
-cy.get('button').click('bottomLeft', { force: true })
+cy.get('#collapse-sidebar').click('bottomLeft', { force: true })
 ```
 
 ### Force a click with relative coordinates
 
 ```javascript
-cy.get('button').click(5, 60, { force: true })
+cy.get('#footer .next').click(5, 60, { force: true })
```
bahmutov: maybe add a comment above explaining
```
// finds element with class "next" inside the element with
// id "footer" and clicks at specific location,
// disabling Cypress standard click tests
```

source/api/commands/click.md
```diff
@@ -100,27 +100,28 @@ cy.get('button').click(15, 40)
...
 ```
 
-### Click all buttons found on the page
+### Click all elements with id starting with 'btn'
 
 By default, Cypress will error if you're trying to click multiple elements. By passing `{ multiple: true }` Cypress will iteratively apply the click to each element and will also log to the {% url 'Command Log' test-runner#Command-Log %} multiple times.
 
 ```javascript
-cy.get('button').click({ multiple: true })
+cy.get('[id^=btn]').click({ multiple: true })
```
bahmutov: nice

bahmutov: I would also add an example showing how to click on an element with dynamic CSS style classes like 
```
// clicks on button with dynamic class
// like <div class="abc-save-123">
cy.get('class=*save').click() 
```

PR #2319, opened 2019-12-18
add network request example and recipe link

bahmutov:
Adds a little bit more to the Network requests guide that is missing assertion examples

### DIFF
--- a/source/guides/guides/network-requests.md
+++ b/source/guides/guides/network-requests.md
@@ -329,8 +329,9 @@ cy.route('POST', '/users').as('new-user')
 // trigger network calls by manipulating web app's user interface, then
 cy.wait('@new-user')
   .should('have.property', 'status', 201)
-// we can grab the XHR object again to run more assertions
-cy.wait('@new-user') // yields XHR object
+// we can grab the completed XHR object again to run more assertions
+// using cy.get(<alias>)
+cy.get('@new-user') // yields the same XHR object
   .its('request.body')
   .should('deep.equal', {
     id: '101',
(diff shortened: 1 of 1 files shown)


source/guides/guides/network-requests.md
```diff
@@ -318,6 +320,47 @@ cy.get('#results')
...
+
+```javascript
+cy.server()
+// spy on POST requests to /users endpoint
+cy.route('POST', '/users').as('new-user')
+// trigger network calls by manipulating web app's user interface, then
+cy.wait('@new-user')
+  .should('have.property', 'status', 201)
+// we can grab the XHR object again to run more assertions
+cy.wait('@new-user') // yields XHR object
```
jennifer-shehane: This will wait for an entirely new request/response to happen, so 2 POSTs. Did you mean to write `cy.get('@new-user')` here?

source/guides/guides/network-requests.md
```diff
@@ -318,6 +320,47 @@ cy.get('#results')
...
+  .its('request.body')
+  .should('deep.equal', {
+    id: '101',
+    firstName: 'Joe',
+    lastName: 'Black'
+  })
+// and we can place multiple assertions in a single "should" callback
+cy.wait('@new-user')
```
jennifer-shehane: also `cy.get('@new-user')`? I'm not toally familiar with whether it's suppose to be 3 POSTs or just one your asserting on since the "we can grab the XHR object again" comment.

source/guides/guides/network-requests.md
bahmutov: ughh, you are right, thanks for catching this. I need to `cy.wait` once, but then use `cy.get` to retrieve it if I need it again. This is awesome doc for our users

bahmutov:
I have added more tests and examples to the XHR assertions examples to confirm this behavior https://github.com/cypress-io/cypress-example-recipes/commit/fa1c8c03bc4faf8daaef2b715085f43ca4238133

Question: if the request happens several times, how do we retrieve each one?

bahmutov:
Relevant in separate recipe https://github.com/cypress-io/cypress-example-recipes/pull/386

PR #2326, opened 2019-12-23
Fix incorrect reference of requestBody

### LINKED ISSUE #2325: Doc: `request.body` is used instead of `requestBody`
## I'm submitting a...

- [x] Bug report
- [ ] Content update
- [ ] Process update (build, deployment, ... )

## Type of bug / changes

in network request https://docs.cypress.io/guides/guides/network-requests.html#Assertions, there is a code snippet regarding request body

```js
cy.get('@new-user') // yields the same XHR object
  .its('request.body')
  .should('deep.equal', {
    id: '101',
    firstName: 'Joe',
    lastName: 'Black'
  })
```

`request.body` is used instead of `requestBody` since I got autocomplete for that. Is using `request.body` expected in this case?

jennifer-shehane:
- close #2325 

This was introduced in newest examples to network requests doc recently


source/guides/guides/network-requests.md
```diff
@@ -336,7 +336,7 @@ cy.wait('@new-user')
 // we can grab the completed XHR object again to run more assertions
 // using cy.get(<alias>)
 cy.get('@new-user') // yields the same XHR object
-  .its('request.body')
+  .its('requestBody')
```
bahmutov: this is not a mistake. The XHR object has `request`, `response` AND `requestBody`, `requestHeaders`, `responseBody`, `responseHeaders` properties. See https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/server-communication__xhr-assertions/cypress/integration/spec.js#L111 for `request.body` access and screenshot https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/server-communication__xhr-assertions#cywait-vs-cyget-for-network-requests 

This is why I put `cy.get('@new-user').then(console.log)` as a tip - so people see the real options

jennifer-shehane: Then our typings are not correct in Cypress. 

Screen Shot 2019-12-23 at 9 58 10 AM

bahmutov:
Could be. What about doing just request property - typings could just not understand nested property access with lodash dot syntax “request.body”

jennifer-shehane:
@bahmutov Yah, just 'request' is fine with typings too.

Issue #2333, opened 2019-12-25
Can't start cypress documentation fork because of eslint-plugin-cypress-dev


viraxslot:
## I'm submitting a...

```
[x] Bug report
[ ] Content update
[ ] Process update (build, deployment, ... )
```
## Type of bug / changes
Hi, I can't start my fork with cypress documentation because of the error:
```
65601 verbose stack Error: eslint-plugin-cypress-dev@3.0.2 postinstall: `node ./deprecated --renamed @cypress/eslint-plugin-dev --message 'you may also need to rename cypress-dev -> @cypress/dev in your .eslintrc'`
65601 verbose stack Exit status 1
```
I've tried to:
- delete node_modules and `npm i` again
- update npm
- install only eslint-plugin-cypress-dev

Nothing helps and I have to idea how to solve it. Could you help please?
You can get the branch here, all changes are made: [cypress docs fork](https://github.com/viraxslot/cypress-documentation)

Log:
[cypress-docs-installation.log](https://github.com/cypress-io/cypress-documentation/files/4000842/cypress-docs-installation.log)

**Update:** I can't start origin documentation as well from `develop` branch
**Update 2:** I've changed node/npm version as written in contributing.md. Same result.
$ node -v; npm -v
v8.6.0
5.10.0

jennifer-shehane:
I do also receive this warning, but this does not exit the install process for me. I've opened a PR to fix the deprecation warning https://github.com/cypress-io/cypress-documentation/pull/2342

PR #2338, opened 2019-12-27
[RU]: menu and why-cypress page are translated to russian

viraxslot:
Added translation to Russian:
- menu
- why-cypress.md

@jennifer-shehane I have a couple of questions :)
1. Please see this [issue](https://github.com/cypress-io/cypress-documentation/issues/2333). I suppose we need to change "eslint-plugin-cypress-dev" dependency to "@cypress/eslint-plugin-dev", because the first one is deprecated. See [in origin repo of this plugin](https://github.com/cypress-io/eslint-plugin-dev#usage). I've fixed locally, but I'm not 100% sure.

2. Despite of all passed e2e tests I have problem in the side menu. Could you please help me how to fix absence of sections selection? For portuguese everything is ok, but russian page has no green mark (please see screenshots below).

3. Yeah, and the side menu is not wide enough for "Who uses Cypress?" phrase in Russian. Could we do something with it?

image
image


source/ru/guides/overview/why-cypress.md
```diff
@@ -0,0 +1,97 @@
+---
+title: Почему Cypress?
+---
+
+{% note info %}
+# {% fa fa-graduation-cap %} Что вы узнаете
+
+- Что такое Cypress и почему вам нужно его использовать
```
bahmutov:
```suggestion
- Что такое Cypress и почему Вам нужно его использовать
```

themes/cypress/languages/ru.yml
```diff
@@ -0,0 +1,217 @@
...
+index:
+  get-started: Быстрый старт
+
+page:
+  contents: Оглавление
+  back-to-top: Наверх
+  improve: Улучшить этот документ
+  prev: Назад
+  next: Вперёд
+  last-updated: "Последнее обновление: %s"
```
bahmutov: Hmm, why "%s"?

viraxslot: I don't know, copied from here https://github.com/cypress-io/cypress-documentation/blob/b8a035ca4b4d1e9deaa7721216865330d7cccd96/themes/cypress/languages/en.yml#L18

viraxslot: @bahmutov it's used for format string, for example here:
https://github.com/cypress-io/cypress-documentation/blob/b8a035ca4b4d1e9deaa7721216865330d7cccd96/themes/cypress/layout/api-toc.swig#L22

Hexo documentation says that "There is support for the printf format in language files": https://hexo.io/docs/internationalization#Language-Files

So it allows to convert date of the last update to string and use in the template :)

bahmutov: interesting, I did not notice that this could be customized per language

PR #2352, opened 2020-01-06
Update Mochawesome commands in reporters documentation

brettz9:
- `cypress-multi-reporters` config file expects camel-casing of reporter name (i.e., `mochaJunitReporterReporterOptions`) in place of bare `reporterOptions`
- `mochawesome-merge` now only accepts directory as main argument
- `mochawesome-report-generator` only has the binary `marge`.

### DIFF
--- a/source/guides/tooling/reporters.md
+++ b/source/guides/tooling/reporters.md
@@ -223,7 +223,7 @@ cypress run --reporter mochawesome \
 Our run will generate files `cypress/results/mochawesome.json, cypress/results/mochawesome_001.json, ...`. Then we can combine them using the {% url 'mochawesome-merge' https://github.com/antontelesh/mochawesome-merge %} utility.
 
 ```shell
-npx mochawesome-merge cypress/results > mochawesome.json
+npx mochawesome-merge "cypress/results/*.json" > mochawesome.json
 ```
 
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
@bahmutov can you take a look at these changes?

brettz9:
Regarding my 3 changes:

- `cypress-multi-reporters` config file expects camel-casing of reporter name (i.e., `mochaJunitReporterReporterOptions`) in place of bare `reporterOptions`

You don't have such an example in that repo, but as per [these lines](https://github.com/YOU54F/cypress-multi-reporters/blame/master/README.md#L183-L186) and in my local testing, the `reporterOptions` should indeed be replaced.

- `mochawesome-merge` now only accepts directory as main argument

At https://github.com/cypress-io/cypress-example-reporters/blob/master/package.json#L12 , you are now accepting the glob as main argument as well (I shouldn't have said "directory".)

But I do see this should be "cypress/results/*", so I have fixed this. My apologies for that.

- `mochawesome-report-generator` only has the binary `marge`.

At https://github.com/cypress-io/cypress-example-reporters/blob/master/package.json#L13 , you are also now using `marge`. Do you want me to add `--reportFilename index.html` too? I am using this in a project, svgedit, with no need for that, and you didn't have such an argument previously either.

bahmutov:
Thank you @brettz9 this is great, I will update the branch and when the tests pass will merge it.

PR #2402, opened 2020-01-16
update ci deps based off of docker image

jennifer-shehane:
I was updating the dependency list from this docker image: https://github.com/cypress-io/cypress-docker-images/blob/master/base/12.13.0/Dockerfile#L3 (mentioned from this issue https://github.com/cypress-io/cypress/issues/5965)

Is this correct? I would prefer a better source of truth for this because I feel like this will get out of date quickly and be forgotten.


jennifer-shehane:
@bahmutov Yeah, it's a bit hard to add a link to the image though because the image will update each release or Node update right?

PR #2407, opened 2020-01-16
Document process profiler

flotwig:
For https://github.com/cypress-io/cypress/pull/6171

### DIFF
--- a/source/guides/guides/debugging.md
+++ b/source/guides/guides/debugging.md
@@ -251,9 +251,6 @@ DEBUG=cypress:* cypress run
 
 ```shell
 set DEBUG=cypress:*
-```
-
-```shell
 cypress run
 ```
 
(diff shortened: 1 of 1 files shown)


source/guides/guides/debugging.md
```diff
@@ -309,6 +306,29 @@ cy.now('task', 123)
...
+**On Windows:**
+
+```shell
+set DEBUG=cypress:server:util:process_profiler
+cypress ...
+```
+
+In the resulting output, processes are grouped by the first word of their name.
+
+By default, the interval at which process information is collected and summarized is once every ten seconds. You can override this by setting the `CYPRESS_PROCESS_PROFILER_INTERVAL` environment variable to the desired interval in milliseconds.
```
jennifer-shehane:
```suggestion
By default, the interval at which process information is collected and summarized is once every 10 seconds. You can override this by setting the `CYPRESS_PROCESS_PROFILER_INTERVAL` environment variable to the desired interval in milliseconds.
```

jennifer-shehane:
Rebasing this to 3.8.3-release branch since that is the targeted release.

PR #2413, opened 2020-01-21
Update error docs around renderer crashes

flotwig:
For https://github.com/cypress-io/cypress/pull/6189


source/guides/references/error-messages.md
```diff
@@ -529,15 +529,31 @@ It's possible to enable debugging these scripts by adding the `crossorigin` attr
...
+- The browser was exited manually, by clicking the "Quit" button or otherwise
+- Your test suite or website code is starving the browser for resources, such as with an infinite loop
+- Tests are running in a memory-starved environment
+- The browser is testing a memory-heavy application
+- Tests are running within Docker (there is an easy fix for this: see {% issue 350 'this thread' %})
+- There are problems with the GPU / GPU drivers
+- There is a bug in the browser <!-- TODO: link to Firefox bug here? https://github.com/cypress-io/cypress/issues/6187-->
+- There is a memory leak in Cypress
+
+<!-- TODO: worth adding image of new error?
```
jennifer-shehane: @flotwig This comment broke the rendering of the page (likely because it has a tag in a comment which hexo doesn't handle well)- so that none of the errors below it display on the page. Please make sure to run the docs locally to double check these things.

flotwig: Believe it or not, I did run the docs locally while making these changes :smile: I didn't notice that this was happening... that's pretty lame. Sometimes it feels like hexo has more bugs than features.

jennifer-shehane: Yes, but I think a lot of it is due to our being stuck on very old versions that no one has dedicated time to update.

jennifer-shehane:
This should be good to go in and is not dependent on the change in the Test Runner to go in first. 👍

PR #2436, opened 2020-01-28
Make Network Stubbing Notice More Prominent

seancrater:
While writing some E2E tests, I ran into hardship understanding how to await responses from fetch requests via Cypress. @bahmutov pointed out that Cypress doesn't support this and I probably should've noticed it since I'd been staring at a docs page with a warning about exactly this.

This PR is to make the network stubbing banner more prominent by doing the following:
- Add a siren emoji to grab reader's attention
- Bold text that explains Fetch requests aren't intercepted
- Change note to "danger" type

image


seancrater:
Thank you for the rapid review @jennifer-shehane, let me know if there's anything else you'd like added in, happy to help any way possible

Issue #495, opened 2018-04-06
Document how "invoke" automatically waits for promises to resolve


bahmutov:
If invoked function returns a promise, `cy.invoke` automatically waits for it - need to document it!

See https://github.com/cypress-io/cypress-example-recipes/issues/133 and example in commit https://github.com/cypress-io/cypress-example-recipes/commit/b5629b3cb6d76e8d13a2f8597533a696cdacd105

8BitJonny:
With the branch merged and the updated documentation online, this issue can be finally closed @bahmutov @jennifer-shehane right?

jennifer-shehane:
Yes, thanks!

PR #2448, opened 2020-01-30
Documented that cy.invoke will wait for promises

8BitJonny:
# Documented that cy.invoke will wait for promises
As mentioned in this issue https://github.com/cypress-io/cypress-documentation/issues/495 the `cy.invoke` function will await if the invoked function returns a promise but it is not yet documented. 
This PR expands the documentation to reflect exactly this behaviour

**Issue: https://github.com/cypress-io/cypress-documentation/issues/495**
**First seen in: https://github.com/cypress-io/cypress-example-recipes/issues/133**

### DIFF
--- a/source/api/commands/invoke.md
+++ b/source/api/commands/invoke.md
@@ -136,7 +136,50 @@ cy.wrap([reverse, double]).invoke(1, 4).should('eq', 16)
 
 In this example we invoke a Vuex action that returns a Promise. `cy.invoke` will wait until the Promise resolves after the specified delay which we pass as an argument and only then will continue executing. 
 
+A minimal Vuex store:
 ```javascript
+function randomId () {
+  return Math.random().toString().substr(2, 10)
+}
+
+const store = new Vuex.Store({
+  state: {
+    todos: []
+  },
+  mutations: {
+    ADD_TODO (state, todoObject) {
+      state.todos.push(todoObject)
+    }
+  },
+  actions: {
+    // example promise-returning action
+    addTodoAfterDelay ({ commit }, { milliseconds, title }) {
+      return new Promise((resolve) => {
+        setTimeout(() => {
+          const todo = {
+            title,
+            completed: false,
+            id: randomId(),
+          }
+          commit('ADD_TODO', todo)
+          resolve()
+        }, milliseconds)
+      })
+    },
+  },
+})
+
+```
+
+The Cypress Test with `cy.inkoke()` awaiting the promise:
+```javascript
+// For this command to work you have to expose your vuex store under window.store for cypress to be able to access it
+// e.g. like this
+// if (isCypress) {
+//    window.store = store
+// }
+const getVuex = () => cy.window({log: false}).its('store')
+
 getVuex().invoke('dispatch', 'addTodoAfterDelay', {
   milliseconds: 2000,
   title: 'async task'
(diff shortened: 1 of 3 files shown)


bahmutov:
Also, @8BitJonny I love the explanation, could you fix the text lint error shown on CI/

8BitJonny:
Thanks for the quick feedback @bahmutov ! I'm happy to be able to contribute to such an awesome testing framework. I'll do the changes right away

source/api/commands/invoke.md
```diff
@@ -132,6 +132,29 @@ const double = (n) => n * n
...
+})
+
+// log message appears after 2 seconds
+cy.log('after invoke')
+
+// assert UI
+getTodoItems().should('have.length', 1).first().contains('async task')
+```
+
+`getVuex()` in this example is a function that returns the Vuex data store wrapped inside Cypress object to allow E2E tests to access the application state. For that the application saved itself as "window.app".
```
jennifer-shehane: I would prefer to have a full example provided. We don't have examples that include hidden code - for example, what is `getVuex`, what is `getTodoItems`?. 

We want everything to be fully copy / pastable to run and if a user copy/pastes this in their code it has zero functionality to them and will error.

I would recommend providing an example of a complete async function that can be included and run by copy/pasting. 

If you want to provide a full recipe in a Vue app, it should go in cypress-example-recipes and linked (as it is at the bottom), but a partial should not be included directly in the docs.

source/api/commands/invoke.md
```diff
@@ -2,7 +2,7 @@
 title: invoke
 ---
 
-Invoke a function on the previously yielded subject.
+Invoke a function on the previously yielded subject. If the invoked function returns a promise, it will automatically wait for it until it resolves.
```
jennifer-shehane: I wouldn't include this in the first line of the description since it's a detail of how it works in a very specific circumstance. This type of detail about 'wait' and 'retry' usually goes within the #Rules section here: https://docs.cypress.io/api/commands/invoke.html#Rules

source/api/commands/invoke.md
8BitJonny: Good point 👍

source/api/commands/invoke.md
```diff
@@ -132,6 +132,70 @@ const double = (n) => n * n
 cy.wrap([reverse, double]).invoke(1, 4).should('eq', 16)
 ```
 
+## Invoking an async function
+
+In this example we invoke a Vuex action that returns a Promise. `cy.invoke` will wait until the Promise resolves after the specified delay which we pass as an argument and only then will continue executing.
```
jennifer-shehane:
```suggestion
In this example we invoke a Vuex action that returns a Promise. `.invoke()` will wait until the Promise resolves after the specified delay which we pass as an argument and only then will continue executing. 
```

source/api/commands/invoke.md
```diff
@@ -132,6 +132,70 @@ const double = (n) => n * n
...
+getVuex().invoke('dispatch', 'addTodoAfterDelay', {
+  milliseconds: 2000,
+  title: 'async task'
+})
+
+// log message appears after 2 seconds
+cy.log('after invoke')
+
+// assert UI
+getTodoItems().should('have.length', 1).first().contains('async task')
```
jennifer-shehane: getTodoItems() is not defined in this example. Can we have an example that does not rely on Vue? Because this is still not copy/pastable. You'd have to require Vuex here. Just a function that resolves a Promise?

If not, I would just link to the example and remove the example inline.

8BitJonny: You are right.
I found it quite hard but I finally came up with a good example that doesn't contain an empty Promise that just resolves after a timeout.

bahmutov:
can we merge it @jennifer-shehane

PR #2459, opened 2020-02-03
Adding cypress-iframe to custom command plugins

kgroat:
This PR adds `cypress-iframe` to the list of custom command plugins.  This plugin allows users to interact with iframes, as well as with the elements inside of them.

This plugin was inspired by this issue in the `cypress` repository:
https://github.com/cypress-io/cypress/issues/136

### DIFF
--- a/source/_data/plugins.yml
+++ b/source/_data/plugins.yml
@@ -254,6 +254,11 @@
       link: https://github.com/javierbrea/cypress-localstorage-commands
       keywords: [commands, localstorage, persistence]
 
+    - name: cypress-iframe
+      description: Custom commands for interacting with iframes
+      link:https://gitlab.com/kgroat/cypress-iframe
+      keywords: [commands, iframe]
+
 - name: Extending other testing frameworks
   plugins:
     - name: cyphell
--- a/source/_data/plugins.yml
+++ b/source/_data/plugins.yml
@@ -256,7 +256,7 @@
 
     - name: cypress-iframe
       description: Custom commands for interacting with iframes
-      link:https://gitlab.com/kgroat/cypress-iframe
+      link: https://gitlab.com/kgroat/cypress-iframe
       keywords: [commands, iframe]
 
 - name: Extending other testing frameworks


source/_data/plugins.yml
```diff
@@ -254,6 +254,11 @@
       link: https://github.com/javierbrea/cypress-localstorage-commands
       keywords: [commands, localstorage, persistence]
 
+    - name: cypress-iframe
+      description: Custom commands for interacting with iframes
+      link:https://gitlab.com/kgroat/cypress-iframe
```
jennifer-shehane:
```suggestion
      link: https://gitlab.com/kgroat/cypress-iframe
```

jennifer-shehane (dismissed):
We run a quick smoke test on plugins to make sure they're working minimally as described. 

I ran into an issue running the plugin with the `frameLoaded` command when the url did not match: https://gitlab.com/kgroat/cypress-iframe/issues/1

The `.iframe` and `.enter` commands are really cool though. Would recommend to users after this gets resolved.

bahmutov:
@kgroat I have tried your plugin and added it to our recipes https://github.com/cypress-io/cypress-example-recipes/pull/416 - works great, love it

kgroat:
Glad you like it @bahmutov

bahmutov (approved):
looks good

jennifer-shehane:
Hey @kgroat, were you able to look at the error with the frameLoaded command?

JessicaSachs (approved):
Hey, @kgroat this is looking good. I added additional details to the issue @jennifer-shehane posted. I think it should be fixed, but the plugin seems to be working properly -- it's just lacking a good failure message when you use `{ url }` and fail to find a match.

I did find a race condition where if you add an iframe behind a setTimeout of < ~100ms, the test fails. Timeouts longer than this and static iframes aren't a problem, so I don't think this bug makes the plugin unusable. I'll write something up and post it on GitLab for you.

We accept plugins that meet the following criteria:

- [x] 🚀 Works with the latest major version of Cypress
- [x] 🛠 Plugin purpose is clearly documented (in a README or docs website)
- [x] 📄 Well-written documentation - A great example ([link](https://github.com/Fredx87/cypress-keycloak-commands#cypress-keycloak-commands))
- [x] 🧪 Tested using Cypress - tests using Cypress can act as both example usage _and_ test coverage
- [x] 👷‍♀️ CI pipeline that's passing (CircleCI and [Github Actions](https://github.com/cypress-io/cypress-tutorial-build-todo/blob/master/.github/workflows) are both free for Open Source)

Also, thank you for using GitLab's CI pipeline! I'm going to use that as an example for future plugin authors.

PR #2462, opened 2020-02-04
Add new blog

mncharlton:
Adding a new blog post, this one is a follow up to my post from August (already linked from this page), now that we've been using Cypress for a few months!

Let me know if I've gone about this in the wrong way! Thanks!


mncharlton:
Sorry @jennifer-shehane, I missed that! Just replaced the image for one that's the correct size - let me know if there's anything else that needs changing. Thanks!

PR #2465, opened 2020-02-05
Add docs related to cy.$$

sainthkh:
Related to cypress-io/cypress#2027, cypress-io/cypress#6327


source/api/utilities/$.md
```diff
@@ -50,3 +50,31 @@ cy.wrap($li)
...
   .should('have.class', 'active')
 ```
+
+# Notes
+
+## Cypress.$ vs. cy.$$
+
+You can also query DOM elements with `cy.$$`. But `Cypress.$` and `cy.$$` are different. 
+
+`Cypress.$` refers to `jQuery` function itself. You can do anything with `Cypress.$` if you can do it with `jQuery` function. So, the both of them below works fine.
```
jennifer-shehane:
```suggestion
`Cypress.$` refers to the `jQuery` function itself. You can do anything with `Cypress.$` if you can do it with the `jQuery` function. So, both of the examples below work.
```

source/api/utilities/$.md
```diff
@@ -50,3 +50,31 @@ cy.wrap($li)
...
+}) // works
+```
+
+```js
+$.each([1, 2, 3], (index, value) => {
+  expect(index).to.eq(value)
+}) // also works
+```
+
+But `cy.$$` is a wrapper of `jQuery.fn.init` function. In other words, you can only query DOM elements with `cy.$$`. Because of that, the jQuery utility functions like `jQuery.each`, `jQuery.grep` don't work with `cy.$$`.
```
jennifer-shehane:
```suggestion
But `cy.$$` is a wrapper of the `jQuery.fn.init` function. In other words, you can only query DOM elements with `cy.$$`. Because of that, the jQuery utility functions like `jQuery.each`, `jQuery.grep` don't work with `cy.$$`.
```

source/faq/questions/using-cypress-faq.md
```diff
@@ -600,3 +600,30 @@ cy.get('@consoleLog').should('be.calledWith', 'Hello World!')
 ```
 
 Also, check out our {% url 'Stubbing `console` Receipe' recipes#Stubbing-and-spying %}.
+
+## {% fa fa-angle-right %} How do I use special characters with cy.get?
+
+[According to the CSS spec](https://www.w3.org/TR/html50/dom.html#the-id-attribute), special characters like `/`, `.` are valid characters for ids.
```
jennifer-shehane: We use special url tags that check each url to ensure there aren't any dead links.

```suggestion
{% url "According to the CSS spec" https://www.w3.org/TR/html50/dom.html#the-id-attribute %}, special characters like `/`, `.` are valid characters for ids. 
```

source/faq/questions/using-cypress-faq.md
```diff
@@ -600,3 +600,30 @@ cy.get('@consoleLog').should('be.calledWith', 'Hello World!')
 ```
 
 Also, check out our {% url 'Stubbing `console` Receipe' recipes#Stubbing-and-spying %}.
+
+## {% fa fa-angle-right %} How do I use special characters with cy.get?
+
+[According to the CSS spec](https://www.w3.org/TR/html50/dom.html#the-id-attribute), special characters like `/`, `.` are valid characters for ids. 
+
+To test elements with those characters in ids, they need to be escaped with [`CSS.escape`](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape) or [`Cypress.$.escapeSelector`](https://api.jquery.com/jQuery.escapeSelector/).
```
jennifer-shehane:
```suggestion
To test elements with those characters in ids, they need to be escaped with {% url "`CSS.escape`" https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape %} or {% url "`Cypress.$.escapeSelector`" https://api.jquery.com/jQuery.escapeSelector/ %}.
```

source/faq/questions/using-cypress-faq.md
```diff
@@ -600,3 +600,30 @@ cy.get('@consoleLog').should('be.calledWith', 'Hello World!')
...
+<html lang="en">
+<body>
+  <div id="Configuration/Setup/TextField.id">Hello World</div>
+</body>
+</html>
+```
+
+```js
+it('test', () => {
+  cy.visit('fixtures/issue-2027.html')
```
jennifer-shehane:
```suggestion
  cy.visit('index.html')
```

source/faq/questions/using-cypress-faq.md
```diff
@@ -600,3 +600,30 @@ cy.get('@consoleLog').should('be.calledWith', 'Hello World!')
...
+  cy.get(`#${CSS.escape('Configuration/Setup/TextField.id')}`).contains('Hello World')
+  cy.get(`#${Cypress.$.escapeSelector('Configuration/Setup/TextField.id')}`).contains('Hello World')
+})
+```
+
+Note that `cy.$$.escapeSelector()` doesn't work. `cy.$$` doesn't refer to `jQuery`. It only queries DOM.
```
jennifer-shehane:
```suggestion
Note that `cy.$$.escapeSelector()` doesn't work. `cy.$$` doesn't refer to `jQuery`. It only queries DOM. {% url "Learn more about why" $#Notes %}
```

jennifer-shehane:
Actually, this does require merging in the other PR. This can go into develop now.

PR #2499, opened 2020-02-12
Add 'uncheckability' requirements that removes 'radio'

### LINKED ISSUE #2498: .uncheck() does not work with type="radio", inconsistent with docs
### Current behavior:

Triggering `.uncheck()` on a `type="radio"` element throws the following error:

Screen Shot 2020-02-12 at 10 46 54
 
Fair enough, but the [documentation](https://docs.cypress.io/api/commands/uncheck.html#Rules) for `.uncheck()` states the following under the `Rules` section:

> .uncheck() requires the element to have type **checkbox or radio.**

### Desired behavior:

`.uncheck()` works for radio types. 

Although I have a suspicion that it is an already deprecated feature for some reason.

Would like a way to unselect a radio button for situations testing required fields. Right now I am filling a form in the `beforeEach` hook, and would like consistent ways to clear/unselect things.

### Test code to reproduce

```html
<label data-test-id={testID(someLabel)}>
<input type="radio" ... >
```

```js
cy.get(testID(someLabel))
  .children('input') // this is the radio element
  .uncheck();
````

### Versions

- `Cypress 4.0.1`
- `macOS Mojave 10.14.6`
- `Chrome 80`

jennifer-shehane:
- close #2498 
- also remove list of requirements and timeouts from TAGS.md as this
has and will go stale.

### DIFF
--- a/TAGS.md
+++ b/TAGS.md
@@ -313,32 +313,6 @@ Describe a command's requirements.
 </ul>
 ```
 
-The following requirements are supported:
-
-- `none`
-- `parent`
-- `child`
-- `dom`
-- `dual`
-- `dual_existence`
-- `dual_existence_single_dom`
-- `clearability`
-- `blurability`
-- `clearability`
-- `focusability`
-- `checkability`
-- `selectability`
-- `scrollability`
-- `submitability`
-- `spread`
-- `exec`
-- `task`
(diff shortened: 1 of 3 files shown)


MullerEsposito:
That functionality would be useful. 🙂

PR #2507, opened 2020-02-14
Add links for browser launch options

jennifer-shehane:
- Add more realistic examples
- Remove examples that make no sense to ever use in Cypress.
- Honestly, I couldn't get a single browser launch preference to take effect for Chrome or Firefox, so I don't know if this even works or I'm doing it wrong. There is no real example provided here.

### DIFF
--- a/source/api/plugins/browser-launch-api.md
+++ b/source/api/plugins/browser-launch-api.md
@@ -49,20 +49,29 @@ The returned `launchOptions` object will become the new launch options for the b
 
 ### Modify browser launch arguments:
 
+Here are args available for the currently supported browsers:
+
+* {% url 'Chromium-based browsers' "https://peter.sh/experiments/chromium-command-line-switches/" %}
+* {% url 'Firefox' "https://developer.mozilla.org/docs/Mozilla/Command_Line_Options" %}
+
 ```js
 module.exports = (on, config) => {
   on('before:browser:launch', (browser = {}, launchOptions) => {
+    // `args` is an array of all the arguments that will
+    // be passed to browsers when it launches
+    console.log(launchOptions.args) // print all current args
+
     if (browser.family === 'chromium' && browser.name !== 'electron') {
-      // `args` is an array of all the arguments that will
-      //  be passed to Chromium-based browsers when it launchers
-      launchOptions.args.push('--start-fullscreen')
+      // auto open devtools
+      launchOptions.args.push('--auto-open-devtools-for-tabs')
 
       // whatever you return here becomes the launchOptions
       return launchOptions
     }
 
     if (browser.family === 'firefox') {
-      launchOptions.args.push('-some-cli-argument')
+      // auto open devtools
+      launchOptions.args.push('-devtools')
 
       return launchOptions
     }
@@ -86,6 +95,12 @@ module.exports = (on, config) => {
 
 ### Changing browser preferences:
 
+Here are preferences available for the currently supported browsers:
+
+* {% url 'Chromium-based browsers' "https://src.chromium.org/viewvc/chrome/trunk/src/chrome/common/pref_names.cc?view=markup" %}
+* {% url 'Electron' "https://github.com/electron/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions" %}
+* Firefox: visit `about:config` URL within your Firefox browser to see all available preferences.
+
 ```js
 module.exports = (on, config) => {
   on('before:browser:launch', (browser, launchOptions) => {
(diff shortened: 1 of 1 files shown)


source/api/plugins/browser-launch-api.md
```diff
@@ -96,22 +111,18 @@ module.exports = (on, config) => {
       // for example, to set `somePreference: true` in Preferences:
       launchOptions.preferences.default.somePreference = true
 
-      // more information about Chromium preferences can be found here:
-      // https://www.chromium.org/developers/design-documents/preferences
```
jennifer-shehane: This is a very confusing doc, and I have a hard time understanding how it is relevant to sending preferences. Although I cannot figure out the signature for sending any perfs in https://src.chromium.org/viewvc/chrome/trunk/src/chrome/common/pref_names.cc?view=markup

source/api/plugins/browser-launch-api.md
```diff
@@ -49,20 +49,29 @@ The returned `launchOptions` object will become the new launch options for the b
...
+      launchOptions.args.push('--auto-open-devtools-for-tabs')
 
       // whatever you return here becomes the launchOptions
       return launchOptions
     }
 
     if (browser.family === 'firefox') {
-      launchOptions.args.push('-some-cli-argument')
+      // auto open devtools
+      launchOptions.args.push('-devtools')
```
jennifer-shehane: @Bkucera Is there any actual usecase for `launchOptions.args` in Electron? Mostly everything I could see pushing in is now in `preferences`

source/api/plugins/browser-launch-api.md
```diff
@@ -86,6 +95,12 @@ module.exports = (on, config) => {
 
 ### Changing browser preferences:
```
jennifer-shehane: @Bkucera This option should have closed this issue: https://github.com/cypress-io/cypress/issues/3671 But I can't reasonably close this when I can't give them a single example of this working.

source/api/plugins/browser-launch-api.md
Bkucera: there is no use case for `args` in electron. we just ignore them

source/api/plugins/browser-launch-api.md
Bkucera: agreed, lets pick the most requested use-case solved by custom preferences and put it as an example here

Bkucera: @flotwig would that be setting downloads folder location?

source/api/plugins/browser-launch-api.md
```diff
@@ -96,22 +111,18 @@ module.exports = (on, config) => {
...
-      launchOptions.preferences['browser.startup.homepage'] = 'https://cypress.io'
+      launchOptions.preferences['some.preference'] = true
 
       return launchOptions
     }
 
     if (browser.name === 'electron') {
-      // launchOptions.preferences is a `BrowserWindow` `options` object:
-      // https://electronjs.org/docs/api/browser-window#new-browserwindowoptions
+      // launchOptions.preferences is a `BrowserWindow` `options` object
```
Bkucera: we should add back the `browserwindowoptions` api doc link

Bkucera: we should add back the `browserwindowoptions` api doc link

Bkucera: ah i see it above, NVM

source/api/plugins/browser-launch-api.md
flotwig: @Bkucera yeah, that's a good one if it can be solved via preferences alone

source/api/plugins/browser-launch-api.md
flotwig: Chromium preferences are confusing in their nature, would it be less confusing if we clarified these are for advanced use only?

There's not really a simple way to explain it. In Chromium, preferences are just internal state saved between launches, so that documentation is actually the simplest docs entrypoint I could find regarding them.

source/api/plugins/browser-launch-api.md
```diff
@@ -86,6 +95,12 @@ module.exports = (on, config) => {
 
 ### Changing browser preferences:
 
+Here are preferences available for the currently supported browsers:
+
+* {% url 'Chromium-based browsers' "https://src.chromium.org/viewvc/chrome/trunk/src/chrome/common/pref_names.cc?view=markup" %}
```
flotwig: nice find

source/api/plugins/browser-launch-api.md
jennifer-shehane: I added a note saying this to top of browser launch args section.

source/api/plugins/browser-launch-api.md
jennifer-shehane: I added a warning note at top of browser preferences saying:

>Browser preferences are available for advanced usage and should not be required in most cases.

jennifer-shehane:
I'm going to merge this in as this is an improvement as is. Would love an example added for https://github.com/cypress-io/cypress/issues/3671 on how to set the download destination within Chrome preferences, which I can then add to the doc later.

source/api/plugins/browser-launch-api.md
Bkucera: @jennifer-shehane  I've made an example changing the download directory here https://github.com/cypress-io/cypress/pull/6494/files#diff-d1e9c5129d8076065e3569141b79b037 I think we can paste that as an example here

Bkucera:
@jennifer-shehane  I've made an example changing the download directory here https://github.com/cypress-io/cypress/pull/6494/files#diff-d1e9c5129d8076065e3569141b79b037 I think we can paste that as an example here

PR #2530, opened 2020-02-20
Add cypress-keycloak-commands to authentication plugins

Fredx87:
This PR adds `cypress-keycloak-commands` to the list of custom authentication plugins. This plugin allows to authenticate users using the [Keycloak](https://www.keycloak.org/) Identity Provider


JessicaSachs (approved):
👋 Hello and thank you for your contribution! We've reviewed your plugin and are happy to add it to the plugins list.

- [x] 🚀 Works with the latest major version of Cypress
- [x] 🛠 Plugin purpose is clearly documented (in a README or docs website)
- [x] 📄 Well-written documentation 
- [x] 🧪 Tested using Cypress - tests using Cypress can act as both example usage _and_ test coverage
- [x] 👷‍♀️ CI pipeline that's passing (CircleCI and [Github Actions](https://github.com/cypress-io/cypress-tutorial-build-todo/blob/master/.github/workflows) are both free for Open Source)

Your docs are great and we'll be pointing future plugin authors to them as an example.

PR #2534, opened 2020-02-21
Document extra config values added to plugins config

chrisbreiding:
TODO:

- [x] Merge in https://github.com/cypress-io/cypress/pull/6317 first

Documents change made in https://github.com/cypress-io/cypress/pull/6317


source/api/plugins/writing-a-plugin.md
```diff
@@ -59,6 +59,11 @@ This configuration contains all of the values that get passed into the browser f
 
 {% url 'For a comprehensive list of all configuration values look here.' https://github.com/cypress-io/cypress/blob/master/packages/server/lib/config.coffee %}
 
+It also includes the following extra values that are not part of the standard configuration. Note that these values cannot be modified from the plugins file.
```
jennifer-shehane: I would probably call this out more. 

```suggestion
It also includes the following extra values that are not part of the standard configuration. 

{% note warning %}
These values are read only and cannot be modified from the plugins file.
{% endnote %}
```

chrisbreiding: The only problem is that there are more lines below this that refer to all the values, so it might be confusing. We'd need to restructure more lines or at least make it clear in the warning that we're only referring to `configFile` and `projectRoot`.

jennifer-shehane:
Close in favor of https://github.com/cypress-io/cypress-documentation/commit/f12d5a13ac60a837171ecbf0ee16f30f9ed1eaa8

PR #2560, opened 2020-03-02
Add more examples of mixing async/sync commands and why its bad

jennifer-shehane:
I have to *very* frequently point out to users that they cannot mix async and sync code. I've been linking them to the `Commands are Asynchrnous` section of our docs, but I've found this lacking in *actual* examples of mixing async/sync code, why this is not a good idea, and how to fix them. 

This adds examples of mixing async/sync code, why it doesn't work, and suggested code to fix it.

### DIFF
--- a/source/guides/core-concepts/introduction-to-cypress.md
+++ b/source/guides/core-concepts/introduction-to-cypress.md
@@ -400,11 +400,13 @@ it('does not work as we expect', () => {
   cy.get('.awesome-selector')   // Still nothing happening
     .click()                    // Nope, nothing
 
-  // Cypress.$ is synchronous
-  // this element only exists after you click the .awesome-selector
-  let el = Cypress.$('.new-el') // evaluates immediately!
+  // Cypress.$ is synchronous, so evaluates immediately
+  // there is no element to find yet because
+  // the cy.visit() was only queued to visit
+  // and did not actually visit the application
+  let el = Cypress.$('.new-el') // evaluates immdeiately as []
 
-  if (el.length) {              // evaluates immediately!
+  if (el.length) {              // evaluates immediately as 0
     cy.get('.another-selector')
   } else {
     // this code will never run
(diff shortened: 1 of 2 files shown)


source/guides/core-concepts/introduction-to-cypress.md
```diff
@@ -385,6 +385,127 @@ it('changes the URL when "awesome" is clicked', function() {
...
+
+```js
+it('does not work as we expect', () => {
+  cy.visit('/my/resource/path') // Nothing happens yet
+
+  cy.get('.awesome-selector')   // Still nothing happening
+    .click()                    // Nope, nothing
+
+  // Cypress.$ is synchronous
+  // this element only exists after you click the .awesome-selector
```
bahmutov: I would explain this a little bit more
"There is no element to find yet - because the page "/my/resource/path" was only scheduled to be visited using `cy.visit()" above. So the iframe is empty - and you will get nothing back!
To really have the element, you should get the element AFTER the `cy.click` clicks on the button, which happens after `cy.visit` completes.

source/guides/core-concepts/introduction-to-cypress.md
```diff
@@ -385,6 +385,127 @@ it('changes the URL when "awesome" is clicked', function() {
...
+
+```js
+it('test', () => {
+  let username = undefined     // evaluates immediately!
+
+  cy.visit('https://app.com') // Nothing happens yet
+  cy.get('.user-name')        // Still, nothing happens yet
+    .then(($el) => {          // Nothing happens yet
+      // this line evaluates after the .then executes
+      username = $el.text()
```
bahmutov: // username has been set correctly

source/guides/core-concepts/introduction-to-cypress.md
```diff
@@ -385,6 +385,127 @@ it('changes the URL when "awesome" is clicked', function() {
...
+    })
+
+  // evaluates immediately!
```
bahmutov: more important this executes BEFORE `.then(($el) => ...` runs, thus it uses `username` that is still `undefined`

source/guides/core-concepts/introduction-to-cypress.md
jennifer-shehane: Yah, oof, it’s hard to tersely explain this stuff 😅

PR #2421, opened 2020-01-22
Add docs for error improvements

chrisbreiding:
Closes #2374 

## TODO:

- [ ] Do not merge until https://github.com/cypress-io/cypress/pull/3930 is merged for specific version

PR #2569, opened 2020-03-03
add 'Run tests directly from IDE' section

mbolotov:
add `Run tests directly from IDE` section
also it may be worth to rename `Intelligent Code Completion` page to something like `Configuring IDE`


jennifer-shehane:
Yes, actually we will be renaming this entire doc to 'IDE Integration' as part of this PR. https://github.com/cypress-io/cypress-documentation/pull/2421

I'm inclined to have this change go into that PR actually since it doesn't make much sense on the Intelligent Code Completion page.

PR #2595, opened 2020-03-09
Updates to `after:screenshot` page

### LINKED ISSUE #2594: Clarify better how setting props in `after:screenshot` actually works
## I'm submitting a...

```
[ ] Bug report
[x] Content update
```

## Type of bug / changes

It's a bit confusing. [I was confused myself](https://github.com/cypress-io/cypress/issues/6648#issuecomment-595690389). I thought that setting the properties in `after:screenshot` would actually mutate the screenshot (change where it was located, change the size, etc), but it's actually meant to only specify any new properties of the screenshot to Cypress. The user is responsible for editing the screenshot (location, size, etc). 

https://docs.cypress.io/api/plugins/after-screenshot-api.html#Usage

jennifer-shehane:
- close #2594
- Add table for details with descriptions of each property
- Make structure of page more closely match other plugin events layout
- Clarify more that the user needs to modify the details of the image
themselves


jennifer-shehane:
@chrisbreiding I'm not totally clear on what a description of `multipart` and `pixelRatio` should be here. Any help?

chrisbreiding:
`multipart` is a boolean that, when `true`, indicates that the screenshot was stitched together from multiple parts, created by taking a screenshot, scrolling down, taking another, and so on until the full page is captured. 

`pixelRatio` is the ratio of the pixels in the screenshots image file to "pixels" displayed by the browser. It should be equivalent to the device pixel ratio. Relevant for retina/high-resolution screens where the screen's resolution is higher than the width x height that the browser reports for its viewport.

Issue #2642, opened 2020-03-19
Add new recipes to docs to match example-recipes repo


CypressCecelia:
## I'm submitting a...

```
[ ] Bug report
[X] Content update
[ ] Process update (build, deployment, ... )
```

## Type of bug / changes

Add links to:

### Fundamentals
Recipe | Description
--- | ---
[Custom browsers](./examples/fundamentals_custom-browsers) | Control which browsers the project can use, or even add a custom browser into the list
[use Chrome Remote Interface](./examples/fundamentals__chrome-remote-debugging) | Use Chrome debugger protocol to trigger hover state and print media style

- [x] Custom browsers
- [x] use Chrome Remote Interface

### Testing the DOM
Recipe | Description
--- | ---
[Waiting for static resource](./examples/testing-dom__wait-for-resource) | Shows how to wait for CSS, image, or any other static resource to load
[CSV load and table test](./examples/testing-dom__csv-table) | Loads CSV file and compares objects against cells in a table
[Evaluate performance metrics](./examples/testing-dom__performance-metrics) | Utilize Cypress to monitor a website
[Root style](./examples/testing-dom__root-style) | Trigger input color change that modifies CSS variable
[Select widgets](./examples/testing-dom__select2) | Working with `<select>` elements and [Select2](https://select2.org/) widgets

- [x] Waiting for static resource
- [x] CSV load and table test
- [x] Evaluate performance metrics
- [x] Root style
- [x] Select widgets

### Preprocessors
Recipe | Description
--- | ---
[Flow with Browserify](./examples/preprocessors__flow-browserify) | Add flow support with browserify

- [x] Flow with Browserify

### Blogs
Recipe | Description
--- | ---
[React DevTools](./examples/blogs__use-react-devtools) | Loads React DevTools Chrome extension automatically
[Expect N assertions](./examples/blogs__assertion-counting) | How to expect a certain number of assertions in a test
[Browser notifications](./examples/blogs__notification) | How to test application that uses [`Notification`](https://developer.mozilla.org/en-US/docs/Web/API/notification)
[Testing iframes](./examples/blogs__iframes) | Accessing elements in 3rd party iframe, spy and stub network calls from iframe

- [x] React DevTools
- [x] Expect N assertions
- [x] Browser notifications
- [x] Testing iframes

### Stubbing and spying
Recipe | Description
--- | ---
[...]

PR #2646, opened 2020-03-20
adds link to workaround for hover command entry

### LINKED ISSUE #2617: Offer workaround for CSS hover or state that there is none
In https://github.com/cypress-io/cypress-documentation/issues/2412 you updated the docs, stating that the trigger() workaround does not affect any CSS triggers, which is great!

Unfortunately, for people in need to test exactly that feature the information ends right there.

It would be great if you could either 

1. Offer a workaround (or a link to a workaround, respectively) or
2. State that there (currently) is no workaround and that such a behaviour is impossible to test with Cypress

Thank you!

CypressCecelia:
links to Chrome remote debugging recipe which is being added in #2642 

https://github.com/cypress-io/cypress-documentation/pull/2645 must merge first for link to make sense. It's not broken, but the recipe referenced isn't there yet :)

Closes #2617

### DIFF
--- a/source/api/commands/hover.md
+++ b/source/api/commands/hover.md
@@ -26,6 +26,8 @@ If the hover behavior depends on a JavaScript event like `mouseover`, you can tr
 Using `.trigger()` will only affect events in JavaScript and will not trigger any effects in CSS. 
 {% endnote %}
 
+As a workaround, check out the {$ url 'recipe leveraging Chrome remote debugging' recipes#Fundamentals} to set pseudo classes like `hover`.
+
 ### Simulating `mouseover` event to get popover to display
 
 ```javascript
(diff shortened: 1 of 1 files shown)


source/api/commands/hover.md
```diff
@@ -26,6 +26,8 @@ If the hover behavior depends on a JavaScript event like `mouseover`, you can tr
 Using `.trigger()` will only affect events in JavaScript and will not trigger any effects in CSS. 
 {% endnote %}
 
+As a workaround, check out the {$ url 'recipe leveraging Chrome remote debugging' recipes#Fundamentals} to set pseudo classes like `hover`.
```
jennifer-shehane: url is not correct formatting. 

```suggestion
As a workaround, check out the {% url 'recipe leveraging Chrome remote debugging' recipes#Fundamentals %} to set pseudo classes like `hover`.
```

CypressCecelia:
🤦🏽‍♀️

PR #2649, opened 2020-03-22
Improves not-valid code warnings

JoeSchr:
Makes them more explicit and also puts some of them into the code sample (the only thing many people actually read)


JoeSchr:
> Thanks. I worry about these invalid code examples, I wish we didn't have them at all tbh, but this is definitely an improvement on the warning.

I actually treasure them, because it's good to see with an example what not to do, because sometime you already have some code in your head, ready to write it, search for some example that looks similar. when you then find something like this, one would see this and say: "wait a minute". let me read on with this chapter... ;)

PR #2658, opened 2020-03-24
Add example for saving variable across domain visits

### LINKED ISSUE #2656: Add example of saving values across cross domain changes
## I'm submitting a...

```
[ ] Bug report
[x] Content update
```

## Type of bug / changes

This example https://github.com/cypress-io/cypress/issues/6562#issuecomment-595042151 maybe in `cy.task()`?? Not really sure.

jennifer-shehane:
- close #2656 

Hopefully self explanatory. Open to suggestions where this is confusing.

### DIFF
--- a/source/api/commands/task.md
+++ b/source/api/commands/task.md
@@ -84,7 +84,7 @@ Option | Default | Description
 
 # Examples
 
-## Command
+## Event
 
 `cy.task()` provides an escape hatch for running arbitrary Node code, so you can take actions necessary for your tests outside of the scope of Cypress. This is great for:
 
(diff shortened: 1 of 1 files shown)


source/api/commands/task.md
```diff
@@ -84,7 +84,7 @@ Option | Default | Description
 
 # Examples
 
-## Command
+## Event
```
bahmutov: hmm, I am not sure what this second level heading means "Event" (or previous "Command"). Like the text that follows it explains typical use cases, right? And then it explains the return value from the event's callback. So maybe the explanation of return value should be moved into the syntax or rules, then we have the heading `Typical use cases` explaining list of things, then we have actual examples below

source/api/commands/task.md
```diff
@@ -176,6 +176,50 @@ module.exports = (on, config) => {
 }
 ```
 
+### Save a variable across non same-origin URL visits
```
bahmutov: this works, ran the code, I will code a recipe just to point people and live tested example. Great stuff, @jennifer-shehane

source/api/commands/task.md
jennifer-shehane: They sections in Examples typically are sectioned by their API signature, so in this case `cy.task(event)` and `cy.task(event, options)` - there's a heading for `Event` and a heading for `Options`. It seems more obvious for other API signatures, but maybe not as obvious for this one cause 'event' is a weird name. But yeah...

Issue #1322, opened 2019-01-18
Document "wrap(promise)" use case


bahmutov:
From https://github.com/cypress-io/cypress/issues/3172#issuecomment-455709571

https://on.cypress.io/wrap

PR #2664, opened 2020-03-25
Makes `wrap(promise)`  more prominent

JoeSchr:
As indicated in #1322 and ff,  it needs to be immediately communicated  to new user that using `.wrap` helps immensely with promises.

just spend a few days ripping my hair out because I didn't see this before

### DIFF
--- a/source/api/commands/wrap.md
+++ b/source/api/commands/wrap.md
@@ -2,7 +2,8 @@
 title: wrap
 ---
 
-Yield the object passed into `.wrap()`.
+- Yield the object passed into `.wrap()`. 
+- `.wrap(promise)` forces the test commands to wait until promise resolves
 
 # Syntax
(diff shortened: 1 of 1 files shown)


JoeSchr:
yeah, I just started with cypress, so I don't feel confident writing that
right now.

but I feel strongly that there should be some indication *at the top of the
page* that there are more uses for `wrap` then just wrapping a "dumb" data
structure. and that it is key to making many promise based approaches work.

I wasted like 2-3 days until I found that out and I must have looked at the
wrap page like 10-15 times while doing that


__

Typed on a mobile, please excuse any typing mistakes or short-spokeness.

jennifer-shehane:
Closing in favor of https://github.com/cypress-io/cypress-documentation/pull/2689

Issue #2667, opened 2020-03-25
Add SSO configuration instructions to Dashboard > Organizations


CypressCecelia:
## I'm submitting a...

```
[ ] Bug report
[X] Content update
[ ] Process update (build, deployment, ... )
```

## Type of bug / changes

Add documentation around how to configure SSO at the organization level for subscribers that have it enabled. 

### Existing Docs to Transfer:

- [ ] Configuring SAML for SSO into the Cypress Dashboard
- [ ] Configuring OKTA for SSO into the Cypress Dashboard

### New Docs to Write:

Will check with Product team.

PR #2669, opened 2020-03-25
adds  SSO  instructions  to dashboard organization section.

CypressCecelia:
Included existing documentation, may need to write more, #2667


CypressCecelia:
@amirrustam  Many of the screenshot images are from the original Doc, and are not consistent in size. We may want to get new screenshots before updating. 🤷‍♀

amirrustam:
@CypressCecelia I'll review after @jennifer-shehane's suggestions are added.

CypressCecelia:
Thanks @jennifer-shehane! Yes, I was copying from the internal wiki entry in markdown. I'll install that extension. 

- [ ] I can make a PR to the `on` package to add a link to the logo.

- [ ] I think it would be better to remove those screenshots entirely until we can get better ones. They look pretty bad, but I wasn't sure if it was just because I have a large screen. I'd rather do no screenshots than bad ones. I'll remove those listed.

jennifer-shehane:
@amirrustam Yeah, I did not verify if any of these instructions are actually accurate 😅

amirrustam:
@jennifer-shehane they come directly from the Google Docs @williebsweet has been sharing with customers, so they've already been pre-vetted.

CypressCecelia:
Added Azure AD documentation and removed image files. Waiting for the updated `on` link for the logo and then it should be ready for review.

CypressCecelia:
https://github.com/cypress-io/cypress-services/issues/2395 has been approved and merged.

CypressCecelia:
Logo on link was released.

CypressCecelia:
@trydionel I based this documentation on the Google docs we were sending to subscribers. Please confirm that the info is still correct before we publish. Thank you!

CypressCecelia:
> * We should add a callout indicating that SSO is a paid addon.

I pushed a commit with the callout. 

> * I feel like this content is placed in the wrong location in the site IA. I expected to see a new "Integrations" guide under the Dashboard dropdown.

For the location, this Integrations section exists within the Organizations page. There is navigation on the right-hand side menu for this section. 

To get a new tab on the left menu we would need to create a new page just for Integrations, which I think may be confusing as we already have a GitHub Integration tab.

PR #2675, opened 2020-03-26
Add example of overwriting type command to mask input

### LINKED ISSUE #1944: Add example on how to not show sensitive data in Command Log
## I'm submitting a...

```
[ ] Bug report
[X] Content update
```

## Type of bug / changes

```js
cy.get('.password').type('password123', { log: false })
```

This example could use a little more love, but you get the idea

```js
Cypress.Commands.overwrite('type', (originalFn, element, text, options) => {
  if (options.sensitive) {
    options._log = Cypress.log({
      $el: element,
      name: 'type',
      message: '*******',
    })
  }

  return originalFn(element, text, options)
})

cy.get('.password')
  .type('password123', { log: false, sensitive: true })
```

jennifer-shehane:
- close #1944 

Screen Shot 2020-03-26 at 1 57 36 PM
Screen Shot 2020-03-26 at 1 57 44 PM

Screen Shot 2020-03-26 at 1 57 53 PM


jennifer-shehane:
@bahmutov There is not an example in kitchensink or recipes. I'll add a link to the blog post. 

My only problem with this example is that it completely wipes out the original logs - so you don't get the nice key events tables, etc. I asked in test-runner channel but never got a response. 

I want all the original logs called and just want to edit one of the values. All I can see is just completely overwriting all the log logic. Can I edit the original Cypress.log from the command?

## Original logs

Screen Shot 2020-03-26 at 1 38 26 PM

## Overwritten logs

Screen Shot 2020-03-26 at 1 39 50 PM

jennifer-shehane:
Oh, also, another failure of this is if you are chaining some assertion off of the secret password - the password will print in the passing / failing assertion text.

PR #2678, opened 2020-03-26
add Happo visual testing plugin

bahmutov:
One more plugin that does a really good job, example repo https://github.com/happo/happo-cypress-example-todomvc

### DIFF
--- a/source/_data/plugins.yml
+++ b/source/_data/plugins.yml
@@ -395,7 +395,7 @@
       keywords: [component, vue, vue.js]
 
 - name: Visual Testing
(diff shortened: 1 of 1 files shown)


source/_data/plugins.yml
```diff
@@ -395,7 +395,7 @@
       keywords: [component, vue, vue.js]
 
 - name: Visual Testing
-  description:
+  description: Visual testing is a great complement to functional testing, read our [full visual testing guide](https://on.cypress.io/visual-testing)
```
jennifer-shehane: Use url tags to verify we are not linking to dead links.

```suggestion
  description: Visual testing is a great complement to functional testing, read our {% url "full visual testing guide" visual-testing %}
```

PR #2747, opened 2020-04-25
add 'Cypress-Pro' plugin link

mbolotov:
'Cypress-Pro' plugin mention


mbolotov:
@JessicaSachs Appreciate any feedback.

source/guides/tooling/IDE-integration.md
```diff
@@ -19,6 +19,7 @@ There are many third-party IDE extensions and plugins to help integrate your IDE
 Compatible with IntelliJ IDEA, AppCode, CLion, GoLand, PhpStorm, PyCharm, Rider, RubyMine, and WebStorm.
 
 - {% url "Intellij-Cypress" https://plugins.jetbrains.com/plugin/13819-intellij-cypress %}: Integrates Cypress under the common Intellij test framework.
+- {% url "Cypress-Pro" https://plugins.jetbrains.com/plugin/13987-cypress-pro %}: An improved version of Intellij-Cypress plugin with debugging from the IDE, advanced autocomplete and other features.
```
bencodezen, line 22: When you say "improved," does this mean that I still need `Intellij-Cypress` with this plugin? Or is this replacing the existing one?

mbolotov: CypressPro is replacing the original Intellij-Cypress plugin. So it has all the functionality form the latter and adds some additional features. 
Do you think this sentence should be rephrased?

bencodezen (changes requested):
Need clarification on language to prevent confusion for IntelliJ users

source/guides/tooling/IDE-integration.md
bencodezen: I think it's confusing because there's two plugins now recommended here for IntelliJ. Can we mark this one as the recommended one somehow or even delete the reference to the original `Intellij-Cypress` plugin?

mbolotov: Okay, I deleted the `Intellij-Cypress` plugin reference.

bencodezen:
@mbolotov Thanks for the update! After reviewing the plugin with the team, there are some issues with the plugin that need to be addressed:

1. Change the logo - The new plugin currently uses the Cypress logo which can give users the impression that this is a plugin supported by the company rather than your plugin
2. Change the name - The name of the plugin `Cypress-Pro` also gives the impression that users are paying Cypress for a service as opposed to you directly

Also, I apologize for the mistake, but I wasn't aware that the new plugin would be a paid service rather than a free one. We should add the original one back as the main one that users are directed to rather than the paid one.

If possible, it would be great to find a new plugin name that still highlights that it's for Cypress while giving credit to the fact that it's your tool rather than the company. Let me know if you have any additional questions, but I hope this helps!

mbolotov:
@bencodezen,
Good, I'll try to address the issues. 
A few questions:
1. Do you think `Cypresso` will be okay for the name of plugin?
2. Should I change the name for the paid version only? So the open sourced version would still have `Cypress` mention in its name

mbolotov:
@bencodezen, could you please answer the questions above?

Shelex:
This is interesting case actually.

IMHO: of course, Cypress has [MIT](https://github.com/cypress-io/cypress/blob/master/LICENSE) license and commercial permission but it will be super strange if link to some paid service will be included as officially supported. As usual paid services provide free trial version or functionally limited package - and in such case it could be included into documentation (in this case - Cypress).  

But I would also like to hear some official position (maybe @bahmutov ) about terms or regulations of plugins development and publicly available checklist for getting your plugin/tool/whatever added to documentation. For now it tells just to submit PR to plugins.yml - [link](https://github.com/cypress-io/cypress-documentation/blob/develop/CONTRIBUTING.md#adding-plugins). Another point - should these rules apply to existing plugins and how?

amirrustam:
Thanks for your input @Shelex. You can refer to @bencodezen for our official position on the plugin review process.

We do not mind linking to other services, and we applaud the developments within Cypress ecosystem by the community. Our primary concern with surfacing projects like this is to ensure that the plugin/project does not cause user confusion and that it is a high quality (well documented, well tested, etc.) offering.

bencodezen:
@mbolotov 

> Do you think Cypresso will be okay for the name of plugin?

After reviewing both versions, it would be best if you could keep the free version named similarly to the paid one so it is consistent on both ends.

Screen Shot 2020-05-18 at 12 07 36 PM

Following a similar method to VS Code extensions, by simply adding a qualifier to what your extension does (e.g., Cypress Helper) would better help to namespace the branding to prevent confusion from any official Cypress plugins that may be released in the future.

Let me know if you have any other questions!

bencodezen (changes requested):
1. Add free version back in to the documentation
2. Change names of both versions to reduce confusion for users

mbolotov:
@bencodezen 
I've changed plugin names and removed the icon.
Could you please review?

bencodezen:
This is great @mbolotov! Once everything passes, I'll be happy to merge this. Thanks so much for being accommodating. Really appreciate it!

bencodezen (approved):

PR #2745, opened 2020-04-24
Add component testing experiment

bahmutov:
- move "Experiments" into its own page, the very last link under "References" section
- add `experimentalComponentTesting` experiment

PR #2701, opened 2020-04-10
docs for use-bundled-mocha-for-reporters

Bkucera:
docs for https://github.com/cypress-io/cypress/pull/6951

PR #2751, opened 2020-04-27
4.5.0 Release

jennifer-shehane:
- 4.5.0 changelog
- updated docs on mocha custom reporter change #2701
- add component testing experimental config #2745


bahmutov:
@jennifer-shehane we specifically moved the experimental page into its own guide, since it will be expanding soon with more info about component testing

jennifer-shehane:
This needs to be reverted / moved out of deployment https://github.com/cypress-io/cypress-documentation/pull/2701

jennifer-shehane:
Moved experiments back to its own page.

source/guides/references/experiments.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+This test would execute in the browser, similar to the full end-to-end test, except there is no website URL:
+
+{% imgTag /img/guides/references/component-test.gif "Example React component test" %}
+
+Feature | Test libraries | Cypress component tests
+--- | --- | ---
+Test runs in real browser | ❌ | ✅
+Uses full mount | ❌ | ✅
```
JessicaSachs, line 42: React Testing Library does a full mount. https://kentcdodds.com/blog/why-i-never-use-shallow-rendering

>  I've expressed this feeling on many occasions and get asked on a regular basis why I feel the way I do about shallow rendering and why React Testing Library will never support shallow rendering.

JessicaSachs: Enzyme is the only one who doesn't do a full mount.. and that's optional.

brian-mann: regardless of the mount - i don't think this is a real differentiator.

it's not so much about full or partial mount, or "whether it runs in a real browser"...

it's really about **seeing your components** and natively interacting with them. we should call to attention that you can see your components just like if you were in dev with a browser open pointing to `localhost`.

we should also call to attention videos and screenshots "just work" out of the box. no need to deal with an artificial DOM. it's the same DOM and the same interaction and rendering engine. there's no translation layer

PR #2802, opened 2020-05-15
Update --config flag options to real wroking examples.

### LINKED ISSUE #2801: Passing an array for testFiles in cli as documented does not work
### Current behavior:
- Passing an array for testFiles and ignoreTestFiles in cli is parsed as a string instead of an array in the cypress runner. This causes the argument to not work.

Command used taken directly from documentation : 
https://docs.cypress.io/guides/references/configuration.html#Command-Line
`cypress open --config watchForFileChanges=false,testFiles=["**/*.js","**/*.coffee"]`

screenshot from Configuration:
Screen Shot 2020-05-09 at 6 34 07 PM

### Desired behavior:
testFiles should be parsed as an array. For example, if i was to set the testFiles in the cypress environment file, it would show up as an array in the configurations as shown below
Screen Shot 2020-05-09 at 7 33 22 PM

### Versions

Tried on Cypress version 4.2.0 and 4.5.0

jennifer-shehane:
- close #2801
- Fixed example in config that was broken
- Update examples to include options that are probably more popularly used.

### DIFF
--- a/source/guides/references/configuration.md
+++ b/source/guides/references/configuration.md
@@ -131,19 +131,21 @@ When {% url 'running Cypress from the Command Line' command-line %} you can pass
 **Examples:**
 
 ```shell
-cypress open --config watchForFileChanges=false,waitForAnimations=false
+cypress open --config pageLoadTimeout=30000,baseUrl=https://myapp.com
 ```
 
 ```shell
-cypress run --config integrationFolder=tests,fixturesFolder=false
+cypress run --config integrationFolder=tests,videoUploadOnPasses=false
 ```
 
 ```shell
-cypress run --record --config viewportWidth=1280,viewportHeight=720
+cypress run --browser firefox --config viewportWidth=1280,viewportHeight=720
 ```
 
+For more complex configuration objects, you may want to consider passing a {% url "JSON.stringified" https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify %} object surrounded by quotes.
+
 ```shell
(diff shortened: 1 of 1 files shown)


source/guides/references/configuration.md
```diff
@@ -131,19 +131,21 @@ When {% url 'running Cypress from the Command Line' command-line %} you can pass
...
-cypress run --config integrationFolder=tests,fixturesFolder=false
+cypress run --config integrationFolder=tests,videoUploadOnPasses=false
 ```
 
 ```shell
-cypress run --record --config viewportWidth=1280,viewportHeight=720
+cypress run --browser firefox --config viewportWidth=1280,viewportHeight=720
 ```
 
+For more complex configuration objects, you may want to consider passing a {% url "JSON.stringified" https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify %} object surrounded by quotes.
```
bahmutov, line 145: surrounded by single quotes

jennifer-shehane, line 145:
```suggestion
For more complex configuration objects, you may want to consider passing a {% url "JSON.stringified" https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify %} object surrounded by single quotes.
```

rchovatiya88:
@jennifer-shehane Thanks so much, being able to add array to the config file lets us use cucumber feature files and regular files spec files 🥇  
`"testFiles": ["**/*.js","**/*.{feature,features}"],`

Nervniyak:
It doesn't work?

image

PR #2828, opened 2020-05-28
add warning about hooks at root level

bahmutov:
Mentions a potential mine when using hooks, users do ask these questions, like https://github.com/cypress-io/cypress/issues/7313

Screen Shot 2020-05-28 at 5 50 48 PM

Links to the blog post https://glebbahmutov.com/blog/run-all-specs/


source/guides/core-concepts/writing-and-organizing-tests.md
```diff
@@ -225,6 +225,10 @@ describe('Hooks', () => {
 {% fa fa-warning %} Before writing `after()` or `afterEach()` hooks, please see our {% url "thoughts on the anti-pattern of cleaning up state with `after()` or `afterEach()`" best-practices#Using-after-or-afterEach-hooks %}.
 {% endnote %}
 
+{% note danger %}
+{% fa fa-warning %} Be wary of placing hooks in the root scope, as they could execute in surprising order when clicking "Run all specs" button. Instead place them inside `describe` or `context` suites. Read {% url "Be careful when running all specs together" https://glebbahmutov.com/blog/run-all-specs/ %}.
```
jennifer-shehane, line 229: I don't think 'root scope' is very clear here. We don't use this language anywhere else. I wouldn't even recommend this workaround necessarily. They shouldn't be running 'run all specs' in my opinion - we should remove it. Maybe:

```suggestion
{% fa fa-warning %} Be wary that root level hooks defined within a `supportFile` execute only once when clicking the "Run all specs" button while normally they would execute once before each spec file. Read {% url "Be careful when running all specs together" https://glebbahmutov.com/blog/run-all-specs/ %}.
```

Also I think this would probably be better in this section: https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests.html#Support-file

bahmutov: Uggh, the focus of my blog post is the opposite problem: hooks from one spec file are executed for _every_ test from other spec files. Let me re-work and expand the section

jennifer-shehane: Eww...this is exactly why we should remove it.

PR #2854, opened 2020-06-04
Network Stubbing Documentation

panzarino:
Documentation for network stubbing features


jennifer-shehane:
Have some suggestions for the stylings. Feel free to voice any feedback on this idea - good or bad! 

I committed the changes, but they can be easily changed/reverted if needed.

- Trying to keep more inline with the theme of current docs - using side border and existing colors for 'warnings'. Plus this seems a lot more noticeable when scrolling down.
- Only color the main content - I think highlighting across the table of contents just looks off. 
- Update highlighting of certain elements so it's not gray on orange and instead white on orange. Also added bg color to table cause it looked really weird. 

#### Previous Experimental

Screen Shot 2020-07-07 at 10 54 30 AM


#### Current Experimental Proposal

Screen Shot 2020-07-08 at 12 54 48 PM

#### Non-experimental

Screen Shot 2020-07-08 at 12 55 26 PM

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+
+```javascript
+cy.route2('/users/**')
+```
+
+## Arguments
+
+**{% fa fa-angle-right %} requestMatcher** ***(String, Glob, RegExp, Object)***
+
+The `requestMatcher` argument allows you to listen for a route matching a specific URL or pattern.
```
flotwig, line 40: The name of the type is `RouteMatcher`, not `RequestMatcher`:

https://github.com/cypress-io/cypress/blob/50a45c1b707ef0ccb0447b94ecb04946e994a4db/packages/net-stubbing/lib/external-types.ts#L209-L210

In general, I feel like the docs and types should be in line.

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+// Regex
+cy.route2(/users\/\d+/)
+
+// Glob to match all paths in the /user path
+cy.route2('/users/**')
+
+// https://localhost:8080/users/1b2c3            <-- matches
+// https://localhost:8080/users/profile/edit  <-- matches
+// https://localhost:8080/users/transaction?month=03&year=2020 <-- matches
+
```
flotwig, line 57:
```suggestion
```

newline is making github's syntax highlighting weird, i think

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+**{% fa fa-angle-right %} requestMatcher** ***(String, Glob, RegExp, Object)***
+
+The `requestMatcher` argument allows you to listen for a route matching a specific URL or pattern.
+
+For simple route matching, passing a string, glob, or RegExp is the simplest method:
+
+```js
+// String
+cy.route2('/users)
```
flotwig, line 46:
```suggestion
cy.route2('/users')
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+`auth` | `null` | Object with a `username` and `password` to match HTTP basic authentication
+`headers` | `null` | Object to match client request headers
+`hostname` | `null` | String, RegExp, or Glob to match based on requested hostname
+`method` | `'ALL'` | Match based on HTTP method (`GET`, `POST`, `PUT`, etc.)
+`path` | `null` | String, RegExp, or Glob to match on request path after the hostname, including query params
+`pathname` | `null` | String, RegExp, or Glob to match on request path after the hostname, without query params
+`port` | `null` | Match based on requested port number
+`query` | `null` | Object to match based on query parameters
+`url` | `null` | String, RegExp, or Glob to match based on full request URL
+
```
flotwig, line 82: Probably worth mentioning that globbing uses the minimatch API, so that users can understand how globbing works: https://github.com/isaacs/minimatch

https://github.com/cypress-io/cypress/blob/50a45c1b707ef0ccb0447b94ecb04946e994a4db/packages/net-stubbing/lib/server/intercept-request.ts#L63-L63

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+```js
+// String
+cy.route2('/users)
+
+// Regex
+cy.route2(/users\/\d+/)
+
+// Glob to match all paths in the /user path
+cy.route2('/users/**')
```
flotwig, line 52: This is another good place to link to minimatch's documentation

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+    statusCode: 200,
+    body: JSON.stringify({
+      profile: {
+        firstName: 'Tony',
+        lastName: 'Jarvis'
+      }
+    })
+  })
+})
+```
```
flotwig, line 99: This is a strange example, this can be done more directly like this:

```suggestion
```js
cy.route2('/users/**', {
  statusCode: 200,
  body: {
    profile: {
      firstName: 'Tony',
      lastName: 'Jarvis'
  }
})
```

Note that the StaticResponse can be supplied directly, without needing to do the dynamic request/response interception.

Additionally, a `body` that is an `object` will automatically be JSONified, and a `Content-Type: application/json` header will be added to the response automatically. This should probably be mentioned somewhere in this doc. Implementation:

https://github.com/cypress-io/cypress/blob/50a45c1b707ef0ccb0447b94ecb04946e994a4db/packages/driver/src/cy/net-stubbing/static-response-utils.ts#L84-L91

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+
+In addition, because you have access to the request, this means that you can modify the request as well:
+
+```js
+cy.route2('/users/**', (req) => {
+  req.headers = { ...req.headers, accept: 'application/json' }
```
flotwig, line 105:
```suggestion
  req.headers['accept'] = 'application/json'
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+  req.body = { ...req.body, note: 'Custom note' }
```
flotwig, line 106: `req.body` will currently be a string in all situations, iirc. Might want to write a test to make sure, let me know if you want to pair on this

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+})
+```
+
+**{% fa fa-angle-right %} method** ***(String)***
+
+Match the route to a specific HTTP method (`GET`, `POST`, `PUT`, etc).
+
+{% note bolt %}
+By default, Cypress will match `ALL` requests by default.
```
flotwig, line 115:
```suggestion
If a `method` is not explicitly set, Cypress will match on any HTTP method.
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+```javascript
+cy.route2('POST', '**/users').as('postUser')
+cy.visit('/users')
+cy.get('#first-name').type('Julius{enter}')
+cy.wait('@postUser')
+```
+
+### Wait on `url` matching glob
+
+Under the hood Cypress uses {% url 'minimatch' https://github.com/isaacs/minimatch %} to match glob patterns of `url`.
```
flotwig, line 147: `minimatch` support applies to any `S` string property on `RouteMatcher` as well: https://github.com/cypress-io/cypress/blob/50a45c1b707ef0ccb0447b94ecb04946e994a4db/packages/net-stubbing/lib/external-types.ts#L125-L167 this includes strings in the `auth`, `headers`, and `query` objects

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+
+You can also (optionally) add a simulated {% urlHash 'delay' Simulate-delay %} or {% urlHash 'throttle' Simulate-throttle %} on the request to be sent out.
+
+```javascript
+cy.route2('/users', (req) => {
+  // modify the headers or body
+  req.headers = {...req.headers, accept: 'application/json'}
+  req.body = {...req.body, foo: 'bar'}
+})
+```
```
flotwig, line 184: see suggestions from where this example was used earlier

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Modify response
+
+After a request that gets passed through, we can modify the response from the server before it is returned through a function. The function exposes a `req` object which contains the request. The `req.send()` function exposes a `res` object which contains the response.
```
flotwig, line 188: Terminology used here is strange. "The function exposes" is not phrasing I've heard before. Something along the lines of "The callback supplied as the `RouteHandler` receives a `req` object that contains request information, as well as a `req.send()` function that accepts a callback that will receive a `res` object containing the real response data." would be closer

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+```javascript
+cy.route2('/users', (req) => {
+  // passing a function to req.send causes the request to pass through
+  // and allows the response from the origin server to be modified
+  req.send((res) => {
+    res.status = 200
+    res.headers['x-new-headers'] = 'from-server'
+
+    // dynamically alter body
+    res.body = res.body.replace('window.top', 'window.self')
```
flotwig, line 199:
```suggestion
    res.body = res.body.replace('window.top', 'window.self')
    // now, the response will continue to the browser
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+  })
+})
+```
+
+## With Stubbing
+
+If you pass a `response` to `cy.route()`, Cypress will stub the response in the request.
```
flotwig, line 206:
```suggestion
If you pass a `response` to `cy.route2()`, Cypress will stub the response in the request.
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+
+### `url` as a string
+
+When passing a `string` as the `url`, the URL must match *exactly* what you've written. You'll want to use the decoded string and not include any hash encoding (ie. use `@` instead of `%40`).
```
flotwig, line 210: not true, it is matched via `minimatch`. also check with tests that this is still true for cy.route2, i believe this was copied from cy.route

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+})
+```
+
+### Response functions
+
+You can also use a function as a response which enables you to add logic and modify properties of the response. The function exposes a `req` object which contains the request as well as a way to `send` with a response.
+
+```javascript
+cy.route2('/login', (req) => {
+  req.send('Success')
```
flotwig, line 242: it's `req.reply` vs `res.send`:

https://github.com/cypress-io/cypress/blob/50a45c1b707ef0ccb0447b94ecb04946e994a4db/packages/net-stubbing/lib/external-types.ts#L44-L50

https://github.com/cypress-io/cypress/blob/50a45c1b707ef0ccb0447b94ecb04946e994a4db/packages/net-stubbing/lib/external-types.ts#L19-L38

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+})
+```
+
+```javascript
+// reply with body and headers
+cy.route2('/login', (req) => {
+  req.send('Success', { 'x-new-headers': 'from-server' })
```
flotwig, line 249: req.reply

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+})
+```
+
+```javascript
+// reply with an object containing more response details
+cy.route2('/login', (req) => {
+  req.send({
```
flotwig, line 256: req.reply

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+cy.route2('/login', (req) => {
+  req.send('Success')
+})
+```
+
+```javascript
+// reply with body and headers
+cy.route2('/login', (req) => {
+  req.send('Success', { 'x-new-headers': 'from-server' })
+})
```
flotwig, line 250: strange example, this would be more straightforward:

```suggestion
// reply with body and headers
cy.route2('/login', {
  body: 'Success', 
  headers: { 'x-new-headers': 'from-server' }
})
```

this is important b/c when the response can be supplied statically, it is no longer necessary to make serializations and round-trips to the browser, so it significantly improves performance if the user short-circuits the response by supplyin a `StaticResponse` as early as possible. the examples should reflect this - we should refrain from using dynamic request/response interception unless we actually need the dynamic features for an example

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+```
+
+### Response functions
+
+You can also use a function as a response which enables you to add logic and modify properties of the response. The function exposes a `req` object which contains the request as well as a way to `send` with a response.
+
+```javascript
+cy.route2('/login', (req) => {
+  req.send('Success')
+})
```
flotwig, line 243:
```suggestion
cy.route2('/login', 'Success')
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+    headers: {
+      'x-new-headers': 'from-server'
+    },
+    // If `destroySocket` is truthy, Cypress will destroy the connection to the
+    // browser and send no response. Useful for simulating a server that is not
+    // reachable. Must not be set in combination with other options.
+    destroySocket: false
+  })
+})
+```
```
flotwig, line 270: again, this can be done statically (and should be):

```suggestion
```javascript
// reply with an object containing more response details
cy.route2('/login', {
  statusCode: 200,
  body: {
    some: 'response'
  },
  headers: {
    'x-new-headers': 'from-server'
  },
  // If `destroySocket` is truthy, Cypress will destroy the connection to the
  // browser and send no response. Useful for simulating a server that is not
  // reachable. Must not be set in combination with other options.
  destroySocket: false
})
```
```

as before, the body will be JSONified automatically

File: source/api/commands/route2.md [hunk 22]
```diff
@@ -0,0 +1,385 @@
...
+
+```javascript
+// continue the HTTP response to the browser, including any modifications made to `res`
+cy.route2('/login', (req) => {
+  req.send()
+})
+```
```
flotwig, line 277: this is functionally identical to the following code, but has worse performance because of sending req/res to the browser unnecessarily

```suggestion
```javascript
// continue the HTTP response to the browser, including any modifications made to `res` <--- what's `res` here?
cy.route2('/login')
```
```

File: source/api/commands/route2.md [hunk 23]
```diff
@@ -0,0 +1,385 @@
...
+
+### Matching requests and routes
+
+Any request that matches the `method` and `url` of a route will be responded to based on the configuration of that route.
```
flotwig, line 281: or `auth`, or `path`, or `query`, or `port`, etc... - should probably mention `RouteMatcher` here

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+cy.get('#search').type('Geotrupidae')
+
+// now when we wait for 'getBeetles' again, Cypress will
+// automatically know to wait for the 2nd response
+cy.wait('@getBeetles')
+
+// we responded with 1 beetle item so now we should
+// have one result
+cy.get('#beetle-results').should('have.length', 1)
+```
```
flotwig, line 322: might need to verify this example, not sure stubbing with an array like this will work, could be a potential bug

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Fixtures
+
+Instead of writing a response inline you can automatically connect a response with a {% url `cy.fixture()` fixture %}.
+
+```javascript
+cy.route2('**/posts/*', 'fixture:logo.png').as('getLogo')
+cy.route2('**/users', 'fixture:users/all.json').as('getUsers')
+cy.route2('**/admin', 'fx:users/admin.json').as('getAdmin')
```
flotwig, line 331: this is no longer the way to do it in `route2`, should be a `StaticResponse` with a `fixture` key set: https://github.com/cypress-io/cypress/blob/50a45c1b707ef0ccb0447b94ecb04946e994a4db/packages/net-stubbing/lib/external-types.ts#L179-L182

```suggestion
cy.route2('**/posts/*', { fixture: 'logo.png' }).as('getLogo')
cy.route2('**/users', { fixture: 'users/all.json' }).as('getUsers')
cy.route2('**/admin', { fixture: 'users/admin.json' }).as('getAdmin')
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,385 @@
...
+cy.get('.user').should('include', 'Jane')
+```
+
+You can also reference fixtures as strings directly in the response by passing an aliased fixture with `@`.
+
+```javascript
+cy.fixture('user').as('fxUser')
+cy.route2('POST', '**/users', '@fxUser')
+```
+
```
flotwig, line 354: this has been removed from `cy.route2` API
```suggestion
```

themes/cypress/layout/media.swig
```diff
@@ -3,143 +3,141 @@
     <div id="content-inner">
```
flotwig, line 3: are the changes in this file related to the PR or can they be reverted?

bencodezen: This is necessary for the experimental styling that you're seeing on the page.

source/api/commands/route2.md
bencodezen: Adding it as a comment!

source/api/commands/route2.md
```diff
@@ -0,0 +1,377 @@
...
+cy.route2('**/posts/**')
+
+// https://localhost:7777/posts/1            <-- matches
+// https://localhost:7777/posts/foo/bar/baz  <-- matches
+// https://localhost:7777/posts/quuz?a=b&1=2 <-- matches
+```
+
+### Modify request
+
+Before a request gets passed through, we can modify the request before it is sent out through a function. The function exposes a `req` object which contains the request.
```
jennifer-shehane, line 173: This could be named anything, weird to define it as `req` object specifically. 

```suggestion
Before a request gets passed through, we can modify the request. The `response` function accepts the request object as a parameter.
```

source/api/commands/route2.md
jennifer-shehane: Yah, I said the same above. It doesn't matter what this parameter is called...it's a param, they can name it anything.

```suggestion
After a request gets passed through, we can modify the response from the server. The `response` function accepts the request object as a parameter. Calling `.send()` on the request exposes a `response` object that can be modified.
```

source/api/commands/route2.md
bencodezen: I've updated the example to show multiple routes, but could use some feedback on whether a more practical example exists!

bencodezen:
Thanks for your feedback @jennifer-shehane! Great improvements and definitely need to figure out how we can update the menuspy / dependency on h1s in the docs given it breaks content accessibility. Another problem for another time though. 😄

source/api/commands/route2.md
```diff
@@ -0,0 +1,381 @@
...
+```javascript
+cy.route2('/login', (req) => {
+  req.send('Success')
+})
+```
+
+```javascript
+// reply with body and headers
+cy.route2('/login', (req) => {
+  req.send('Success', { 'x-new-headers': 'from-server' })
```
bencodezen, line 249:
```suggestion
  req.reply('Success', { 'x-new-headers': 'from-server' })
```

source/api/commands/route2.md
bencodezen: @flotwig Agreed. My goal is at some point these should be automatically synced when the core codebase changed so it's autogenerated rather than manual syncing.

source/api/commands/route2.md
```diff
@@ -0,0 +1,376 @@
...
+cy.route2('/users/**')
+```
+
+## Arguments
+
+**{% fa fa-angle-right %} routeMatcher** ***(String, Glob, RegExp, Object)***
+
+The `routeMatcher` argument allows you to listen for a route matching a specific URL or pattern.
+
+For simple route matching, passing a string, glob, or RegExp is the simplest method:
```
flotwig, line 42: simple simple simple simple

```suggestion
To match on requests with a certain URL, passing a string, glob, or RegExp is the simplest method:
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,376 @@
...
+
+```js
+// String
+cy.route2('/users')
+
+// Regex
+cy.route2(/users\/\d+/)
+
+// Glob to match all paths in the /user path
+// Uses minimatch for matching: https://github.com/isaacs/minimatch
```
flotwig, line 52: to be specific, matches against URLs

```suggestion
// Uses minimatch for matching against request URLs: https://github.com/isaacs/minimatch
```

source/api/commands/route2.md
flotwig: still need a mention here - passing a string to `auth.username`, `auth.password`, `headers.*`, `hostname`, `method`, `path`, `pathname`, `url` (any string property matcher) will use minimatch to match under the hood

flotwig: here is the backend's implementation for matching `StringMatcher` fields: https://github.com/cypress-io/cypress/blob/0d60f7cd3ede4c5a79e151b646fa2377a7ddb16c/packages/net-stubbing/lib/server/intercept-request.ts#L41-L64

source/api/commands/route2.md
flotwig: still needs to be resolved

source/api/commands/route2.md
flotwig: as mentioned earlier, this section could also be updated to note that minimatch is used for any `StringMatcher` field

source/api/commands/route2.md
flotwig: see resolution to https://github.com/cypress-io/cypress-documentation/pull/2854/files/45ae8f49408128281ba5864dfa59354521b1d587#diff-6066033e28279d50ba94ae43b9b5b5a6

source/api/commands/route2.md
```diff
@@ -0,0 +1,376 @@
...
+// https://localhost:7777/posts/1            <-- matches
+// https://localhost:7777/posts/foo/bar/baz  <-- matches
+// https://localhost:7777/posts/quuz?a=b&1=2 <-- matches
+```
+
+### Modify request
+
+Before a request gets passed through, we can modify the request. The `response` function accepts the request object as a parameter.
+
+You can also (optionally) add a simulated {% urlHash 'delay' Simulate-delay %} or {% urlHash 'throttle' Simulate-throttle %} on the request to be sent out.
```
flotwig, line 174: seems like a random place to mention this without an example - the user could also forceNetworkError or reply with a fixture, but neither of those are mentioned here, so why this here?

also, it would be good to document that the throttle/delay is applied to the incoming response. currently the outgoing request cannot be throttled, only the incoming response.

source/api/commands/route2.md
```diff
@@ -0,0 +1,376 @@
...
+})
+```
+
+## With Stubbing
+
+If you pass a `response` to `cy.route2()`, Cypress will stub the response in the request.
+
+### `url` as a string
+
+When passing a String as the `url`, the URL must match *exactly* what you've written. You'll want to use the decoded string and not include any hash encoding (ie. use `@` instead of `%40`).
```
flotwig, line 209: this is not true - the URL specifically is checked for a substring match, otherwise it is globbed and matched with minimatch, see implementation: https://github.com/cypress-io/cypress/blob/0d60f7cd3ede4c5a79e151b646fa2377a7ddb16c/packages/net-stubbing/lib/server/intercept-request.ts#L41-L64

source/api/commands/route2.md
```diff
@@ -0,0 +1,376 @@
...
+
+```javascript
+cy.route2('https://localhost:7777/users/customer?email=john@doe.com', [
```
flotwig, line 212: this example should be double-checked if it hasn't, urlencoding may affect the `@` character

source/api/commands/route2.md
```diff
@@ -0,0 +1,376 @@
...
+  body: {
+    profile: {
+      firstName: 'Tony',
+      lastName: 'Jarvis'
+    }
+  }
+})
+```
+
+In addition, because you have access to the request, this means that you can modify the request as well:
```
flotwig, line 99:
```suggestion
In addition, you can pass a callback that can modify the outgoing request:
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,376 @@
...
+  req.body = { ...req.body, note: 'Custom note' }
+})
+```
+
+**{% fa fa-angle-right %} method** ***(String)***
+
+Match the route to a specific HTTP method (`GET`, `POST`, `PUT`, etc).
+
+{% note bolt %}
+By default, Cypress will match `ALL` requests.
```
flotwig, line 113:
```suggestion
If a `method` is not explicitly set, Cypress will match on any HTTP method.
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,376 @@
...
+cy.route2('**/posts/**')
+
+// https://localhost:7777/posts/1            <-- matches
+// https://localhost:7777/posts/foo/bar/baz  <-- matches
+// https://localhost:7777/posts/quuz?a=b&1=2 <-- matches
+```
+
+### Modify request
+
+Before a request gets passed through, we can modify the request. The `response` function accepts the request object as a parameter.
```
flotwig, line 172: it's confusing to call this the `response` function. that's the parameter's name, but the type is a `RouteHandler` in the code, which narrows to a `HttpRequestInterceptor` if it's a function. so 

```suggestion
Before a request gets sent to the real endpoint, we can modify the request. The supplied `RouteHandler` callback receives the request object as a parameter.
```

another place where it would be good to have links to types

source/api/commands/route2.md
flotwig: ^ this would be a good thing to centralize, since this applies the same anywhere a `StaticResponse` is used

source/api/commands/route2.md
```diff
@@ -0,0 +1,376 @@
...
+cy.route2('/users', (req) => {
+  // modify the headers or body
+  req.headers = {...req.headers, accept: 'application/json'}
+  req.body = {...req.body, foo: 'bar'}
+})
+```
+
+### Modify response
+
+After a request gets passed through, we can modify the response from the server. The `response` function accepts the request object as a parameter. Calling `.send()` on the request exposes a `response` object that can be modified.
```
flotwig, line 186: What is the "`response` function"? It is not clear to me what this is referring to...

source/api/commands/route2.md
flotwig: previous review comment: https://github.com/cypress-io/cypress-documentation/pull/2854#discussion_r461826856

bencodezen:
@flotwig Sent you a longer update in Slack in regards to next steps. Let me know if you have any questions!

source/api/commands/route2.md
```diff
@@ -0,0 +1,394 @@
...
+})
+```
+
+### Modify response
+
+After a request gets passed through, we can modify the response from the server. This is done by passing the `req.send()` method a callback function which receives the original response (i.e., `res`) as the first argument.
+
+```js
+cy.route2('/users', (req) => {
+  // passing a function to req.send causes the request to pass through
```
flotwig, line 208:
```suggestion
  // passing a function to req.reply causes the request to pass through
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,394 @@
...
+    // dynamically alter body
+    res.body = res.body.replace('window.top', 'window.self')
+    // now, the response will continue to the browser
+  })
+})
+```
+
+## With Stubbing
+
+If you pass a `response` to `cy.route2()`, Cypress will stub the response in the request.
```
flotwig, line 223:
```suggestion
If you pass a `routeHandler` as the second argument to `cy.route2()`, Cypress will allow you to stub the response to the request.
```

source/api/commands/route2.md
flotwig: ^ still not true

source/api/commands/route2.md
```diff
@@ -0,0 +1,394 @@
...
+    some: 'response'
+  },
+  headers: {
+    'x-new-headers': 'from-server'
+  },
+  // If `destroySocket` is truthy, Cypress will destroy the connection
+  // to the browser and send no response. Useful for simulating a
+  // server that is not reachable. Must not be set in combination
+  // with other options.
+  destroySocket: false
```
flotwig, line 282:
```suggestion
  // If `forceNetworkError` is truthy, Cypress will destroy the connection
  // to the browser and send no response. Useful for simulating a
  // server that is not reachable. Must not be set in combination
  // with other options.
  forceNetworkError: false
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,394 @@
...
+cy.route2(url, response)
+cy.route2(method, url)
+cy.route2(method, url, response)
+```
+
+```typescript
+// TypeScript
+cy.route2(url: RouteMatcher, response?: RouteHandler): Chainable<null>
+cy.route2(method: string, url: RouteMatcher, response?: RouteHandler)): Chainable<null>
+```
```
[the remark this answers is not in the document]
bencodezen: @flotwig Good question. So one of the things I wanted to avoid was introducing TypeScript to users who aren't familiar with the syntax. In the new docs, it'd be have more flexibility in terms of letting users choose which one they want to see rather than rendering both.

As far as why there are so many variations, it is to keep consistent with past API signature definitions to show the various ways it could be used.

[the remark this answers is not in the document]
flotwig: > So one of the things I wanted to avoid was introducing TypeScript to users who aren't familiar with the syntax.

By that logic, let's not have 2 sections for signatures, let's just have 1. That way users don't have to be confused about why there would magically be a different API if they're using Typescript.

> As far as why there are so many variations, it is to keep consistent with past API signature definitions to show the various ways it could be used.

Well, it leaves the `routeMatcher` and `routeHandler` options undocumented in the signatures, so I would say it is failing to show the ways it could be used.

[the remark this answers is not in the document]
bencodezen: @flotwig That's fair. I think given the experimental nature of the feature and what we discussed in Slack. Let's move forward on being a little more progressive with API definitions.

source/api/commands/route2.md
```diff
@@ -0,0 +1,394 @@
...
+
+**{% fa fa-check-circle green %} Correct Usage**
+
+```javascript
+cy.route2('/users/**')
+```
+
+## Arguments
+
+**{% fa fa-angle-right %} url**
```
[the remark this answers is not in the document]
bencodezen: Going to leave it as `url` since this is what most users are familiar with, but there is an explicit definition of `Type` inside of it so the `routeMatcher` piece is not lost.

source/api/commands/route2.md
```diff
@@ -0,0 +1,394 @@
...
+`port` | `null` | *Number* | Requested port number
+`query` | `null` | *Object* | Query parameters
+`url` | `null` | *String, RegExp, Glob* | Full request URL
+
+**{% fa fa-angle-right %} response**
+
+- **Description**: Allows you to supply a response `body` to {% url 'stub' stubs-spies-and-clocks#Stubs %} in the matching route.
+
+- **Expects**: `String` | `Object` | `Array` | `Function`
+
```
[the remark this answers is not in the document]
bencodezen: I've added a temporary placeholder inside of the docs prior to a docs rewrite. Hope this helps move the needle on this!

source/api/commands/route2.md
```diff
@@ -0,0 +1,449 @@
...
+// that will be sent back to the browser
+// to fulfill the request
+cy.route2('/users', {
+  body: '{ "message": "Custom message" }',
+  headers: {
+    accept: 'application/json'
+  },
+  statusCode: 200,
+  forceNetworkError: false
+})
```
flotwig, line 82: You're defining a response, so the `accept` header isn't used. Also, you can set `body` to an object to automatically set the content-type: json header.

```suggestion
cy.route2('/users', {
  body: { message: "Custom message" },
  statusCode: 200,
  forceNetworkError: false
})
```

source/api/commands/route2.md
```diff
@@ -0,0 +1,449 @@
...
+```
+
+### Simulate delay
+
+You can specify a delay (in ms) before the response is sent to the client.
+
+```javascript
+cy.route2('/users', (req) => {
+  req.delay(200)
```
flotwig, line 437: there is no req.delay, only res.delay

source/api/commands/route2.md
```diff
@@ -0,0 +1,449 @@
...
+})
+```
+
+### Simulate throttle
+
+You can specify the speed at which a response is served at (in kilobytes per second).
+
+```javascript
+cy.route2('/users', (req) => {
+  req.throttle(3000)
```
flotwig, line 447: there is no req.throttle, only res.throttle

source/api/commands/route2.md
```diff
@@ -0,0 +1,449 @@
...
+cy.route2('/users', (req) => {
+  // passing a function to req.reply causes the request to pass through
+  // and allows the response from the origin server to be modified
+  req.reply((res) => {
+    res.status = 200
+    res.headers['x-new-headers'] = 'from-server'
+
+    // dynamically alter body
+    res.body = res.body.replace('window.top', 'window.self')
+    // now, the response will continue to the browser
```
flotwig, line 270: so this is the only place where req.reply is mentioned...

but the res object doesn't seem to be documented. none of these functions are documented anywhere, even though they provide critical functionality to modifying origin responses:

https://github.com/cypress-io/cypress/blob/cc357f98992ace16d39bbfcdd155421c9d474b15/packages/net-stubbing/lib/external-types.ts#L19-L39

source/api/commands/route2.md
```diff
@@ -0,0 +1,449 @@
...
+
+### Modify response
+
+After a request gets passed through, we can modify the response from the server. This is done by passing the `req.reply()` method a callback function which receives the original response (i.e., `res`) as the first argument.
+
+```js
+cy.route2('/users', (req) => {
+  // passing a function to req.reply causes the request to pass through
+  // and allows the response from the origin server to be modified
+  req.reply((res) => {
```
flotwig, line 264: once `res.send()` is documented, it should be noted that `res.send()` will implicitly be called once the function finishes (or if the function returns a Promise, it will be called once the Promise resolves)

source/api/commands/route2.md
flotwig: ex. res.send, res.delay, res.throttle...

source/api/commands/route2.md
```diff
@@ -0,0 +1,449 @@
...
+  req.body = JSON.stringify({
+    ...requestBody, 
+    note: 'Custom note' 
+  })
+})
+```
+
+### Modify response
+
+After a request gets passed through, we can modify the response from the server. This is done by passing the `req.reply()` method a callback function which receives the original response (i.e., `res`) as the first argument.
```
flotwig, line 258: the other invocation of `req.reply` (passing no argument, or passing a StaticResponse) is not documented anywhere as far as i can tell

flotwig: once req.reply() is documented, it should be noted that req.reply() will implicitly be called once the `routeHandler` finishes (or if the `routeHandler` returns a Promise, it will be called once the Promise resolves)

bencodezen:
After discussing with @flotwig, the plan is to release the docs in its current form and iterate on this in a future PR.

source/api/commands/route2.md
```diff
@@ -0,0 +1,446 @@
...
+  path: '/users/**',
+  method: 'POST',
+})
+
+// StaticResponse
+// Allows you to define a response
+// that will be sent back to the browser
+// to fulfill the request
+cy.route2('/users', {
+  body: { message: "Custom message" },
```
flotwig, line 76:
```suggestion
  body: { message: 'Custom message' },
```

mdlint

source/api/commands/route2.md
```diff
@@ -0,0 +1,446 @@
...
+  statusCode: 200,
+  forceNetworkError: false
```
flotwig, line 78:
```suggestion
  statusCode: 200
```

PR #2858, opened 2020-06-05
Add shadow dom experimental flag

chrisbreiding:
- Add `experimentalShadowDomSupport` description to experiments page.
- [x] Dependent on https://github.com/cypress-io/cypress/pull/7469 being merged in


JessicaSachs:
Going to wait on merging/approving until we update the flag name

source/guides/references/experiments.md
```diff
@@ -17,7 +17,7 @@ Option | Default | Description
...
 `experimentalComponentTesting` | `false` | When set to `true`, Cypress allows you to execute component tests using framework-specific adaptors. By default  `cypress/component` is the path for component tests. You can change this setting by setting the `componentFolder` configuration option. For more details see the {% url "cypress-react-unit-test" https://github.com/bahmutov/cypress-react-unit-test %} and {% url "cypress-vue-unit-test" https://github.com/bahmutov/cypress-vue-unit-test %} repos.
 `experimentalSourceRewriting` | `false` | Enables AST-based JS/HTML rewriting. This may fix issues caused by the existing regex-based JS/HTML replacement algorithm. See {% issue 5273 %} for details.
-`experimentalShadowDomSupport` | `false` | Enables shadow DOM support. Adds the `cy.shadow()` command and the `ignoreShadowBoundaries` option to some DOM commands.
+`experimentalShadowDomSupport` | `false` | Enables shadow DOM support. Adds the `cy.shadow()` command and the `includeShadowDom` option to some DOM commands.
```
brian-mann, line 20: Can we make `experimentalShadowDomSupport` and `experimentalComponentTesting` clickable hash based anchor links that scroll you down to their associated detailed docs?

I think you use `urlHash` helper to achieve this.

brian-mann: I would also consider reducing down this table to not include so many implementation details - and instead just write a brief <10 words description, and then have a dedicated section for each one of these so the template is consistent.

As it stands, this table is virtually unscannable, and it's easy to miss what the actual available experiments are.

Even if the sections for each experiment are brief, at the very least we could list out the issues, add screenshots, or explain **why** you would want to enable this experimental option.

JessicaSachs: Do we want to block the release on this or can we do this afterwards?

JessicaSachs: I agree with everything you're saying :-)

brian-mann: experimental docs should never be blocking

chrisbreiding: In that case, I'm going to merge this and open a new PR with the changes suggested.

PR #2876, opened 2020-06-11
Add instruction to check if element never existed

eric-burel:
See discussion at https://github.com/cypress-io/cypress/issues/7651#issuecomment-642442799

"Existence" might be ambiguous: you may want an element to be removed, but you may also check that an element never existed. For example check that a loader does not appear immediately, or that a server-side render was successful.

### DIFF
--- a/source/guides/references/assertions.md
+++ b/source/guides/references/assertions.md
@@ -229,19 +229,11 @@ cy.get('a').parent('span.help').should('not.contain', 'click me')
 cy.get('button').should('be.visible')
 ```
 
-## Removal
+## Existence
 
 ```javascript
 // retry until loading spinner no longer exists
 cy.get('#loading').should('not.exist')
-```
-
-## Existence
-
-```javascript
-// check synchronously that the element does not exist (no retry)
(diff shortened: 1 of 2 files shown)


source/guides/references/assertions.md
```diff
@@ -230,13 +229,21 @@ cy.get('a').parent('span.help').should('not.contain', 'click me')
...
 // retry until loading spinner no longer exists
 cy.get('#loading').should('not.exist')
 ```
 
+## Existence
+
+```javascript
+// check synchronously that the element does not exist (no retry)
+// for example just after a server-side render
+cy.get('#ssr-error', { timeout: 0}).should('not.exist')
```
jennifer-shehane, line 244: Linting error about spacing.

```suggestion
cy.get('#ssr-error', { timeout: 0 }).should('not.exist')
```

jennifer-shehane:
I'm thinking perhaps this would be better in the retry-ability doc, if at all.

jennifer-shehane (dismissed):

PR #2856, opened 2020-06-05
4.8.0 Release

jennifer-shehane:
- 4.8.0 Changelog
- Section for new spec file header #2852
- Update screenshots for error design changes #2853 
- [x] Test config options + isBrowser() changes #2697 
- [x] Update wording around 'test options' / 'test config'
- [x] `experimentalShadowDom` experimental flag #2858 

---

**Preview of main content for 'test config' feature**

Screen Shot 2020-06-08 at 2 16 00 PM

Screen Shot 2020-06-08 at 2 16 09 PM

PR #2896, opened 2020-06-18
docs: fix incorrect syntax in example

bencodezen:
thanks to @JessicaSachs for noticing it!


jennifer-shehane:
I think this may be wrong in a lot more places. Cause I mistakenly put the arg before the config before and then just removed it I think. https://github.com/cypress-io/cypress-documentation/pull/2856/files

PR #2917, opened 2020-06-23
add missing feature to 4.8.0 changelog

bahmutov:


source/_changelogs/4.8.0.md
```diff
@@ -34,6 +34,7 @@
 - TypeScript types for `Cypress.automation` have been added. Addresses {% issue 7519 %}.
 - We now display a more accurate error message when passing a browser to the `--browser` flag that is not supported by Cypress. Addresses {% issue 6979 %}.
 - We're continuing to make progress in converting our codebase from CoffeeScript to JavaScript. Addresses {% issue 2690 %} in {% PR 7588 %} and {% PR 7591 %}.
+- We added support for scaffolding a project with examples to `cypress run` mode, see {% issue 7554 %} and {% PR 7558 %}.
```
jennifer-shehane, line 37: The user facing changelog on the PR says there was no user facing changelog. If this isn't documented, then it shouldn't be in the changelog. I didn't think we documented `CYPRESS_INTERNAL` env vars.

bahmutov: I would agree with you, but now I think this was an omission. As I was setting up another test job on CI I thought "hmm, was that released? which version was that? How do I find the syntax to use?"  So probably adding this into the changelog makes sense

jennifer-shehane: Not if it's not user facing. If you need to find something internal, then search you'll need to search the GitHub repo.

PR #2932, opened 2020-06-26
Don't Stub responses

gamb:
This seemed to me to be a mixed-message, the title **Don't Stub responses** is confusing as suggests you shouldn't stub, but then you're actually encouraging stubbing within the same page.

Also doesn't seem accurate to claim this is "Guaranteed to work in production".


jennifer-shehane:

gamb:
Thanks @jennifer-shehane , signed 👍

PR #2945, opened 2020-06-30
Update typescript doc to specify minimum required TS version

jennifer-shehane:
For https://github.com/cypress-io/cypress/pull/7806


sainthkh:
How about adding `Minimum TypeScript version 2.9 => 3.4` to history?

jennifer-shehane:
@sainthkh Yeah, honestly I didn't know what the minimum was before. Will specify this more clearly in the Migration guide also.

bahmutov:
Yeah we were never explicit about min TS version before and should be now from now on. Also even have a min TS test job too

PR #2959, opened 2020-07-07
Docs: add `bail` cli arg/module option

loren138:
Goes with: https://github.com/cypress-io/cypress/pull/7892


amirrustam:
Closing for now since this is an old PR. We can reopen once there is some consensus from cypress-io/cypress#7892.

PR #2991, opened 2020-07-16
document movetoend and movetostart special chars for type

jennifer-shehane:
Awaiting some clarification on how this is different from end and home special chars https://github.com/cypress-io/cypress/pull/7970

### DIFF
--- a/source/api/commands/type.md
+++ b/source/api/commands/type.md
@@ -50,6 +50,8 @@ Sequence | Notes
 `{home}` | Moves cursor to the start of the line
 `{insert}` | Inserts character to the right of the cursor
 `{leftarrow}` | Moves cursor left
+`{movetoend}` | Moves cursor to end of typeable element
+`{movetostart}` | Moves cursor to the start of typeable element
 `{pagedown}` | Scrolls down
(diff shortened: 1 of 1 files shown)


source/api/commands/type.md
```diff
@@ -53,9 +53,16 @@ Sequence | Notes
 `{pagedown}` | Scrolls down
 `{pageup}` | Scrolls up
 `{rightarrow}` | Moves cursor right
-`{selectall}` | Selects all text by creating a `selection range`
 `{uparrow}` | Moves cursor up
 
+Text passed to `.type()` may include special actions to be performed during typing:
```
Bkucera, line 58: these are still `special character sequences` and they will be affected by `parseSpecialCharSequences` so it seems strange to call them `special actions` here. Maybe clarify that these are special char seqs that perform actions instead of single keys?

Bkucera (changes requested):

source/api/commands/type.md
jennifer-shehane: yeah, it was a bit different so that's why I split it, but yah, I can see how this is important to clarify since the option still applies. Will update.

PR #2998, opened 2020-07-17
docs: vue quick start guide

bencodezen:


source/guides/quick-start/vue-quick-start.md
```diff
@@ -0,0 +1,429 @@
...
+
+{% video local /img/guides/vue-quick-start/vue-quick-start-debugging-02.mp4 %}
+
+With these automatic DOM snapshots, you'll have the {% url "ability to time-travel" test-runner#Hovering-on-Commands %} through the states of your UI for quick debugging or concisely understanding the behavior of your app.
+
+## Next Steps
+
+In this guide, we have covered the essentials you need for getting started, but there is much more in the world of Cypress.
+
+Thanks for reading our guide and hope it has helped you create more trust in your app.
```
jennifer-shehane, line 429: I can't explain why, but this sentence comes off a bit...odd. I think because it's more focused on our concerns of them trusting our app as opposed their concerns of what they get with Cypress. Maybe below is better?

```suggestion
Thanks for reading our guide and we hope if helps you successfully start testing your Vue applications.
```

bencodezen:
> Overall I think this is very clear, touching on everything needed to 'get started'. I have some nitpick things as usual.
> 
> No mention of Dashboard 😢 But it's so hard to introduce it in an intro series.

Agreed. I'm thinking that I'll create a follow up guide as a way of better introducing people to why it's important and how it solves a common problem developers have.

source/guides/quick-start/vue-quick-start.md
bencodezen: Agreed. This does read odd on another pass. How about we keep it short and simple?

`Thanks for reading our guide!`

amirrustam:
We will revive this content as part of the pending docs transition.

Issue #2102, opened 2019-09-24
Add doc for .rightclick() command


jennifer-shehane:
## I'm submitting a...

```
[ ] Bug report
[x] Content update
[ ] Process update (build, deployment, ... )
```

## Type of bug / changes

`.rightclick()` is being introduced as part of https://github.com/cypress-io/cypress/pull/3030 We need a doc for this and to add it to other places.

PR #3044, opened 2020-07-31
docs: add workaround for large windows in headless (#2102)

ryaninvents:
Add notes from [#2102](https://github.com/cypress-io/cypress/issues/2102) about working with large window sizes in a headless environment

### DIFF
--- a/source/guides/guides/continuous-integration.md
+++ b/source/guides/guides/continuous-integration.md
@@ -723,6 +723,21 @@ module.exports = (on, config) => {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


source/guides/guides/continuous-integration.md
```diff
@@ -723,6 +723,21 @@ module.exports = (on, config) => {
 }
 ```
 
+In addition, {% issue 2102 "this issue" %} reports that headless runs do not render a full-size image for viewports larger than 1280x720. To test larger viewports, set `viewportWidth` and `viewportHeight` in `cypress.json`, and amend the shared memory fix to set the maximum window size:
```
jennifer-shehane, line 726: The `viewportWidth` and `viewportHeight` have nothing to do with the window size. This is the size that the application will render at *within* the Cypress Test Runner. We create an iframe and show the preview at these dimensions.

amirrustam:
Closing for now based on https://github.com/cypress-io/cypress/issues/2102#issuecomment-755882519. We can reopen as needed.

PR #3078, opened 2020-08-18
Update trigger doc with new eventConstructor option

sainthkh:
Related to cypress-io/cypress#8305

# Base branch is develop. Change it appropriately before merge.


jennifer-shehane:
Converting to draft until the original PR is merged in.

source/api/commands/trigger.md
```diff
@@ -169,6 +170,16 @@ This overrides the default auto-positioning based on the element itself. Useful
 cy.get('button').trigger('mousemove', { clientX: 200, clientY: 300 })
 ```
 
+## Fire other Event types.
+
+By default, `cy.trigger()` fires {% url `Event` https://developer.mozilla.org/en-US/docs/Web/API/Event %}. But for any reason, you might want to trigger other events like `MouseEvent`, `KeyboardEvent`. 
+
+In that case, use `eventType` option.
```
jennifer-shehane, line 177:
```suggestion
By default, `cy.trigger()` fires {% url `Event` https://developer.mozilla.org/en-US/docs/Web/API/Event %}. But you may want to trigger other events like `MouseEvent` or `KeyboardEvent`. 

In that case, use the `eventType` option.
```

source/api/commands/trigger.md
```diff
@@ -183,6 +194,10 @@ cy.get('button').trigger('mousemove', { clientX: 200, clientY: 300 })
 
 `cy.trigger()` is meant to be a low-level utility that makes triggering events easier than manually constructing and dispatching them. Since any arbitrary event can be triggered, Cypress tries not to make any assumptions about how it should be triggered. This means you'll need to know the implementation details (which may be in a 3rd party library) of the event handler(s) receiving the event and provide the necessary properties.
 
+### Why should I manually set the event type?
+
+As you can see the documentation of {% url `MouseEvent` https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent %}, most properties of event class instances are read-only. Because of that, it's sometimes impossible to set the value of some properties like `pageX`, `pageY`. It can be problematic in some tests.
```
jennifer-shehane, line 199:
```suggestion
As you can see the documentation of {% url `MouseEvent` https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent %}, most properties of event class instances are read-only. Because of that, it's sometimes impossible to set the value of some properties like `pageX`, `pageY`. This can be problematic in when testing some situations.
```

PR #3088, opened 2020-08-20
Update changelog.md

chghulamnabi:
By using this command your Cypress version will be updated instead of installing your Cypress from start.

### DIFF
--- a/source/guides/references/changelog.md
+++ b/source/guides/references/changelog.md
@@ -1,5 +1,6 @@
 ---
 title: Changelog
 ---
-
+Update your Cypress version by using this command:
+npm install cypress@5.0.0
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
This is not typically part of the changelog - as this assumes the package manager in use and the strategy for updates taken by the user. 

Also, this is adding this line to the top of the entire changelog regardless of version. Closing.

chghulamnabi:
I am just giving the Suggestion to add a line like this one.Because When I was a totally new to Cypress then to update the cypress version was a major issue for me.there is no Command is written in Cypress's documentation to update it's Version. It should be 'How to update the Cypress's Version.?'

PR #3102, opened 2020-08-25
Update visit.md to show that options can only be set as single parameter

john19696:
The 2 parameter form of cy.visit() generates the body of the HTTP request. Not as currently stated, which is that the second parameter is for options. This most probably is only half the story, but without going through the code in detail, this is what appears to be happening


jennifer-shehane:
@john19696 I'm not sure what you're referring to, but this is not the behavior of `cy.visit()` at all. `cy.visit()` accepts a `url` and an `options` object as specified in the docs currently. Maybe you're thinking of `cy.route()`?

PR #3105, opened 2020-08-26
Convert dependencies to a single line in Getting Started

vipulgupta2048:
Dependencies in a single line would be easier to copy and paste in the terminal though might need to ignore the 70/84 character limit per line. Not sure if the right way to make this change, but let me happy to fix up as you wish.

### DIFF
--- a/source/_partial/linux_dependencies.md
+++ b/source/_partial/linux_dependencies.md
@@ -1,13 +1,11 @@
 #### Ubuntu/Debian
 
 ```shell
-apt-get install libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev libgconf-2-4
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
@vipulgupta2048 Could you please sign our CLA?

I put it on separate lines so that a scrollbar wouldn't appear, hiding the content originally.

vipulgupta2048:
@jennifer-shehane Right, yeah. That could be solved with a documentation UI tweak, no? Hope this doesn't bring it back now.

vipulgupta2048:
Something else to look out for is if you know there are other terminal commands throughout the docs written the same way, consistency needs to be maintained overall then.

jennifer-shehane (approved):

PR #3114, opened 2020-08-28
docs: add notice about diff b/t Cypress.env and real env vars

flotwig:


source/guides/guides/environment-variables.md
```diff
@@ -123,7 +125,7 @@ Cypress.env('api_server') // 'http://localhost:8888/api/v1/'
 
 ## Option #3: `CYPRESS_*`
 
-Any environment variable on your machine that starts with either `CYPRESS_` or `cypress_` will automatically be added and made available to you.
+Any OS-level environment variable on your machine that starts with either `CYPRESS_` or `cypress_` will automatically be added and made available to you.
```
bencodezen, line 128:
```suggestion
Any OS-level environment variable on your machine that starts with either `CYPRESS_` or `cypress_` will automatically be added to Cypress' environment variables and made available to you.
```

When I first read this, I was wondering "where it would be added" and took a best guess at what the sentence was implying. Let me know if it's incorrect!

PR #3121, opened 2020-08-30
Phrase "log out" can be confusing in this doc

iNishant:
Because this doc is for cypress' module for cookies, the phrase "log out" can be confusing as cookies are commonly used for auth - login/log-out but in this context "log out" means logging...

Making just one change for now even though multiple changes are required in this doc. Want to confirm if maintainers agree with me on this or not.

### DIFF
--- a/source/api/cypress-api/cookies.md
+++ b/source/api/cypress-api/cookies.md
@@ -4,7 +4,7 @@ title: Cypress.Cookies
 
 `Cookies.preserveOnce()` and `Cookies.defaults()` enable you to control Cypress' cookie behavior.
 
-`Cookies.debug()` enables you to log out whenever any cookies are modified.
+`Cookies.debug()` enables you to log (to the console) whenever any cookies are modified.
 
 Cypress automatically clears all cookies **before** each test to prevent state from building up.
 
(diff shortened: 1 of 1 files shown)


source/api/cypress-api/cookies.md
```diff
@@ -4,7 +4,7 @@ title: Cypress.Cookies
 
 `Cookies.preserveOnce()` and `Cookies.defaults()` enable you to control Cypress' cookie behavior.
 
-`Cookies.debug()` enables you to log out whenever any cookies are modified.
+`Cookies.debug()` enables you to log (to the console) whenever any cookies are modified.
```
bencodezen, line 7:
```suggestion
`Cookies.debug()` enables you to generate logs to the console whenever any cookies are modified.
```

iNishant:
@bencodezen cool. Also just to remind you as I mentioned in the PR description, "log out" is used in multiple places in the same doc. It might not be necessary to make those changes now since we clarified the line at the top but still just a heads up

Screenshot 2020-09-01 at 2 03 02 AM

bencodezen:
Thanks for the reminder! Would you be wiling to open a new issue with the comment you just created? Want to give you credit for being the original poster!

PR #3138, opened 2020-09-08
docs(install): update pre-release install docs

flotwig:
This simplifies and updates the pre-release install documentation.

Originally, I was doing this for cypress-io/cypress#8482, but I noticed that the docs here were already out of date.

So I updated the instructions to direct users to the install comments that `cypress-bot` leaves for further instructions. This matches the existing behavior and will remain current once cypress-io/cypress#8482 is merged.


bahmutov (approved):
so DRY

PR #3149, opened 2020-09-14
Remove external link validation

amirrustam:
This build validation is not functional and significantly slows down the build process. Removing in favor of other scalable options such as external link validation only on changes locally or a link checking service.


jennifer-shehane (changes requested):
Any reason you're commenting out code instead of just removing it? Seems messy to leave around.

PR #3164, opened 2020-09-15
Add cypress-expect to the list of plugins

bahmutov:
tested, open-source, documented


bencodezen (approved):
Thanks for the PR @bahmutov!

PR #3174, opened 2020-09-19
Update key-differences.md: configuration#blockhosts url

dverbiest:
https://docs.cypress.io/guides/overview/configuration#blockhosts
should be:
https://docs.cypress.io/guides/references/configuration.html#blockHosts
I **guess** is this commit fixes the url correctly? (I wonder how the '{% url' find the relative links?)
Please review!


amirrustam:
Thank you for the contribution. This fix actually got included in another PR that included other link fixes as well.

PR #3175, opened 2020-09-21
add plugin cypress-teamcity-reporter

prma85:
PR to add the cypress-teamcity-reporter to the list of plugins.

### DIFF
--- a/source/_data/plugins.yml
+++ b/source/_data/plugins.yml
@@ -619,6 +619,12 @@
       keywords: [testrail, reporter]
       badge: community
 
+    - name: cypress-teamcity-reporter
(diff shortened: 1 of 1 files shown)


amirrustam:
@prma85 Thanks for this work. Any reason why you needed to create a new module instead of contributing changes to https://github.com/cypress-io/mocha-teamcity-reporter

prma85:
> @prma85 Thanks for this work. Any reason why you needed to create a new module instead of contributing changes to https://github.com/cypress-io/mocha-teamcity-reporter

Not really. It has been out for almost 2 years and I just update it. I didn't realize that there is one official form cypress.

jennifer-shehane (approved):

PR #3193, opened 2020-09-29
Fix incorrect comment in code sample

### LINKED ISSUE #3192: Incorrect comment in code sample, in "Introduction to Cypress", section "Mixing Async and Sync code"
## I'm submitting a...

```
[x] Bug report
[ ] Content update
```
There is a code sample in section [Mixing Async and Sync code](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Mixing-Async-and-Sync-code), for "Incorrect Usage", which says:

```javascript
  if (el.length) {              // evaluates immediately as 0
    cy.get('.another-selector')
  } else {
    // this code will never run
    // because the 'el' will never exist
    // when the code executes
    cy.get('.optional-selector')
  }
```

The comments under `else` are incorrect. That code block *will always run*, but the comment says "this code will never run".

mdunicorn:
closes #3192

### DIFF
--- a/source/guides/core-concepts/introduction-to-cypress.md
+++ b/source/guides/core-concepts/introduction-to-cypress.md
@@ -409,8 +409,8 @@ it('does not work as we expect', () => {
   if (el.length) {              // evaluates immediately as 0
     cy.get('.another-selector')
   } else {
-    // this code will never run
-    // because the 'el' will never exist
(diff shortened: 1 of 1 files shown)


jennifer-shehane (approved):
yeah, true 😅 Thanks for the catch.

PR #3201, opened 2020-09-30
docs: remove duplicate comment in route2.md

MrLeebo:
Hi there! The route2 glob example has the same comment twice. So I removed the extra one. That's it. This PR is not very interesting.


flotwig (approved):
thanks for the PR!

PR #3202, opened 2020-09-30
docs: route2 docs v2

flotwig:
Rewrote the route2 docs from scratch with a little more emphasis on the functionality of route2. Previously, it still closely followed the cy.route docs, which were completely different.

I think the general layout of these new docs is much more intuitive. Also, this fixes several technical errors in the route2 docs which I didn't notice when it was initially created.

### DIFF
--- a/source/api/commands/route2.md
+++ b/source/api/commands/route2.md
@@ -4,7 +4,7 @@ containerClass: experimental
 ---
 
 {% note warning %}
-{% fa fa-warning orange %} **This is an experimental feature. In order to use it, you must set the {% url "`experimentalNetworkStubbing`" experiments %} configuration option to `true`.** See {% issue 687 %} for more details.
+{% fa fa-warning orange %} **This is an experimental feature. In order to use it, you must set the {% url "`experimentalNetworkStubbing`" experiments %} configuration option to `true`.**
 {% endnote %}
 
 Use `cy.route2()` to manage the behavior of HTTP requests at the network layer.
--- a/source/api/commands/route2.md
+++ b/source/api/commands/route2.md
@@ -4,264 +4,261 @@ containerClass: experimental
 ---
 
 {% note warning %}
-{% fa fa-warning orange %} **This is an experimental feature. In order to use it, you must manually configure the {% url "`experimentalNetworkStubbing`" experiments %} option to `true`.** See {% issue 687 %} for more details.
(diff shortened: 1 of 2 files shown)


flotwig:
comparison of old and new outline:

### existing outline

* Comparison to cy.route()
* Syntax
* Arguments
    * url
    * routeHandler
    * method
* Yields
* Relevant Types
    * IncomingHttpRequest
    * IncomingHttpResponse
    * StaticResponse
* Examples
    * Without Stubbing
        * Wait on GET request matching url
        * Wait on matching method and url
        * Match route against any UserId (note: incorrect comment here)
        *  Use glob to match all segments
        * Modify request
        * Modify response
    * With Stubbing
        * url as a String
        * url as a RegExp
        * Response functions
        * Matching requests and routes
        * Specify the method
        * Making multiple requests to the same route
        * Fixtures
    * Options
        * Server redirect
        * Simulate delay
        * Simulate throttle
* See also

### new outline

* Comparison to cy.route()
* Syntax
* Arguments
    * url
    * method
    * routeMatcher
    * routeHandler
* Yields
* Examples
    * Waiting on a request
        * With URL
        * With RouteMatcher
        * Using the yielded object
    * Stubbing a response
        * With a string
        * With a fixture
        * With a `StaticResponse` object
    * Intercepting a request
        * Asserting on a request
        * Modifying an outgoing request
        * Returning a Promise
        * Passing a request to the next route handler
    * Intercepting a response
        * Asserting on a response
        * Returning a Promise
        * Modifying an incoming response
* See also

jennifer-shehane (approved):
There was something broken with this page since all uses of the `urlHash` tag broke the build, but I've fixed it. It was some hexo weirdness where you can't start a sentence with a url tag. 😅 Yah, I dunno why, I just have these quirks memorized at this point.

Honestly, I haven't used `route2()` yet, so I can't speak much to the accuracy or helpfulness of the examples, but I fixed some things that will hopefully make the doc more readable and easier to maintain in the future.

*Updates I made:*

- A lot of spacing, preventing the code blocks from scrolling
- Convert all markdown links to urlHash or url tags
- Fix issue that was causing urlHash to always error on build.
- Convert args sections to h3 headers
- Formatting of text/sections
- Removed all backticks from codeblock comments (this causes weird issues in hexo).
- Replace uses of ‘foo’ with more realistic example names
- Styled dashboard ad to look better on experimental page

Let me know of any remaining thoughts on my changes.

PR #3203, opened 2020-10-01
Update docs to mention exclamation marks not being supported.

jennifer-shehane:
Related to https://github.com/cypress-io/cypress/issues/8683


jennifer-shehane:
@chrisbreiding Do you think there's somewhere else this would be useful to mention?

chrisbreiding:
> Do you think there's somewhere else this would be useful to mention?

I don't think so. The changelog should be sufficient.

chrisbreiding (approved):

PR #3210, opened 2020-10-01
update link to zone.js

ivandotv:
Zone.js repository has moved. The original location is archived.


jennifer-shehane (approved):
Thanks for updating!

PR #3213, opened 2020-10-02
Added mention about lack of default method

JonahKK:
One difference that was not described before. Took me a moment to find out why my migration to `route2` failed ;).

### DIFF
--- a/source/api/commands/route2.md
+++ b/source/api/commands/route2.md
@@ -23,6 +23,7 @@ Unlike {% url "`cy.route()`" route %}, `cy.route2()`:
 
 * can intercept all types of network requests including Fetch API, page loads, XMLHttpRequests, resource loads, etc.
 * does not require calling {% url "`cy.server()`" server %} before use - in fact, `cy.server()` does not influence `cy.route2()` at all.
+* does not have method set to `GET` by default
 
 # Usage
 


jennifer-shehane (changes requested):
@JonahKK The method should match `GET` by default as mentioned in this section: https://docs.cypress.io/api/commands/route2.html#routeMatcher-RouteMatcher

Can you give an example of the unexpected behavior you encountered? We'd love to update the docs to be more helpful.

JonahKK:
@jennifer-shehane Okay, so it might be a defect then. But before I report anything, could you check this case:
Our app uses external API, because of CORS, before making GET request, app sends OPTIONS request. Now, when I don't include method in `route2` it matches both requests. You can see on the screenshots, that there is also a `validate_name` alias, that is matched just once, but `route2` responsible for spying it, has method specified.
2020-10-06_10h36_46
2020-10-06_10h36_53

jennifer-shehane:
The documentation for this was incorrect. It does not match `GET` by default, and instead matches `*`, so any method. This doc has been fixed here: https://github.com/cypress-io/cypress-documentation/pull/3235

JonahKK:
So this PR was about this difference. But in my PR the difference was mentioned in `Comparison to cy.route()` section as this change makes `cy.route2` different from `cy.route`.

jennifer-shehane:
Yah, ok I can see that.

jennifer-shehane (approved):

PR #3235, opened 2020-10-12
Make route2 routeMatcher example clearer.

sainthkh:
```js
cy.route2({
  url: 'http://example.com/search',
  query: { q: 'expected terms' },
}).as('login')

// once a POST request to http://example.com/login with a querystring containing
// 'q=expected+terms' responds, this 'cy.wait' will resolve
cy.wait('@login')
```

The alias name and comments sound wrong.

* url uses `search` but the comment says `http://example.com/login`
* By default, routeMatcher uses `GET` method, not `POST`.
* The better name for the alias is search, not login. 

I changed 3 of them.

### DIFF
--- a/source/api/commands/route2.md
+++ b/source/api/commands/route2.md
@@ -174,11 +174,11 @@ cy.wait('@getSettings')
 cy.route2({
   url: 'http://example.com/search',
   query: { q: 'expected terms' },
-}).as('login')
+}).as('search')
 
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
I think the fact that GET matching is default may be improperly documented actually. I think it matches `*` by default.

jennifer-shehane (approved):
Thanks!

PR #3236, opened 2020-10-12
Using an image fixture for upload

nassimerrahoui:
Other way to upload an image fixture without using jquery plugin


jennifer-shehane (approved):
Checked that this works in latest version of cypress 👍 Thanks!

PR #3251, opened 2020-10-19
Document missing commands that also follow actionability rules

jennifer-shehane:
- Some commands were missing as documented under the actionability checks. Found while reviewing https://github.com/cypress-io/cypress/pull/8837 I added those.


lukeapage:
submit, focus (and blur which isn't here) do not strictly follow the actionability in the docs - they do accept a force parameter but they don't do things mentioned on this page, like scroll to the element.

jennifer-shehane:
K, that's why I was asking for review. I wasn't 100% sure. Thanks!

PR #3284, opened 2020-10-25
Clarification on return value

Doogiemuc:
There are many situations when beginners might stumble over this. Here is one example:

cy.get('#domElemWithTextContent').should('have.length.of.at.least', 5)      <= DOES NOT WORK!!!   this checks for the length of the array not the length of the text!
cy.get('#domElemWithTextContent').invoke('text').should('have.length.of.at.least', 5)         <= Only this works.

### DIFF
--- a/source/api/commands/get.md
+++ b/source/api/commands/get.md
@@ -58,6 +58,8 @@ Option | Default | Description
 
 {% yields sets_dom_subject cy.get %}
 
+Keep in mind, that cy.get() **always** returns an Array of DOM elements. (Same as JQuery does.) All chained methods work on all returned elements.
+
 # Examples
 
 ## Selector


source/api/commands/get.md
```diff
@@ -58,6 +58,8 @@ Option | Default | Description
 
 {% yields sets_dom_subject cy.get %}
 
+Keep in mind, that cy.get() **always** returns an Array of DOM elements. (Same as JQuery does.) All chained methods work on all returned elements.
```
jennifer-shehane, line 61: `cy.get()`, as well as many other commands, actually return a jQuery Object, not an array. https://learn.jquery.com/using-jquery-core/jquery-object/

jennifer-shehane (changes requested):

jennifer-shehane:
Unfortunately we have to close this PR due to inactivity. Please open a new PR addressing the original issue and any requested changes.

PR #3305, opened 2020-11-03
Add link to Mocha custom reporters docs

Izhaki:
To make clear what the API of such reporters is.
We were trying to write a custom cucumber reporter, not adhering to the requirements of Mocha reporters. We go an error "Could not load reporter by name" even though the correct paths was listed in "We searched for the reporter in these paths".


jennifer-shehane (approved):
Great idea!

PR #3323, opened 2020-11-12
docs(net-stubbing): deprecate cy.route(), remove experimentalNetworkStubbing

flotwig:
for https://github.com/cypress-io/cypress/pull/9185

- in addition to marking cy.route() and cy.server() as deprecated and removing experimentalNetworkStubbing...
- ...also updated existing documentation around network requests to refer to cy.http instead of cy.route/cy.server, where applicable
	- some older external links will need to be updated separately/kept for reference


jennifer-shehane:
Trying to make this gigantic warning on cy.route()/cy.server() more helpful and readable. This is what I got. Not done reviewing currently.

Screen Shot 2020-11-19 at 1 54 00 PM

source/api/commands/exec.md
```diff
@@ -89,12 +89,11 @@ cy.exec('npm run my-script').its('stdout').should('contain', 'Done running the s
...
-cy.server()
-cy.route('POST', '/comments').as('postComment')
+cy.http('POST', '/comments').as('postComment')
 cy.get('.add-comment').click()
-cy.wait('@postComment').then((xhr) => {
-  cy.exec(`echo ${JSON.stringify(xhr.responseBody)} >cypress/fixtures/comment.json`)
-  cy.fixture('comment.json').should('deep.eq', xhr.responseBody)
+cy.wait('@postComment').then(({ response }) => {
+  cy.exec(`echo ${JSON.stringify(response.body)} >cypress/fixtures/comment.json`)
+  cy.fixture('comment.json').should('deep.eq', response.body)
```
jennifer-shehane, line 96: This example doesn't work, so is not the equivalent of the previous version. The assertion seems to fail now because there are newlines in the body. (I'm still reviewing but this is one example of a change needed that I've run into)

```js
  it('cy.route() - passes', () => {
    cy.visit('https://example.cypress.io/commands/network-requests')
    cy.server()
    cy.route('POST', '/comments').as('postComment')
    cy.get('.network-post').click()
    cy.wait('@postComment').then((xhr) => {
      cy.exec(`echo ${JSON.stringify(xhr.responseBody)} >cypress/fixtures/comment1.json`)
      cy.fixture('comment.json').should('deep.eq', xhr.responseBody)
    })
  })

  it('cy.http() - fails', () => {
    cy.visit('https://example.cypress.io/commands/network-requests')
    cy.http('POST', '**/comments').as('postComment')
    cy.get('.network-post').click()
    cy.wait('@postComment').then(({ response }) => {
      cy.exec(`echo ${JSON.stringify(response.body)} >cypress/fixtures/comment2.json`)
      cy.fixture('comment.json').should('deep.eq', response.body)
    })
  })
```

Screen Shot 2020-11-19 at 8 17 43 PM

source/guides/guides/network-requests.md
```diff
@@ -358,31 +314,30 @@ cy.get('#results')
...
-  .its('requestBody') // alternative: its('request.body')
-  .should('deep.equal', {
+cy.get('@new-user') // yields the same request object
+  .its('request.body')
+  .should('deep.equal', JSON.stringify({
     id: '101',
     firstName: 'Joe',
     lastName: 'Black'
-  })
+  }))
```
jennifer-shehane, line 339: This example doesn't work either.

```js
it('cy.http() - route responses to matching requests', () => {
  cy.visit('https://example.cypress.io/commands/network-requests')

  cy.http('POST', '/comments').as('postComment')
  cy.get('.network-post').click()

  cy.wait('@postComment')
    .should('have.property', 'response.statusCode', 200)

  cy.get('@postComment') // yields the same request object
    .its('request.body')
    .should('deep.equal', JSON.stringify({
      id: '101',
      firstName: 'Joe',
      lastName: 'Black',
    }))
```

Screen Shot 2020-11-20 at 2 30 47 PM

source/guides/guides/network-requests.md
```diff
@@ -358,31 +314,30 @@ cy.get('#results')
...
+  .should(({ request, response }) => {
+    expect(request.url).to.match(/\/users$/)
+    expect(request.method).to.equal('POST')
     // it is a good practice to add assertion messages
     // as the 2nd argument to expect()
-    expect(xhr.response.headers, 'response headers').to.include({
+    expect(response.headers, 'response headers').to.include({
       'cache-control': 'no-cache',
       expires: '-1',
       'content-type': 'application/json; charset=utf-8',
```
jennifer-shehane, line 351: maybe there's something wrong with my examples? This doesn't work for me either. 

```js
  it('cy.http() - route responses to matching requests', () => {
    cy.visit('https://example.cypress.io/commands/network-requests')

    cy.http('POST', '/comments').as('postComment')
    cy.get('.network-post').click()

    cy.wait('@postComment')

    cy.get('@postComment')
    .should(({ request, response }) => {
      expect(request.url).to.match(/\/users$/)
      expect(request.method).to.equal('POST')
      // it is a good practice to add assertion messages
      // as the 2nd argument to expect()
      expect(response.headers, 'response headers').to.include({
        'cache-control': 'no-cache',
        expires: '-1',
        'content-type': 'application/json; charset=utf-8',
      })
    })
})
```

Screen Shot 2020-11-20 at 2 33 17 PM

jennifer-shehane (changes requested):
Found some examples that don't seem to work for me, primarily when trying to assert off of the request / response. 

*A comment on the feature in general:* I don't like that you have to use JSON.parse() to make assertions on the response body. I think this is a poorer experience than how you could assert of cy.route() and will trip up a lot of people when they move over.

```js
// was
expect(xhr.responseBody).to.have.property('name', 'Using POST in cy.route()')

// now
expect(JSON.parse(String(response.body))).to.have.property('name', 'Using POST in cy.http()')
```

source/guides/guides/network-requests.md
```diff
@@ -308,44 +274,34 @@ In this example, there are many possible sources of failure. In most testing too
...
 
-<!--
-To reproduce the following screenshot:
-it('test', () => {
-  cy.server()
-  cy.route('foo/bar').as('getSearch')
-  cy.wait('@getSearch')
-})
--->
-
```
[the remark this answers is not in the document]
flotwig: I see, I missed the first line of this comment, my b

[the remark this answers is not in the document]
flotwig: Thought it was unused code :laughing:

flotwig:
> A comment on the feature in general: I don't like that you have to use JSON.parse() to make assertions on the response body. I think this is a poorer experience than how you could assert of cy.route() and will trip up a lot of people when they move over.

Totally agree, unfortunately changing it now would be breaking change... I thought about using the `content-type` to automatically parse JSON, for example.

jennifer-shehane:
I'm going to merge this in because the conflicts in 6.0 branch are getting out of hand for me, but these examples still need addressing.

jennifer-shehane:
@flotwig Wouldn't 6.0 be the place to make breaking changes?

PR #3333, opened 2020-11-18
The egghead cypress course is not free

henrique-barbase:
the egghead cypress course is not free


jennifer-shehane (approved):
True. Thanks for updating

PR #3350, opened 2020-11-24
Fix example aliasing example; Add "Aliasing individual GraphQL requests" section with additional examples

kevinold:
Fix original aliasing example (`req.body.includes("mutation") to req.body.hasOwnProperty("mutation")`) since `req.body.includes` throws an "includes is not a function" error if run in Cypress.

Rename section to be GraphQL specific and add variations of `routeMatcher` for queries and mutations with and without an `operationName` attribute.

> The operation name is a meaningful and explicit name for your operation. It is only required in multi-operation documents, but its use is encouraged because it is very helpful for debugging and server-side logging.

 [Operation Name | GraphQL](https://graphql.org/learn/queries/#operation-name)

### DIFF
--- a/source/api/commands/intercept.md
+++ b/source/api/commands/intercept.md
@@ -217,7 +217,7 @@ cy.wait('@gqlMutation')
 
 ### Aliasing individual GraphQL requests
 
-Aliases can be set on a per-request basis by setting the `alias` property of the intercepted request:
+Aliases can be set on a per-request basis by setting the `alias` property of the intercepted request.
 
 This is useful against GraphQL endpoints to wait for specific Queries and Mutations.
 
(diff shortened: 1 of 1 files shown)


kevinold:
@jennifer-shehane @amirrustam @flotwig   Let me know your thoughts on adjusting this area of the docs.  If we do not want to flush this out into a GraphQL specific section, I will update to only include the fix for `req.body.includes("mutation") to req.body.hasOwnProperty("mutation")`.

flotwig:
@kevinold since this is the only place that `req.alias` is documented, can we keep an "Aliasing individual requests" section and either just update it with `.hasOwnProperty` or add a new section for the graphql example?

kevinold:
> @kevinold since this is the only place that `req.alias` is documented, can we keep an "Aliasing individual requests" section and either just update it with `.hasOwnProperty` or add a new section for the graphql example?

Absolutely!  I'll break it out now.

kevinold:
@flotwig Updated.

flotwig (approved):
nice :)

PR #3353, opened 2020-11-25
fix: fix bug about add extensions

iam-frankqiu:
Fixed a bug about how to load extensions in chrome. According to this article: https://www.cypress.io/blog/2020/01/07/how-to-load-the-react-devtools-extension-in-cypress/#load-the-extension. close [#9304](https://github.com/cypress-io/cypress/issues/9304)

### DIFF
--- a/source/api/plugins/browser-launch-api.md
+++ b/source/api/plugins/browser-launch-api.md
@@ -7,7 +7,7 @@ Before Cypress launches a browser, it gives you the opportunity to modify the br
 # Syntax
 
 ```js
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
The blog article is outdated, not our docs. Our docs always reflect the current API of the latest version. This API changed in 4.0. See our migration guide here: https://on.cypress.io/migration-guide#Plugin-Event-before-browser-launch

I can let our marketing team know about the blog being out of date. Thanks.

PR #3360, opened 2020-11-30
Add casting to get rid of error

My-:
Add Casting as IDE shows error otherwise.
Error: Typescript Error: Property 'files' does not exist on type 'HTMLElement'
Fix reference: https://stackoverflow.com/questions/49827325/typescript-error-property-files-does-not-exist-on-type-htmlelement


jennifer-shehane:
This would only be relevant for TypeScript users. Not all of our users use TypeScript.

PR #3369, opened 2020-12-02
describe how to access the retry counter from the test

bahmutov:
seems like plugins writers might need this information, I needed it to write https://github.com/cypress-io/cypress-repeat-retry

### DIFF
--- a/source/guides/guides/test-retries.md
+++ b/source/guides/guides/test-retries.md
@@ -224,3 +224,14 @@ it('does something differently on retry', { retries: 3 }, () => {
 ```
 
 The above `attempt` variable will have values 0 through 3 (the first default test execution plus three allowed retries). The `retries` constant in this case is always 3.
+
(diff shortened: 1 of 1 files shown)


source/guides/guides/test-retries.md
```diff
@@ -207,3 +207,24 @@ No. Tests recorded during `cypress run` with the `--record` flag will be counted
...
+  // cy.state('runnable') returns the current test object
+  // we can grab the current attempt and
+  // the total allowed attempts from its properties
+  const attempt = cy.state('runnable')._currentRetry
+  const retries = cy.state('runnable')._retries
+
+  if (attempt < retries) {
+    expect(true).to.be.false // fail the test
+  }
+  // passes on the last attempt
```
jennifer-shehane, line 226: I think including this assertion is just going to mislead people when they copy/paste this - as this is not a use case anyone should ever need to actually write. 

```suggestion
```

jennifer-shehane (changes requested):

PR #3381, opened 2020-12-08
Add section for migration cy.route to cy.intercept

jennifer-shehane:
TR-553

- Added a section to the migration guide for migration cy.route() to cy.intercept(). 
- I think these examples are pretty minimal and I'd probably like more complicated examples added.
- Mostly to just allow us to have a place to link and share with people, but I imagine this section will move into the 7.0 migration guide when that is released.


flotwig (approved):
Looks good

PR #3393, opened 2020-12-11
Add doc for before:spec event

chrisbreiding:
For https://github.com/cypress-io/cypress/pull/9646

When I implement the `after:spec` event and add a doc for that, I'll add a 'Spec Events' section to the [Plugins Guide](https://github.com/cypress-io/cypress-documentation/blob/develop/source/guides/tooling/plugins-guide.md).

### DIFF
--- a/source/api/plugins/before-spec-api.md
+++ b/source/api/plugins/before-spec-api.md
@@ -4,11 +4,10 @@ title: before:spec
 
 The `before:spec` event fires before a spec file is run.
 
-The event timing is different depending on if Cypress is run via `cypress run` (run mode) or `cypress open` (interactive mode).
+The event timing is different depending on if Cypress is run via `cypress run` or `cypress open`.
 
-In run mode, the event is fired as soon as possible when a spec is about to be run.
-
-In interactive mode, since there is not a discreet before or after to a spec running, the event is fired once before the browser is launched for a spec.
+- During `cypress run`, the event is fired as soon as possible when a spec is about to run.
+- During `cypress open`, since there is not a discreet before or after to a spec running, the event is fired once before the browser is launched for a spec.
 
 # Syntax
 
@@ -47,7 +46,7 @@ module.exports = (on, config) => {
 
 ## Log the absolute spec path, but only when in run mode
 
-You can utilize `config.isInteractive` flag to determine if currently running via `cypress run` (run mode) or `cypress open` (interactive mode).
+You can utilize the `config.isInteractive` flag to determine if the spec is currently running via `cypress run` (run mode) or `cypress open` (interactive mode).
 
 ```javascript
 module.exports = (on, config) => {
@@ -59,3 +58,8 @@ module.exports = (on, config) => {
   })
 }
 ```
+
+# See also
+
+- {% url "Plugins Guide" plugins-guide %}
+- {% url "Writing a Plugin" writing-a-plugin %}
--- a/source/api/plugins/before-spec-api.md
+++ b/source/api/plugins/before-spec-api.md
@@ -26,6 +26,8 @@ Property | Description
 
 # Usage
 
+You can return a promise from the `before:spec` event handler and it will be awaited before Cypress proceeds running the spec.
+
 ## Log the relative spec path to stdout before the spec is run
 
 ```javascript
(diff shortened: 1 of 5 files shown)


chrisbreiding:
Failing test caused by this: https://github.com/cypress-io/cypress-documentation/commit/50c50d0649f91b1151d7b2f19e6e468edd9738a3. I made sure Gleb is aware of it.

source/api/plugins/before-spec-api.md
```diff
@@ -0,0 +1,61 @@
...
+    // }
+
+    console.log('Running', spec.relative)
+  })
+}
+```
+
+## Log the absolute spec path, but only when in run mode
+
+You can utilize `config.isInteractive` flag to determine if currently running via `cypress run` (run mode) or `cypress open` (interactive mode).
```
jennifer-shehane:
```suggestion
You can utilize the `config.isInteractive` flag to determine if the spec is currently running via `cypress run` (run mode) or `cypress open` (interactive mode).
```

source/api/plugins/before-spec-api.md
```diff
@@ -0,0 +1,61 @@
...
+title: before:spec
+---
+
+The `before:spec` event fires before a spec file is run.
+
+The event timing is different depending on if Cypress is run via `cypress run` (run mode) or `cypress open` (interactive mode).
+
+In run mode, the event is fired as soon as possible when a spec is about to be run.
+
+In interactive mode, since there is not a discreet before or after to a spec running, the event is fired once before the browser is launched for a spec.
```
jennifer-shehane:
```suggestion
- In **run mode**, the event is fired as soon as possible when a spec is about to be run.
- In **interactive mode**, since there is not a discreet before or after to a spec running, the event is fired once before the browser is launched for a spec.
```

source/api/plugins/before-spec-api.md
```diff
@@ -0,0 +1,65 @@
+---
+title: before:spec
+---
+
+The `before:spec` event fires before a spec file is run.
+
+The event timing is different depending on if Cypress is run via `cypress run` or `cypress open`.
+
+- During `cypress run`, the event is fired as soon as possible when a spec is about to run.
+- During `cypress open`, since there is not a discreet before or after to a spec running, the event is fired once before the browser is launched for a spec.
```
jennifer-shehane, line 10: Would like some clarification on how this works when running 'Run all intergration specs / run all component tests'

jennifer-shehane (changes requested):
I can't remember exactly how we planned to document these the last time we planned to do this. We were going to have a single page for each event?

I was thinking there would be a 'Specs' page, where it would have `before:spec` and `after:spec` documented. If that's what we plan to do, I would prefer to skip this intermediary 'before-spec' page and just put this in a Specs page. It will just end up being some dead link we have to fix up later and reroute. 

Would love to hear your thoughts. As it is now this page is not displaying in the sidebar at all. 

Screen Shot 2020-12-14 at 10 27 59 AM


I would also like a better example of use for this than 'console.log', but I'm not really sure what to show here either. The only thing I can think of is some 'initialize' command, like seed stuff for a database? that you later clean up in an after:spec.

chrisbreiding:
> I can't remember exactly how we planned to document these the last time we planned to do this. We were going to have a single page for each event?

> I was thinking there would be a 'Specs' page, where it would have before:spec and after:spec documented. If that's what we plan to do, I would prefer to skip this intermediary 'before-spec' page and just put this in a Specs page. It will just end up being some dead link we have to fix up later and reroute.

I made it its own page because every other plugin event has its own page. I don't really know how to structure it with syntax/examples/usage for each event if we're putting them on a single page. I'll be getting `after:spec` done in the next day or so, so maybe let's punt on it for now and figure it out in the `after:spec` PR since then we'll have its content to work with.

> I would also like a better example of use for this than 'console.log', but I'm not really sure what to show here either. The only thing I can think of is some 'initialize' command, like seed stuff for a database? that you later clean up in an after:spec.

I'd like a better example too, but I don't really have one. I'm not sure it makes sense to seed a database with this event, so I'd rather not mislead users by creating an example with that. 

> We'll also need a new 'on' link for any new page we decide on.

I'll submit a PR once we decide on which direction we're going with one versus two pages.

jennifer-shehane:
Yah ok, I see they all have their own page now. We just never had multiple events for a similarly themed task before like 'before:spec', 'after:spec'.

jennifer-shehane (approved):

PR #3424, opened 2020-12-18
docs(intercept): document matchUrlAgainstPath

flotwig:
For https://github.com/cypress-io/cypress/pull/14241

### DIFF
--- a/source/api/commands/intercept.md
+++ b/source/api/commands/intercept.md
@@ -622,6 +622,7 @@ The available functions on `res` are:
 ```
 
 {% history %}
+{% url "6.2.0" changelog#6-2-0 %} | Added `matchUrlAgainstPath` option to `RouteMatcher`.
 {% url "6.0.0" changelog#6-0-0 %} | Renamed `cy.route2()` to `cy.intercept()`.
 {% url "6.0.0" changelog#6-0-0 %} | Removed `experimentalNetworkStubbing` option and made it the default behavior.
 {% url "5.1.0" changelog#5-1-0 %} | Added experimental `cy.route2()` command under `experimentalNetworkStubbing` option.
--- a/source/api/commands/intercept.md
+++ b/source/api/commands/intercept.md
@@ -7,7 +7,7 @@ Use `cy.intercept()` to manage the behavior of HTTP requests at the network laye
 With `cy.intercept()`, you can:
 
 * stub or spy on any type of HTTP request.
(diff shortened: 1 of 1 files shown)


source/api/commands/intercept.md
```diff
@@ -82,6 +82,11 @@ All properties are optional. All properties that are set must match for the rout
    * If 'false', only HTTP requests will be matched.
    */
   https?: boolean
+  /**
+   * If `true`, will match the supplied `url` against incoming `path`s.
+   * Cannot be set without `url` or with `path`.
```
jennifer-shehane, line 87:
```suggestion
   * Requires a `url` argument. Cannot be used with a `path` argument.
```

source/api/commands/intercept.md
```diff
@@ -159,6 +164,19 @@ The `routeHandler` defines what will happen with a request if the {% urlHash "`r
 
 ## Matching URL
 
+{% note info %}
+**Note:** passing a URL as a string or RegExp to `cy.intercept()` will automatically set `matchUrlAgainstPath` to `true`. This means that the supplied string or RegExp will be matched against the **path** if matching against the **URL** fails.
+{% endnote %}
```
jennifer-shehane, line 169: I moved the note to the top of the section and combined them.

source/api/commands/intercept.md
```diff
@@ -159,6 +164,19 @@ The `routeHandler` defines what will happen with a request if the {% urlHash "`r
...
+
+You can provide the entire URL to match
+
+```js
+// will match any request that exactly matches the URL
+//   matches GET https://prod.cypress.io/users
+//   won't match GET https://staging.cypress.io/users
+cy.intercept('https://prod.cypress.io/users')
+```
```
jennifer-shehane, line 178: Added this example since there wasn't a full URL example.

jennifer-shehane (commented):
Might want to add an example for `matchUrlAgainstPath: false`. I'm not sure how useful you see this being though, so that's up to you. 

Will need matchUrlAgainstPath in the History table also. I assume it'll get into 6.2.0.

flotwig:
> Might want to add an example for matchUrlAgainstPath: false. I'm not sure how useful you see this being though, so that's up to you.

I see `matchUrlAgainstPath` as something that users will be using implicitly 99% of the time, not explicitly, so I don't feel that an extra example is needed.

jennifer-shehane (approved):

PR #3425, opened 2020-12-18
Add doc for before:run event

chrisbreiding:
For https://github.com/cypress-io/cypress/pull/14238

### DIFF
--- a/source/api/plugins/before-run-api.md
+++ b/source/api/plugins/before-run-api.md
@@ -1,0 +1,78 @@
+---
+title: Before Run API
+---
+
+The `before:run` event fires before a run starts. The event only fires when running via `cypress run`.
+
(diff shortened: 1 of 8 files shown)


source/api/plugins/before-run-api.md
```diff
@@ -0,0 +1,78 @@
+---
+title: Before Run API
```
jennifer-shehane, line 2: you know what? I'm just realizing all of these pages should be experimental, so we have a class that makes the page look 'experimental' - it basically makes the whole page orange. 

Can you add this to the other event pages also?

```suggestion
title: Before Run API
containerClass: experimental
```

jennifer-shehane (approved):

source/api/plugins/before-run-api.md
chrisbreiding: Going to do this in https://github.com/cypress-io/cypress-documentation/pull/3428

PR #3432, opened 2020-12-22
Update reporters.md

amandaflagg:
command line command provided for running mochawesome was not sending files to `cypress/results`, it was incorrectly routing to a new report directory named `reporter-config.json`

### DIFF
--- a/source/guides/tooling/reporters.md
+++ b/source/guides/tooling/reporters.md
@@ -215,7 +215,7 @@ We need to configure the reporter in your {% url "configuration file (`cypress.j
 
 ```shell
 cypress run --reporter mochawesome \
-  --reporter-options reportDir=results,overwrite=false,html=false,json=true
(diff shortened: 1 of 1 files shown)


source/guides/tooling/reporters.md
```diff
@@ -215,7 +215,7 @@ We need to configure the reporter in your {% url "configuration file (`cypress.j
 
 ```shell
 cypress run --reporter mochawesome \
-  --reporter-options reportDir=reporter-config.json,overwrite=false,html=false,json=true
+  --reporter-options reportDir=results,overwrite=false,html=false,json=true
```
jennifer-shehane, line 218: Shouldn't this be `cypress/results` to match the config file above?

```suggestion
  --reporter-options reportDir="cypress/results",overwrite=false,html=false,json=true
```

jennifer-shehane (changes requested):

PR #3483, opened 2021-01-12
add common problems section to the configuration reference

bahmutov:
with `baseUrl` inside the `env` block

Screen Shot 2021-01-12 at 9 24 58 AM


jennifer-shehane (changes requested):
I think it's generally a bad idea to put non-working example in code in the docs. People do not read and I've seen them copy non-working examples before. 

I'm kind of not a fan of documenting every way that you could write Cypress wrong in general.

bahmutov:
I think this is an acceptable approach: clearly marking the problematic way. At least the users can see that. I would rather have something we can reply with when the base URL is set incorrectly, that respond by writing the same text again and again and again.

Screen Shot 2021-01-27 at 2 25 01 PM

bahmutov:
I updated the description to make "will not work" vs "the correct way" more prominent

jennifer-shehane (approved):

PR #3494, opened 2021-01-19
Fix example in the route -> intercept migration guide

DominikPalo:
There is a mistake in one of the examples (there should be `intercept` instead of `route` command)


bahmutov:
nice, thank you for noticing and taking the time to fix it

PR #3495, opened 2021-01-19
add note about cy.request bypassing cy.intercept

bahmutov:
To answer https://github.com/cypress-io/cypress/discussions/14631


CypressCecelia (approved):
👍

PR #3512, opened 2021-01-22
Update cypress-vue-unit-test links to @cypress/vue

cmitz:
I've updated links to example folders in the component section of the documentation: https://docs.cypress.io/guides/component-testing/introduction.html#What-is-Cypress-Component-Testing

The examples still pointed to the old cypres-vue-unit-test package, while that package was adopted by the `@cypress` organization a short time ago.

The "install" command for `yarn add` was also outdated, so it points to `@cypess/vue` too.


jennifer-shehane (approved):
Thanks!

PR #3514, opened 2021-01-22
add multiple assertion advice

bahmutov:
because it could be confusing


bahmutov:
example tested in https://glebbahmutov.com/cypress-examples/6.3.0/commands/assertions.html#all-assertions-must-pass-at-once

jennifer-shehane (approved):
Reworded the 'subject' reference portion a bit.

PR #3525, opened 2021-01-28
clarify spec parameter is relative to the current working directory

bahmutov:
for confusion reported by the user https://github.com/cypress-io/cypress/issues/14532

### DIFF
--- a/source/guides/guides/command-line.md
+++ b/source/guides/guides/command-line.md
@@ -303,7 +303,7 @@ cypress run --reporter junit --reporter-options mochaFile=result.xml,toConsole=t
 
 ### `cypress run --spec <spec>`
 
-Run tests specifying a single test file to run instead of all tests.
+Run tests specifying a single test file to run instead of all tests. The path should be relative to the current working directory.
 
 ```shell
 cypress run --spec "cypress/integration/examples/actions.spec.js"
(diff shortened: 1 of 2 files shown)


source/guides/guides/command-line.md
```diff
@@ -303,7 +303,7 @@ cypress run --reporter junit --reporter-options mochaFile=result.xml,toConsole=t
 
 ### `cypress run --spec <spec>`
 
-Run tests specifying a single test file to run instead of all tests.
+Run tests specifying a single test file to run instead of all tests. The path should be relative to the current working directory.
```
jennifer-shehane, line 306: I thought this needs to be relevant to the project root? Not necessarily the cwd.

jennifer-shehane (changes requested):

source/guides/guides/command-line.md
bahmutov: it is relative to the current folder. For example, let's say we are outside the folder completely and use `--project` parameter. Notice how the `--spec` still uses relative to the CWD, not to the project:

Screen Shot 2021-01-29 at 8 57 39 AM

We can also use the absolute path to spec

Screen Shot 2021-01-29 at 9 00 40 AM

But not the path relative to the project's root

Screen Shot 2021-01-29 at 9 00 10 AM

bahmutov: I have updated the file and added the clarification and more examples

jennifer-shehane: Oh, seems like the error we were going to throw always said it was relative to the project directory though.

jennifer-shehane (approved):

PR #3557, opened 2021-02-11
Fix `toMatchImageSnapshot` example with image settings

AMS777:
On the `toMatchImageSnapshot` example for visual testing, the documentation places the image settings on the top level of the configuration, like this:

```
  cy.get('.todoapp').toMatchImageSnapshot({
    threshold: 0.001,
  })
```

The image settings must be inside the `imageConfig` object:

```
  cy.get('.todoapp').toMatchImageSnapshot({
    imageConfig: {
      threshold: 0.001,
    },
  })
```


amirrustam (approved):
Thanks for catching!

PR #3560, opened 2021-02-12
Dashboard Jira Integration Guide

kevinold:


amirrustam (changes requested):
I made some updates to the intro and instructions, besides that all looks good. The only thing remaining is the installation section, which we can't do at the moment while we wait for Atlassian marketplace approval.

PR #3561, opened 2021-02-14
Update cypress-studio.md - small fix for one paragraph.

kareman:


source/guides/core-concepts/cypress-studio.md
```diff
@@ -148,7 +148,7 @@ describe('Cypress Studio Demo', () => {
 
 ## Adding a New Test
 
-You can add a new test to any existing `describe` or `context` block. Next, we can add a new test, by clicking "Add New Test" our defined `describe` block.
+You can add a new test to any existing `describe` or `context` block, by clicking "Add New Test" on our defined `describe` block.
```
panzarino, line 151:
```suggestion
You can add a new test to any existing `describe` or `context` block, and we'll do so by clicking "Add New Test" on our defined `describe` block.
```

So I agree that the original version is a bit wordy, but I think these changes should still try to encapsulate how we're walking through a demo example since this is a guide.

amirrustam (approved):

PR #3574, opened 2021-02-18
Changed the react documentation

Xvier:
The react documentation was still using the old plugin references. I changed that to the current state of the project.


jennifer-shehane (approved):
Thanks!

PR #3576, opened 2021-02-18
clarify adding headers in cy.intercept

bahmutov:
to resolve the confusion https://github.com/cypress-io/cypress/issues/15139


jennifer-shehane (approved):
I feel like there are some other issues we can close after I see this caveat - because I've seen a few other issues where they're pointing at the network tab not having the right headers.

bahmutov:
Yup! I was just reminded when I saw this issue was opened that we need to clarify this

PR #3622, opened 2021-03-01
fix: various fixes

elevatebart:
- [x] When vscode git engine was refreshing the load, Nuxt was watching it. I added .git to the nuxtignore
- [x] aslant was set to use vue3 while we use vue2, I fixed it
- [x] Code group was failing when using it because of slots and because of a css scoping issue, fixed it too


matthamil (approved):
looks good to me :+1:

PR #3655, opened 2021-03-08
Update docker.md

piopi:
Hi,
I have an example of my repo where I added a desktop environment and noVNC to the cypress Docker image and wanted to share this example with the community.

Thank you,


amirrustam:
@piopi what's the intended use-case of this Docker image?

piopi:
@amirrustam 

It allows the users to view Cypress ("Interactive mode") in the container without the need to install a VNC client or X11 server on the host machine. They only need to navigate with their browser to the container IP:PORT and they will be able to view the GUI of Cypress.

amirrustam (approved):

PR #3684, opened 2021-03-11
docs: add section on CT specific config

lmiller1990:
Until we support a way to configure both test runners in a single file (such `cypress.config.js`, proposal [here](https://github.com/cypress-io/cypress/pull/15280)), we recommend having a separate config and plugins file for the CT runner.

`cypress.config.js` is certainly ideal, but there are some details that need to be sorted out. This is a simple alternative for alpha users. Once we have something like `cypress.config.js`, we will remove this recommendation.

Ticket: https://cypress-io.atlassian.net/browse/CT-338


lmiller1990:
Closing due to suckage

PR #3688, opened 2021-03-12
Remove duplicated sentence

Tbaut:
Removes a sentence that is repeated twice in the doc (see the line below).

ps: Congrats, your documentation is really good, it's such a pleasure to read and follow :tada:


bahmutov (changes requested):
I don't think this line is a duplicate, I think this is an intentional description of "learn more" link that is present in some error messages

Tbaut:
This is actually what the line below says, with almost the exact same words "some error messages contain a Learn More..."

bahmutov:
I see what you mean, good catch

bahmutov (approved):
good catch

PR #3689, opened 2021-03-12
add assertion example disabled and enabled

bahmutov:
since users are confused https://twitter.com/manekinekko/status/1370406934063558658


manekinekko:
This is awesome. Thank you @bahmutov ❤️

amirrustam (approved):

PR #3730, opened 2021-03-23
fix: add cy.then behavior when raw HTMLElements are used.

sainthkh:
Related with cypress-io/cypress#15624


amirrustam:
Since we now have a new docs site, this PR needs to be ported over.

PR #3742, opened 2021-03-25
Add RWA cy.intercept use cases

kevinold:
Document use cases from the RWA (https://github.com/cypress-io/cypress-realworld-app/pull/676) migration to `cy.intercept`

PR #3756, opened 2021-03-29
restore setThrottle example from #3742

kevinold:
Restore throttle example missing from merge of #3742

### DIFF
--- a/content/api/commands/intercept.md
+++ b/content/api/commands/intercept.md
@@ -584,6 +584,34 @@ cy.intercept('/users', (req) => {
 })
 ```
 
+#### Throttle or delay response all incoming responses
+
+You can throttle or delay all incoming responses using a `beforeEach()` in the `cypress/support/index.js` file
+
+```ts
+// Code from Real World App (RWA)
+// cypress/support/index.ts
+import { isMobile } from './utils'
+import './commands'
+// Throttle API responses for mobile testing to simulate real world conditions
+if (isMobile()) {
+  cy.intercept({ url: 'http://localhost:3001', middleware: true }, (req) => {
+    req.on('response', (res) => {
+      // Throttle the response to 1 Mbps to simulate a mobile 3G connection
+      res.setThrottle(1000)
+    })
+  })
+}
+```
+
+<Alert type="info">
+
+##### <Icon name="graduation-cap"></Icon> Real World Example
+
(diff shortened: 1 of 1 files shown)


content/api/commands/intercept.md
```diff
@@ -584,6 +584,34 @@ cy.intercept('/users', (req) => {
...
+You can throttle or delay all incoming responses using a `beforeEach()` in the `cypress/support/index.js` file
+
+```ts
+// Code from Real World App (RWA)
+// cypress/support/index.ts
+import { isMobile } from './utils'
+import './commands'
+// Throttle API responses for mobile testing to simulate real world conditions
+if (isMobile()) {
+  cy.intercept({ url: 'http://localhost:3001', middleware: true }, (req) => {
```
flotwig, line 598: This won't match, right. shouldn't it be...

```suggestion
  cy.intercept({ url: 'http://localhost:3001/*', middleware: true }, (req) => {
```

content/api/commands/intercept.md
```diff
@@ -584,6 +584,34 @@ cy.intercept('/users', (req) => {
...
+#### Throttle or delay response all incoming responses
+
+You can throttle or delay all incoming responses using a `beforeEach()` in the `cypress/support/index.js` file
+
+```ts
+// Code from Real World App (RWA)
+// cypress/support/index.ts
+import { isMobile } from './utils'
+import './commands'
+// Throttle API responses for mobile testing to simulate real world conditions
```
flotwig, line 596: it feels confusing to reference these imports from RWA in the example section of these docs, idk, is this necessary/is there a better way to present this?

kevinold: Sure, but we have them elsewhere in the guide and want to direct folks from the docs to the RWA for a runnable example.  @amirrustam Might have more thoughts.

amirrustam (approved):

content/api/commands/intercept.md
flotwig: Idk why this was marked as resolved and merged with no comment, I think this needs to be revisited still. This is a pretty clunky way to introduce RWA imo.

amirrustam: @flotwig I do agree, I have it on the list to discuss and improve in our docs improvement meeting today. I think this was just intermediary add for the use case in time for the Monday release.

Overall, we want some better context for users when we talk about the RWA.

PR #3797, opened 2021-04-07
fix: Add doc for selecting an element with id that uses CSS characters.

sainthkh:
Related with cypress-io/cypress#15020


bahmutov (approved):
nice, definitely fills a void

jprealini:
@bahmutov is there any way that Cypress could escape those characters automatically if we pass some value with "escapable" characters??

Issue #2120, opened 2019-10-01
More inclusive language in the introduction docs


carolstran:
## I'm submitting a...

```
[ ] Bug report
[x] Content update
[ ] Process update (build, deployment, ... )
```

## Type of bug / changes

:wave: I noticed that there are many instances of the word "simple" in your documentation. Unfortunately, words like this can make people feel frustrated and isolated while reading (myself included, hence the issue) - especially if they are facing issues.

If it's alright with your team, I'd like to take a shot at updating the content to make it more inclusive. Particularly the [Cypress Is Simple](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Cypress-Is-Simple) section of the Introduction, but perhaps some other places where it pops up as well. I enjoy using Cypress so I'd be happy to help 😊 

Also if you're interested or want more information on this topic, I'd highly recommend watching Jim Fisher's [Don't Say Simply](https://www.youtube.com/watch?v=gsT2BBWBVmM) talk from Write the Docs Prague 2018 ✍️ 🤓 

Thank you for the work you do and please let me know if you need more details!

sbinlondon:
Just wanted to add a +1 to @carolstran's issue. I used Cypress for the first time in an open source repo this week. While I love the documentation on the whole and think it's been one of the more thorough and well-explained bits of software I've used, I did chuckle a bit at the 'Cypress is simple'. I also ran into some bits of frustration that any person using something completely new for the first time might run into, and it didn't feel 'simple' then. Integration testing can be a hard concept. You've got a great product and great documentation - I think this change will definitely help make it even better.

flotwig:
@carolstran You're right, it may even be more accurate to update that title from "_Cypress Is Simple_" to "_Cypress Can Be Simple (Sometimes)_" :) 

We definitely don't want to frustrate anyone by implying that issues they're having are trivial. We'd be happy to merge a PR that updates the docs to make them more inclusive.

carolstran:
Awesome! I'll get to work, thank you @flotwig 🙌

carolstran:
Not sure if I should keep the issue open or close it so I'll leave it to you 😄

flotwig:
@carolstran When you open your PR, put this text in the beginning:

> Closes #2120

GitHub will link it to this issue and automatically close this when your PR is merged.

PR #3830, opened 2021-04-12
DX-75 Intercept API Doc Improvements

mccataldo:


content/api/commands/intercept.md
```diff
@@ -39,119 +39,29 @@ cy.intercept(url, routeMatcher, routeHandler)
...
+| headers    | null    | HTTP request headers                                                               |
+| hostname   | null    | HTTP request hostname                                                              |
+| https      | null    | `true`: only secure (https://) requests `false`: only insecure (http://) requests  |
+| method     | "*"     | HTTP request method                                                                |
+| middleware | false   | Pass the request on to the next `RouteMatcher` after the request handler completes |
+| path       | null    | HTTP request path after the hostname, including query parameters                   |
+| pathname   | null    | Like `path`, but without query parameters                                          |
+| port       | null    | HTTP request port(s)                                                               |
+| query      | null    | Parsed query string parameters                                                     |
+| url        | null    | Full HTTP request URL                                                              |
```
mccataldo, line 64: @amirrustam I updated this to be more concise and more consistent with the rest of the docs, but now it's really low-res compared to the typedef snippet. For example, we're not communicating where regex is another acceptable input. I thought about adding a `type` column and/or adding to the descriptions, but then it could be just as difficult to digest. Without these details, I would usually pull up the typedefs myself in vs code but I'm sure we can do better. I'm still working this out but thought I'd think aloud here.

content/api/commands/intercept.md
```diff
@@ -22,136 +22,64 @@ Unlike [cy.route()](/api/commands/route), `cy.intercept()`:
...
-cy.intercept(url, routeHandler?)
-cy.intercept(method, url, routeHandler?)
-cy.intercept(routeMatcher, routeHandler?)
+### Route-matching Only (No Stubbing or Intercepting)
+```js
+cy.intercept(url)
+cy.intercept(method, url)
+cy.intercept(routeMatcher)
+cy.intercept(url, routeMatcher)
+cy.intercept(method, url, routeMatcher)
```
mccataldo, line 33: I need to verify that using both `method` and `url` with `routeMatcher` is valid.

content/api/commands/intercept.md
```diff
@@ -22,136 +22,64 @@ Unlike [cy.route()](/api/commands/route), `cy.intercept()`:
...
+```
+
+### Route-matching With Stubbing and Intercepting
+```js
+cy.intercept(url, routeHandler)
+cy.intercept(method, url, routeHandler)
+cy.intercept(routeMatcher, routeHandler)
 cy.intercept(url, routeMatcher, routeHandler)
+cy.intercept(method, url, routeMatcher, routeHandler)
 ```
```
mccataldo, line 43: Breaking it out like this really helped me digest this a bit more easily since I consider stubbing part of a later phase of route aliasing.

content/api/commands/intercept.md
```diff
@@ -22,136 +22,64 @@ Unlike [cy.route()](/api/commands/route), `cy.intercept()`:
...
+**<Icon name="check-circle" color="green"></Icon> Correct Usage**
+
+```js
+cy.intercept('/users/**')
+```
 ### Arguments
 
 #### **<Icon name="angle-right"></Icon> url** **_(`string | RegExp`)_**
 
 Specify the URL to match. See the examples for [Matching URL](#Matching-URL) to see how URLs are matched.
```
mccataldo, line 58: Implied standards for an argument in an Arguments section:
* not in a header, but bold and camelCase
* prefixed with an angle-right icon
* types are italicized and PascalCase
* types are wrapped in parens and separated by commas

content/api/commands/intercept.md
```diff
@@ -53,32 +53,32 @@ cy.intercept('/users/**')
...
+| https      | null    | `true`: only secure (https://) requests, `false`: only insecure (http://) requests |
 | method     | "*"     | HTTP request method                                                                |
 | middleware | false   | Pass the request on to the next `RouteMatcher` after the request handler completes |
 | path       | null    | HTTP request path after the hostname, including query parameters                   |
 | pathname   | null    | Like `path`, but without query parameters                                          |
-| port       | null    | HTTP request port(s)                                                               |
-| query      | null    | Parsed query string parameters                                                     |
+| port       | null    | HTTP request port(s) (_number_ or _array_)                                         |
+| query      | null    | Parsed query string (_object_)                                                     |
 | url        | null    | Full HTTP request URL                                                              |
```
mccataldo: Here's my attempt at increasing resolution while still staying concise.

content/api/commands/intercept.md
mccataldo: I wonder what happens if we specify `method` and `url` with `routeMatcher` that contains `method` and `url`.

content/api/commands/intercept.md
```diff
@@ -22,136 +22,64 @@ Unlike [cy.route()](/api/commands/route), `cy.intercept()`:
...
-Specify the HTTP method to match on.
+Specify the URL to match. See [Matching URL](#Matching-URL) for examples.
 
-```ts
-cy.intercept('POST', 'http://example.com/widgets', {
-  statusCode: 200,
-  body: 'it worked!',
-})
-```
+**<Icon name="angle-right"></Icon> method** **_(String)_**
```
mccataldo, line 61: Implied standards for an argument in an Arguments section:
* not in a header, but bold and camelCase
* prefixed with an angle-right icon
* types are italicized and PascalCase
* types are wrapped in parens and separated by commas

content/api/commands/intercept.md
```diff
@@ -22,136 +22,64 @@ Unlike [cy.route()](/api/commands/route), `cy.intercept()`:
...
 
-#### **<Icon name="angle-right"></Icon> routeMatcher** **_(`RouteMatcher`)_**
+Specify the [HTTP request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) (`GET`, `POST`, `PUT`, etc.) to match on.
 
-`routeMatcher` is an object used to match which incoming HTTP requests will be handled by this route.
+**<Icon name="angle-right"></Icon> routeMatcher** **_(`RouteMatcher`)_**
```
mccataldo, line 64: Implied standards for an argument in an Arguments section:
* not in a header, but bold and camelCase
* prefixed with an angle-right icon
* types are italicized and PascalCase
* types are wrapped in parens and separated by commas

content/api/commands/intercept.md
```diff
@@ -22,136 +22,64 @@ Unlike [cy.route()](/api/commands/route), `cy.intercept()`:
...
-```ts
-cy.intercept('POST', 'http://example.com/widgets', {
-  statusCode: 200,
-  body: 'it worked!',
-})
-```
+**<Icon name="angle-right"></Icon> method** **_(String)_**
 
-#### **<Icon name="angle-right"></Icon> routeMatcher** **_(`RouteMatcher`)_**
+Specify the [HTTP request method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) (`GET`, `POST`, `PUT`, etc.) to match on.
```
mccataldo, line 62: Do all links like this have to be short URLs?

content/api/commands/intercept.md
```diff
@@ -22,136 +22,70 @@ Unlike [cy.route()](/api/commands/route), `cy.intercept()`:
...
 
-`routeMatcher` is an object used to match which incoming HTTP requests will be handled by this route.
+<Alert type="bolt">
 
-All properties are optional. All properties that are set must match for the route to handle a request. If a `string` is passed to any property, it will be glob-matched against the request using [`minimatch`](https://github.com/isaacs/minimatch).The available `routeMatcher` properties are listed below:
+If no method is defined Cypress will match `GET` requests by default.
```
mccataldo, line 66: I'm assuming this is true.

mccataldo: Nope! It's not. I corrected the message.

mccataldo:
I don't understand how `cy.intercept` yields `null` as stated in the docs. Otherwise, how could it be aliased?

content/api/commands/intercept.md
```diff
@@ -4,163 +4,78 @@ title: intercept
...
-- If an **object with no [`StaticResponse`][staticresponse] keys** is passed, it will be sent as a JSON response body. For example, passing `{ foo: 'bar' }` is equivalent to passing `{ body: { foo: 'bar' } }`.
-- If a **callback** is passed, it will be called whenever a request matching this route is received, with the first parameter being the request object. From inside the callback, you can modify the outgoing request, send a response, access the real response, and much more. See ["Intercepted requests"][req] for more information.
+The `routeHandler` function accepts the matched request as its argument which has methods used for modifying the outgoing request, the incoming response, or both.
+
+If you only need to modify the response you can pass in a [`StaticResponse`][staticresponse] object (to define the body, headers, status code and more). And if it's specifically the response _body_ you're wanting to stub you can simply pass that in as plain JSON instead.
```
mccataldo: To me, the term `handler` implies a function, so I decided to start with that and then introduce the other types as alternatives to that.

content/api/commands/intercept.md
```diff
@@ -170,6 +85,8 @@ The `routeHandler` defines what will happen with a request if the [routeMatcher]
 
 ## Examples
 
+Despite its name, `cy.intercept` can be used to passively listen for matching network requests and apply aliases to them without manipulating the request or its response in any way. As you'll see, this alone is powerful as it allows you to wait for these requests, resulting in more reliable tests.
```
mccataldo, line 88: I might move this up to the intro

content/api/commands/intercept.md
```diff
@@ -922,42 +839,15 @@ A `StaticResponse` represents a stubbed response to an HTTP request. You can sup
...
-}
-```
+| Option            | Default | Description                                            |
+|-------------------|---------|--------------------------------------------------------|
+| fixture           | null    | Serve a fixture as the HTTP response body              |
+| body              | null    | Serve a static string/JSON object as the response body |
+| headers           | {}      | HTTP response headers                                  |
+| statusCode        | 200     | HTTP response status code                              |
+| forceNetworkError | null    | Force an error by destroying the browser session       |
+| delay             | 0       | Minimum network latency/delay (milliseconds)           |
```
mccataldo, line 849: In the snippet, we don't specify this default, but it still makes sense.

mccataldo: There's latency in the network irrespective of this, so I wonder if we _add_ this delay to that.

content/api/commands/intercept.md
```diff
@@ -922,42 +839,15 @@ A `StaticResponse` represents a stubbed response to an HTTP request. You can sup
...
+| throttleKbps      | null    | Maximum network throughput (kilobits/second)           |
```
mccataldo, line 850: Chrome devTools has some example network profiles we might be able to reference

content/api/commands/intercept.md
```diff
@@ -4,163 +4,78 @@ title: intercept
...
-With `cy.intercept()`, you can:
+## Syntax
 
-- stub or spy on any type of HTTP request.
-  - If `cy.intercept()` provides a response object, or a fixture, or calls `req.reply()` then the request will NOT go to the server, and instead will be mocked from the test.
-  - Otherwise the request will go out to the server, and the test spies on the network call. The spying intercept can even modify the real response from the server before it is returned to the web application under test.
-- [modify an HTTP request's body, headers, and URL](#Intercepting-a-request) before it is sent to the destination server.
-- stub the response to an HTTP request, either dynamically or statically.
-- [modify real HTTP responses](#Intercepting-a-response), changing the body, headers, or HTTP status code before they are received by the browser.
-- and much more - `cy.intercept()` gives full access to all HTTP requests at all stages.
```
mccataldo, line 15: I plan to add content back to the intro as part of DX-79

content/api/commands/intercept.md
mccataldo: make them a header, remove bold?

content/api/commands/intercept.md
amirrustam: They don't. That URL is fine.

mccataldo:
I'm taking an approach such that the reader gets a bit at a time. The examples build upon themselves. I'm deliberately borrowing code from previous examples.

content/api/commands/intercept.md
```diff
@@ -170,116 +94,153 @@ The `routeHandler` defines what will happen with a request if the [routeMatcher]
...
+// https://staging.cypress.io/users
+// http://localhost/users
 
-```javascript
-// will match any HTTP method to urls that end with 3 or 5
 cy.intercept('**/users?_limit=+(3|5)')
+// matches all of these:
+// https://prod.cypress.io/users?_limit=3
+// http://localhost/users?_limit=5
+
```
mccataldo, line 140: I decided to sell glob pattern matching here. I'm likely going to move the bulk of this further down and link to it from here.

content/api/commands/intercept.md
```diff
@@ -170,116 +94,153 @@ The `routeHandler` defines what will happen with a request if the [routeMatcher]
...
-```javascript
-cy.intercept(/\/users\?_limit=(3|5)$/).as('users')
-cy.get('#load-users').click()
-cy.wait('@users').its('response.body').should('have.length', 3)
+If you don't pass a `method` argument, then all HTTP methods (`GET`, `POST`, `PUT`, `PATCH`, `DELETE`, etc.) will match.
 
-// intercepts _limit=5 requests
-cy.get('#load-five-users').click()
-cy.wait('@users').its('response.body').should('have.length', 5)
-```
```
mccataldo, line 231: I decided to omit this example since I don't think matching with regex should be encouraged. There's still a line in a previous snippet showing that regex is an option.

content/api/commands/intercept.md
```diff
@@ -170,116 +94,153 @@ The `routeHandler` defines what will happen with a request if the [routeMatcher]
...
 
 ```js
-cy.intercept({
-  url: 'http://example.com/search*',
-  query: { q: 'expected terms' },
-}).as('search')
-
-// once any type of request to http://example.com/search with a querystring containing
-// 'q=expected+terms' responds, this 'cy.wait' will resolve
-cy.wait('@search')
```
mccataldo, line 255: I'll demonstrate routematcher like this in an "advanced matching" section

content/api/commands/intercept.md
```diff
@@ -170,116 +94,153 @@ The `routeHandler` defines what will happen with a request if the [routeMatcher]
...
-})
+cy.intercept('GET', '**/users').as('getAllUsers')
+cy.intercept('POST', '**/users').as('createUser')
+
+// remember the `@` prefix when referencing aliases
+cy.get('button.createUser').click()
+cy.wait('@createUser')
+// the commands below will not execute until a response is received from `POST /users`
+cy.get('#field.firstName').should('equal', 'Adam')
+
```
mccataldo, line 214: I'm building on previous examples. Rather than use `@someRoute`, I'm continuing to use `@createUser`, which should reduce **cognitive load**.

content/api/commands/intercept.md
```diff
@@ -170,116 +94,153 @@ The `routeHandler` defines what will happen with a request if the [routeMatcher]
...
+### Using the yielded object
+You could also access the entire object which represents the request/response cycle that was intercepted:
 
-// assert that a request to this route received a response body that includes 'id'
-cy.wait('@someRoute').its('response.body').should('include', 'id')
+```js
+cy.wait('@createUser').then((ic) => {
+  const { request, response, id, ...rest } = ic
+  // ...
+})
```
mccataldo, line 238: I don't see a use case for using the entire "interception" object so I led with the more common scenario.

content/api/commands/intercept.md
```diff
@@ -170,116 +94,153 @@ The `routeHandler` defines what will happen with a request if the [routeMatcher]
...
 ```
 
-#### Aliasing individual requests
 
+#### Aliasing individual requests
+<!-- does this have other use cases outside of graphQL? -->
```
mccataldo, line 243: I might create a graphQL section further down for all this. The first example in this section uses `routeHandler` which we haven't introduced yet.

content/api/commands/intercept.md
```diff
@@ -2,165 +2,89 @@
...
-Unlike [cy.route()](/api/commands/route), `cy.intercept()`:
+Use `cy.intercept()` to manage the behavior of HTTP requests at the network layer.
 
-- can intercept all types of network requests including Fetch API, page loads, XMLHttpRequests, resource loads, etc.
-- does not require calling [cy.server()](/api/commands/server) before use - in fact, `cy.server()` does not influence `cy.intercept()` at all.
-- does not have method set to `GET` by default, but intercepts `*` methods.
+## Syntax
 
-## Usage
+### Route-matching Only (No Stubbing or Intercepting)
```
mccataldo, line 18: Technically we are intercepting so I need to update this.

content/api/commands/intercept.md
```diff
@@ -2,165 +2,89 @@
...
-- stub the response to an HTTP request, either dynamically or statically.
-- [modify real HTTP responses](#Intercepting-a-response), changing the body, headers, or HTTP status code before they are received by the browser.
-- and much more - `cy.intercept()` gives full access to all HTTP requests at all stages.
+<!-- TODO DX-79 Refactor introduction to concisely highlight the the key capabilities of the API
+The current introductory content is extraneous and can be condensed to cover:
+Modifying HTTP requests and responses, dynamically and statically.
+Ability to intercept all network request types: Fetch API, page loads, XMLHttpRequests, resource loads, etc.
+The introduction should also list some common applications of the API. For example:
+Creating error scenarios to for testing
+Slowing network traffic for testing user experience
```
mccataldo, line 11: @amirrustam I'd like to look at the req's for this if they exist

content/api/commands/intercept.md
```diff
@@ -2,165 +2,100 @@
...
-  - If `cy.intercept()` provides a response object, or a fixture, or calls `req.reply()` then the request will NOT go to the server, and instead will be mocked from the test.
-  - Otherwise the request will go out to the server, and the test spies on the network call. The spying intercept can even modify the real response from the server before it is returned to the web application under test.
-- [modify an HTTP request's body, headers, and URL](#Intercepting-a-request) before it is sent to the destination server.
-- stub the response to an HTTP request, either dynamically or statically.
-- [modify real HTTP responses](#Intercepting-a-response), changing the body, headers, or HTTP status code before they are received by the browser.
-- and much more - `cy.intercept()` gives full access to all HTTP requests at all stages.
+<Alert type="warning">
 
-## Comparison to `cy.route()`
+    `cy.route` was replaced by `cy.intercept` in Version 6.0.0. See [Comparison to cy.route](#Comparison-to-cy-route)
```
jennifer-shehane, line 18: Technically deprecated, not replaced.

content/api/commands/intercept.md
```diff
@@ -2,165 +2,100 @@
...
 
-Unlike [cy.route()](/api/commands/route), `cy.intercept()`:
+</Alert>
 
-- can intercept all types of network requests including Fetch API, page loads, XMLHttpRequests, resource loads, etc.
-- does not require calling [cy.server()](/api/commands/server) before use - in fact, `cy.server()` does not influence `cy.intercept()` at all.
-- does not have method set to `GET` by default, but intercepts `*` methods.
+## Syntax and Usage
```
jennifer-shehane, line 22: I'm not opposed to changing this if we want, but just informing that every API command doc has standardized format for this section currently if you look at all the other API docs. I've found it useful for myself that they follow a similar format.

Screen Shot 2021-05-13 at 12 27 38 PM

content/api/commands/intercept.md
```diff
@@ -2,165 +2,100 @@
...
-  /**
-   * Match against the full request URL.
-   */
-  url?: string | RegExp
-}
-```
+<!-- TODO add an example link for each option -->
 
-`routeMatcher` usage examples:
+| Option     | Description                                                                             |
```
jennifer-shehane, line 76: I feel like having a `type` column would make this a lot easier to read. Like https://docs.cypress.io/api/plugins/browser-launch-api#Syntax

content/api/commands/intercept.md
```diff
@@ -364,102 +424,218 @@ cy.intercept('GET', '/should-err', { forceNetworkError: true }).as('err')
 cy.wait('@err').should('have.property', 'error')
 ```
 
-### Stubbing a response
+### routeHandler
+
+The previous examples have dealt with just spying and aliasing routes so we can wait for requests and make assertions on them in test. The remaining examples will build upon previous ones simply by specifying a `routeHandler` as the last argument to `cy.intercept`.
```
jennifer-shehane, line 429: Avoid use of the word 'simply'. We had a textlint that used to run on this repo to catch words we want to avoid/miscapitalizations, but I'm not sure it was ever set up to run again. https://github.com/cypress-io/cypress-documentation/blob/master/.textlintrc#L115:L115

Historical context: https://github.com/cypress-io/cypress-documentation/issues/2120

```suggestion
The previous examples have dealt with just spying and aliasing routes so we can wait for requests and make assertions on them in test. The remaining examples will build upon previous ones by specifying a `routeHandler` as the last argument to `cy.intercept`.
```

content/api/commands/intercept.md
```diff
@@ -364,102 +424,218 @@ cy.intercept('GET', '/should-err', { forceNetworkError: true }).as('err')
...
-  statusCode: 404,
-  body: '404 Not Found!',
-  headers: {
-    'x-not-found': 'true',
-  },
+// stub out the response without interacting with a real back-end
+cy.intercept('POST', '/users', (req) => {
+  req.reply({
+    headers: {
+      Set-Cookie: 'newUserName=Peter Pan;'
```
jennifer-shehane, line 519:
```suggestion
      'Set-Cookie': 'newUserName=Peter Pan;'
```

content/api/commands/intercept.md
```diff
@@ -612,76 +711,25 @@ Clone the <Icon name="github"></Icon> [Real World App (RWA)](https://github.com/
...
+The object representing the intercepted request has several properties from the HTTP request itself.
 
-```js
-cy.intercept('/api', (req) => {
-  // `req` represents the intercepted HTTP request
-})
-```
-
-From here, you can do several things with the intercepted request:
+| Property    | Description                                          |
```
jennifer-shehane, line 718: `type` header again

jennifer-shehane (commented):
I was just checking up on this PR because essentially I still get a lot of one off questions on use from the current intercept docs. See [Discussions with intercept in the title](https://github.com/cypress-io/cypress/discussions?discussions_q=in%3Atitle+intercept) in cypress repo. (Wow it's a lot).

Made a few notes reading through. Some broader notes:

- I feel like there is confusion over how overriding intercepts works in general, since this is newer I guess. There's a [complicated explanation](https://docs.cypress.io/api/commands/intercept#Interception-lifecycle), but I think it'd help more to just see examples. 
- I think a lot of the current examples are like 'you can do this' and 'you can do this' but not necessarily what applications that may be useful in specifically. Maybe because they're generic 'user' examples they seem less helpful. I can't quite put my finger on it. I'd like more direct links to give people with their answer, not just a link to all the features that they're capable of using with intercept. I'm not sure this comment has ended up being helpful. 😅

content/api/commands/intercept.md
cowboy: It would be nice to standardize on these columns (and names) for object properties across all of our docs

mccataldo:
@jennifer-shehane @cowboy This is too big of an iteration to work with efficiently so I plan to break this up into smaller PRs after I finish resolving all the comments. Thanks for all your feedback so far!

mccataldo:
See the last comment.

PR #3844, opened 2021-04-15
add cy.get vs .find clear explanation

### LINKED ISSUE #3842: explain better the difference between cy.get and cy.find
Need to explain when to use cy.get vs cy.find since users are probably trying to do `cy.get(....).then( .. cy.get)` again, but the new `cy.get` starts the search again.

Screen Shot 2021-04-14 at 7 14 18 PM

bahmutov:
closes #3842
added test in https://glebbahmutov.com/cypress-examples/7.1.0/commands/querying.html#cy-get-vs-find

### DIFF
--- a/content/api/commands/get.md
+++ b/content/api/commands/get.md
@@ -135,7 +135,7 @@ cy.get('form').within(() => {
 })
 ```
 
-### Get vs .find
+### Get vs Find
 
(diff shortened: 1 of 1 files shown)


content/api/commands/get.md
```diff
@@ -135,6 +135,32 @@ cy.get('form').within(() => {
 })
 ```
 
+### Get vs .find
```
amirrustam, line 138:
```suggestion
### Get vs Find
```

amirrustam (changes requested):

PR #3880, opened 2021-04-26
Fix: gh link with hash that has quotes around it.

### LINKED ISSUE #3879: GitHub issue link is broken when including hash
## Type of bug / changes

On this section: https://docs.cypress.io/faq/questions/using-cypress-faq#How-do-I-preserve-cookies-localStorage-in-between-my-tests There's a link for 'you can read more here'. That links 404s because the url to the issue comment is being wrapped in extraneous quotes. 

Link should be: `https://github.com/cypress-io/cypress/issues/461#issuecomment-325402086`
Actual link: `https://github.com/cypress-io/cypress/issues/'461#issuecomment-325402086'`

This can be easily fixed but wondering how this happened to begin with and whether other links are affected.

jennifer-shehane:
- Close #3879 

This is easily fixed but wondering how this happened to begin with and whether other links are affected.

### DIFF
--- a/content/faq/questions/using-cypress-faq.md
+++ b/content/faq/questions/using-cypress-faq.md
@@ -559,7 +559,7 @@ Cypress.Cookies.defaults({
 })
 ```
 
-You **cannot** currently preserve localStorage across tests and can read more [here](https://github.com/cypress-io/cypress/issues/'461#issuecomment-325402086').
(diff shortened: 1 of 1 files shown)


matthamil:
Good catch. I'll see if I can do a search to find other instances of this issue. Probably occurred when porting the old hexo tags from `{% issue foo bar %}` to markdown via regex.

PR #3885, opened 2021-04-27
add link to LayerCI example

bahmutov:
for https://github.com/cypress-io/cypress/pull/16233


jennifer-shehane (changes requested):
@bahmutov Can you add what's used as the CIBuildID to this table? https://docs.cypress.io/guides/guides/parallelization#CI-Build-ID-environment-variables-by-provider Do we know that?

bahmutov:
So we don't know it yet, the example is using `--ci-build-id $LAYERCI_JOB_ID-$RETRY_INDEX` but this is up to the services team to set up after we release 7.3.0 and send LayerCI variables

https://github.com/bahmutov/cypress-example-layerci/blob/main/Layerfile#L58

jennifer-shehane (approved):

PR #3887, opened 2021-04-27
Fixed link to visual testing plugin section

hco:


jennifer-shehane (approved):
That's weird that it has this whitespace char in the link. 🤔 Thanks for the fix.

PR #3896, opened 2021-04-28
add section on types of tests

bahmutov:
because I hear this question again and again and it would be nice to clarify it once and for all

https://github.com/cypress-io/cypress/discussions/16074


content/guides/core-concepts/writing-and-organizing-tests.md
```diff
@@ -23,7 +23,7 @@ We recently gave a "Best Practices" conference talk at AssertJS (February 2018).
 
 </Alert>
 
-## Folder Structure
+## Folder structure
```
bahmutov, line 26: every other 2 word heading here is lowercase

jennifer-shehane (changes requested):
I added some sections so that it's clearer to glance through the type of tests. I feel like this would be better in the 'Why Cypress" page. The introduction page is already pretty intimidating. 

I dunno. Thoughts?

bahmutov:
Sure, I initially put it there, then moved, definitely can be in the why cypress page

jennifer-shehane (approved):

amirrustam:
We're currently not ready to present unit Node testing in such a manner and in such a top-level location within the docs. If there is a question about this type of content in the future, please reach out to me.

These changes will be reverted.

bahmutov:
the entire section? the users ask about this a lot, and no one understands they can do unit tests there as well, why revert? Clarify maybe, maybe remove the node unit testing, but let's keep the content there

JessicaSachs:
1. I'm fine keeping the CT content there. It's correct, doesn't confuse me, and drives adoption.
2. No opinion on API testing.
3. I agree with Amir on Node testing.

amirrustam:
To clarify, I'm just referring to the unit/node testing section.

PR #3903, opened 2021-05-03
chore: remove windi vue-preprocessor

harlan-zw:
See: https://github.com/windicss/windicss-webpack-plugin/issues/42

There is an additional issue minus the vue-processor, windi does not have first-class support for scss files, when you import the content in your vue style blocks it doesn't allow the @apply's to be passed .

Not sure if there's a specific reason why the page layouts are setup the way they are, but I would recommend creating new layout file and have the windi classes used within the template, this would reduce a lot of boilerplate and simplify your styling.

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -92,19 +92,6 @@ export default {
    */
   build: {
     loadingScreen: process.env.NODE_ENV === 'development',
-    extend(config) {
-      config.module.rules.push({
-        test: /\.vue$/,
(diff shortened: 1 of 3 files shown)


matthamil:
@JessicaSachs this PR may or may not be of interest to you since it's against the spike branch you made into windicss. I'm going to merge it since it seems OK. Not sure if you had any intentions of revisiting that branch or if you wanted more spike work done into windicss.

PR #3905, opened 2021-05-03
add link to the example repo

bahmutov:
since the component testing is missing even a single example link


jennifer-shehane (approved):
Yeah, I was looking for example repos on this page when I first set it up.

PR #3906, opened 2021-05-03
[TS] Use a one file model for typing custom commands

elevatebart:


elevatebart:
I could as well transform this documentation to have the merged files as the main explanation.
It is a change, but would really help our users I believe.

content/guides/tooling/typescript-support.md
```diff
@@ -95,9 +95,8 @@ If your specs files are in TypeScript, you should include the TypeScript definit
 
 Even if your project is JavaScript only, the JavaScript specs can know about the new command by referencing the file using the special triple slash `reference path` comment.
 
-```javascript
+```typescript
```
jennifer-shehane, line 98: This area appears to be addressing JS specifically, which is why the example had the triple slash references, but I do think this is likely confusing people. I would remove this paragraph "Even if your project is Javascript only....". The doc is about TypeScript after all.

jennifer-shehane (changes requested):
There's this section also in the Custom commands section:

>If your specs files are in TypeScript, you should include the TypeScript definition file, `cypress/support/index.d.ts`, with the rest of the source files.

This makes it sound like I have to do an extra step, like an 'include'. But it seems like my TS is working for my custom command without doing anything else? I would remove this line if it's no longer necessary.

elevatebart:
Thank you @jennifer-shehane,

I updated the docs to include your remarks. 
I think I should as well add a link to here from the [custom commands docs](https://docs.cypress.io/api/cypress-api/custom-commands#See-also)

elevatebart:
It is now linked !

content/api/cypress-api/custom-commands.md
```diff
@@ -783,6 +783,7 @@ You can describe the method signature for your custom command, allowing IntelliS
 
 ## See also
 
+- See how to add Typescript support in the documentation [Types for custom commands](/guides/tooling/typescript-support#Types-for-custom-commands)
```
jennifer-shehane, line 786:
```suggestion
- See how to add [TypeScript support for custom commands](/guides/tooling/typescript-support#Types-for-custom-commands)
```

jennifer-shehane (approved):

PR #3910, opened 2021-05-05
docs(cy.intercept): add `times` option to `RouteMatcher`

flotwig:
For https://github.com/cypress-io/cypress/pull/14513


jennifer-shehane (approved):
I'd like more examples, but I think there's a PR already underway to revamp intercept docs, so maybe it can be added there.

PR #3912, opened 2021-05-06
DX-189 remove Unit section

mccataldo:


amirrustam (approved):

bahmutov:
For the record, this seems like a weird section to remove. It is like hiding that we can do this type of testing, no? Not sure why we would do that, since it gives the users more arguments for using cypress. If we are thinking about the dashboard and test counts then we can mention it. But omitting the browser unit tests from the docs is like chaining one hand behind the back on purpose.

PR #3919, opened 2021-05-11
DX-214 Add Session API Docs

mccataldo:
Closes DX-214

### DIFF
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -1,0 +1,3 @@
+---
+title: session
+---
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -4,7 +4,24 @@ title: session
 
 This command is pretty much all you need.
 
-What is a session?
+#### What's a Session?
+
+A session is the state of your application that can be persisted by a browser's cookies and [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage). The "state" of your application can be
+
+- information about the current authenticated user
+  - authentication key
+  - session key
+  - permissions
+  - feature toggles
+  - location
+  - user behavior
+- settings that dictate how your application should appear or behave
+  - language
+  - currency
+  - color scheme
+  - accessibility features
+  - other UI preferences
+- form data
 
 ## Usage & Syntax
 
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -202,11 +202,6 @@ Those properties are derived from the JavaScript Browser APIs: Chrome [`Network.
 
 ### Session Usage
 
-## Dashboard
-
-A new concept!
-Just like the Command Log section, show how this is represented in the Dashboard
-
 ## Notes
 
 ## Rules
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -4,6 +4,10 @@ title: session
 
 This command is pretty much all you need.
 
+The `session` command is used to preserve browser state between tests.
+
+<!-- Technically it's more the application's state than the state of the browser. It's more like state *stored* in the browser. The tricky thing here is that we're only preserving cookies and local storage, so saying that we're saving "application state" is misleading. And kind of a bummer because it would be awesome if we did. For now we have app actions for that. Saying that this is a checkpoint or snapshot in time is also similarly ambiguous and misleading. -->
+
 #### What's a Session?
 
 A session is the state of your application that can be persisted by a browser's cookies and [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage). The "state" of your application can be
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -109,7 +109,7 @@ function signIn(username, password) {
 
 #### SessionOptions
 
-The third and last argument you can optionally pass to `session` is `SessionOptions`.
+You can optionally pass `SessionOptions` to `session`.
 
 ### `validate` (`Function`)
 
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -38,6 +38,11 @@ cy.session(
       ],
       localStorage: [
         // local storage items to exclude
+      ],
+    },
+  }
+)
+```
 
 ### Arguments
 
@@ -63,13 +68,6 @@ See [SessionOptions]()
 ### Yields
 
 `session` yields session data (`SessionData`), which is an `object` with properties `cookies` and `localStorage`. See [SessionData]().
-
-
-      ],
-    },
-  }
-)
-```
 
 ## Examples
 
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -33,13 +33,17 @@ Other examples of state persisted in the browser:
 
 ## Usage & Syntax
 
-##### Define the Session
+A session must be first be defined:
+
+##### **Define the Session**
 
 ```js
 cy.session(name, setupFn, sessionOptions)
 ```
 
-##### Use the (Previously Defined) Session
+Once defined, it can be used later:
+
+##### **Use the (Previously Defined) Session**
 
 ```js
 cy.session(name)
@@ -77,7 +81,7 @@ cy.session(
 
 #### <Icon name="angle-right"></Icon> setupFn (`Function`)
 
-`setupFn` is the function that, when executed, updates the browser's local storage, cookies, or both, with data that Cypress can preserve so you can use the session again later without having to re-authenticate again - unless it's no longer valid (see [Validate]()).
+`setupFn` is the function that, when executed, updates the browser's storage, cookies, or both, with data that Cypress can preserve so you can use the session again later without having to execute again - unless it's no longer valid (see [Validate]()).
 
 #### <Icon name="angle-right"></Icon> sessionOptions (`SessionOptions`)
 
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -38,6 +38,33 @@ cy.session(
       ],
       localStorage: [
         // local storage items to exclude
+
+### Arguments
+
+#### <Icon name="angle-right"></Icon> name (`string`)
+
+`name` is the name you would like to refer to this session as. This could be the name of the user or type of user you're defining or using - like "Adam" or "Administrator".
+
+#### <Icon name="angle-right"></Icon> setupFn (`Function`)
+
+`setupFn` is the function that, when executed, updates the browser's local storage, cookies, or both, with data that Cypress can preserve so you can use the session again later without having to re-authenticate again - unless it's no longer valid (see [Validate]()).
+
+#### <Icon name="angle-right"></Icon> sessionOptions (`SessionOptions`)
+
+`sessionOptions` contains two properties:
+
+- `validate` (`Function`)
+  - returns `true`: The session is valid and will be used.
+  - returns `false`: Cypress will execute the `setupFn` to generate a new session to use.
+- `exclude` (`object`) - contains properties `cookies` and `localStorage` for specifying items that you don't want Cypress to preserve.
+
+See [SessionOptions]()
+
+### Yields
+
+`session` yields session data (`SessionData`), which is an `object` with properties `cookies` and `localStorage`. See [SessionData]().
+
+
       ],
     },
   }
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -94,19 +94,32 @@ See [SessionOptions]()
 
 `SessionData` is the [subject]() yielded by `session`. The `object` contains two properties - `cookies` and `localStorage`
 
-### Cookies
+### Cookie Object
 
-Derived from the JavaScript Web API reference for [`cookies.Cookie`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/Cookie)
-| Key | Type | Description |
-|------------|------------------|----------------------------------------------------------------------------|
-| `name` | `string` | name of the cookie |
-| `value` | `string` | value of the cookie |
-| `path` | `string` | path of the cookie |
-| `domain` | `string` | domain the cookie belongs to |
-| `httpOnly` | `boolean` | cookie is inaccessible to client-side scripts |
-| `secure` | `boolean` | cookie scope is limited to secure channels (typically HTTPS) |
-| `expiry` | `number` | cookie expiration date (Unix timestamp). Not provided for session cookies. |
-| `sameSite` | `SameSiteStatus` | SameSite state of the cookie (`no_restriction`, `strict` or `lax`) |
+`cookies` is an array of `Cookie` objects containing the following properties:
+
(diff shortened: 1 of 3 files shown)


content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
+---
+title: session
+---
+
+Use the `session` command to preserve the browser's state between tests. Once the steps to establish a session are performed, the session can be restored at a later time without executing those steps again.
```
amirrustam, line 5:
```suggestion
Preserve and restore sessions, which consist of browser [cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies), [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), and [`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) between tests.
```

amirrustam: My previous suggestion (above) was just a pass to be more direct and concise. We still need to mention something along the lines of:

> The `session` command provides a way to define and reuse a "session" with an application under test to:
> - list out 3 top-level use cases
> - and more...

We also need an info alert box that reminds people that Cypress clears cookies and localStorage between tests. It should also have a one-sentence explainer on why we do that.

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+
+<!--
+Technically it's more the application's state or state *stored* in the browser. The tricky thing here is that we're only preserving cookies and local storage, so saying that we're saving "application state" is misleading. And kind of a bummer because it would be awesome if we did. For now we have app actions for that. Saying that this is a checkpoint or snapshot in time is also similarly ambiguous and misleading.
+-->
+
+A session is persisted by a browser's cookies, [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) and [`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).
+
+When your front-end application makes a request and the response contains a `Set-Cookie` header, the browser sets that value as a cookie (or multiple cookies).
+Alternatively, your application may store a value (from a response, user input, etc.) in browser storage (`localStorage` or `sessionStorage`).
+Either way, this persisted data is used by your application and therefore affects its behavior.
```
amirrustam, line 15: We don't need to mention this here or really at all I this page. If you think it's important to mention this, then we should put it in an info box.

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+title: session
+---
+
+Use the `session` command to preserve the browser's state between tests. Once the steps to establish a session are performed, the session can be restored at a later time without executing those steps again.
+
+<!--
+Technically it's more the application's state or state *stored* in the browser. The tricky thing here is that we're only preserving cookies and local storage, so saying that we're saving "application state" is misleading. And kind of a bummer because it would be awesome if we did. For now we have app actions for that. Saying that this is a checkpoint or snapshot in time is also similarly ambiguous and misleading.
+-->
+
+A session is persisted by a browser's cookies, [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) and [`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage).
```
amirrustam, line 11: This could be removed in favor of the above suggestions.

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+  - color scheme
+  - accessibility features
+  - other UI preferences
+- form data
+
+## Usage & Syntax
+
+A session must be first be defined:
+
+##### **Define the Session**
```
amirrustam, line 38: Do we need to bold if the line is hashed?

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+  - feature toggles
+  - location
+  - user behavior
+- settings that dictate how your application should appear or behave
+  - language
+  - currency
+  - color scheme
+  - accessibility features
+  - other UI preferences
+- form data
```
amirrustam, line 32: I think it's important to point out examples of the persisted state in this doc, but we don't need to get into explanations like on line 17. Especially at the intro part of this API doc.

With API docs, we always want to give very short and concise explanations a the top then and get to usage and examples as fast as possible.

amirrustam (changes requested):

content/api/commands/session.md
mccataldo: The style is messed up. I already created an issue for it.

cowboy (changes requested):

content/api/commands/session.md
mccataldo: Thanks. It should look better now.

content/api/commands/session.md
mccataldo: I moved this down to Notes until I can figure out how to present it.

content/api/commands/session.md
mccataldo: Yep. Done!

content/api/commands/session.md
mccataldo: I moved that to the Notes section while I figure out how to present it.

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+  cy.get('.username').type(username)
+  cy.get('.password').type(password)
+  cy.contains('.button', 'Sign In').click()
+  cy.wait('@signIn')
+})
+```
+
+#### SessionOptions
+
+The third and last argument you can optionally pass to `session` is `SessionOptions`.
```
jennifer-shehane, line 125: Just probably wouldn't hardcode the arg position in the docs, in case we change the API, add another argument so it's not the last, we may not catch this to update it later.

```suggestion
You can optionally pass `SessionOptions` to `session`.
```

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+
+```js
+cy.session(
+  'read-only',
+  () => {
+    signIn('Adam', 'Pa$$w0Rd')
+  },
+  {
+    validate: () => {
+      return cy.request('GET', '/profile').should((xhr) => {
```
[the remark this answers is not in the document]
jennifer-shehane: Yah we don't return Cypress commands. I don't think this code would work since it wouldn't yield a boolean. 

I still don't understand the point of this validate function 😅

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+#### `cookies` (`sessionExcludeFilterCookies`)
+
+```ts
+type sessionExcludeFilterCookies =
+  | {
+      domain?: string | RegExp
+      name?: string | RegExp
+    }
+  | string
+  | RegExp
```
jennifer-shehane, line 198: Would be nice to show examples of string and RegExp exclusions

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+| `name`     | `string`         | name of the cookie                                                            |
+| `value`    | `string`         | value of the cookie                                                           |
+| `path`     | `string`         | path of the cookie                                                            |
+| `domain`   | `string`         | domain the cookie belongs to                                                  |
+| `httpOnly` | `boolean`        | cookie is inaccessible to client-side scripts                                 |
+| `secure`   | `boolean`        | cookie scope is limited to secure channels (typically HTTPS)                  |
+| `expiry`   | `number`         | cookie expiration date ([Unix time](https://en.wikipedia.org/wiki/Unix_time)) |
+| `sameSite` | `SameSiteStatus` | SameSite state of the cookie (`no_restriction`, `strict` or `lax`)            |
+
+This is the same type of object yielded by [`getCookies`](/api/commands/getcookies).
```
jennifer-shehane, line 241: Just to be consistent with how we display commands in the rest of the docs.

```suggestion
This is the same type of object yielded by [`cy.getCookies()`](/api/commands/getcookies).
```

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+
+## See Also
+
+- [`Cypress.Cookies`](/api/cypress-api/cookies)
+- [`clearCookie`](/api/commands/clearcookie)
+- [`clearCookies`](/api/commands/clearcookies)
+- [`getCookie`](/api/commands/getcookie)
+- [`getCookies`](/api/commands/getcookies)
+- [`setCookie`](/api/commands/setcookie)
+- `clearLocalStorage`(/api/commands/clearlocalstorage)
```
jennifer-shehane, line 273:
```suggestion
- [`Cypress.Cookies`](/api/cypress-api/cookies)
- [`cy.clearCookie()`](/api/commands/clearcookie)
- [`cy.clearCookies()`](/api/commands/clearcookies)
- [`cy.getCookie()`](/api/commands/getcookie)
- [`cy.getCookies()`](/api/commands/getcookies)
- [`cy.setCookie()`](/api/commands/setcookie)
- [`cy.clearLocalStorage()`](/api/commands/clearlocalstorage)
```

jennifer-shehane (changes requested):

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+    },
+  }
+)
+```
+
+### Arguments
+
+#### <Icon name="angle-right"></Icon> name (`string`)
+
+`name` is the name you would like to refer to this session as. This could be the name of the user or type of user you're defining or using - like "Adam" or "Administrator".
```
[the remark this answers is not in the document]
mccataldo: First I want to test what happens when a user defines a session with a duplicate name. And what if it's unique to the test file but it exists in a global hook?

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+`setupFn` is the function that, when executed, updates the browser's storage, cookies, or both, with data that Cypress can preserve so you can use the session again later without having to execute again - unless it's no longer valid (see [Validate]()).
+
+#### <Icon name="angle-right"></Icon> sessionOptions (`SessionOptions`)
+
+`sessionOptions` contains two properties:
+
+- `validate` (`Function`)
+  - returns `true`: The session is valid and will be used.
+  - returns `false`: Cypress will execute the `setupFn` to generate a new session to use.
+- `exclude` (`object`) - contains properties `cookies` and `localStorage` for specifying items that you don't want Cypress to preserve.
```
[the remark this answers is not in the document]
mccataldo: Yes but it hasn't landed. I plan to update this once it does.

content/api/commands/session.md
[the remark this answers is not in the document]
mccataldo: So we're using cache to preserve the browser's state?

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+    validate: () => {
+      return cy.request('GET', '/profile').should((xhr) => {
+        expect(xhr.status).to.equal(200)
+      })
+    },
+  }
+)
+```
+
+Data can become invalidated over time. For example, a user may only be authenticated for a finite duration after which the user must authenticate again. This expiration or invalidation can result in flaky tests. To remedy this, Cypress invokes `validate` on the session before using it. If the session is valid (`validate` returns `true`), then it will be used as is. Otherwise, if the session is invalid (`validate` returns `false`), then Cypress will execute the setup function first.
```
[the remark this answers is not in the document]
mccataldo: I chose `data` to imply a portion of the session data but that works, too.

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+#### Name & Setup Function
+
+The `name` should represent the kind of session you're defining. For example, if you're setting up a session for a user with read-only permissions, then you might choose the name "read-only".
+The contents of `setupFn` should be the steps to generate the session you want to preserve for later use.
+
+```js
+cy.session('read-only', () => { signIn('Adam', 'Pa$$w0Rd') })
+
+function signIn(username, password) {
+  cy.intercept('POST', '/auth').as('signIn')
```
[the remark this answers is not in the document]
mccataldo: Thanks. I'll start with updating the path and go from there.

content/api/commands/session.md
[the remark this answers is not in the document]
mccataldo: It depends on how `validate` works, really. I'm assuming we have to return a truthy value to consider it valid.

content/api/commands/session.md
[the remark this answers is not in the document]
mccataldo: It's my own musings @jennifer-shehane 😄  I don't include these in the final product. In fact, I got rid of it in my local changes. I'll push those now...

content/api/commands/session.md
mccataldo: Yep, I'm working on that.

content/api/commands/session.md
mccataldo: Good point. Thanks!

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+
+`setupFn` is the function that, when executed, updates the browser's storage, cookies, or both, with data that Cypress can preserve so you can use the session again later without having to execute again - unless it's no longer valid (see [Validate]()).
+
+#### <Icon name="angle-right"></Icon> sessionOptions (`SessionOptions`)
+
+`sessionOptions` contains two properties:
+
+- `validate` (`Function`)
+  - returns `true`: The session is valid and will be used.
+  - returns `false`: Cypress will execute the `setupFn` to generate a new session to use.
```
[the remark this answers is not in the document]
mccataldo: Good feedback. I'll unconfusify it.

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+    exclude: {
+      cookies: [
+        // cookies to exclude
+      ],
+      localStorage: [
+        // local storage items to exclude
+      ],
+    },
+  }
+)
```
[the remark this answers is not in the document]
mccataldo: It's tricky to keep it realistic and simple but I'll give it a go.

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+cy.session('read-only', () => { signIn('Adam', 'Pa$$w0Rd') })
+
+function signIn(username, password) {
+  cy.intercept('POST', '/auth').as('signIn')
+  cy.visit('/sign-in')
+  cy.get('.username').type(username)
+  cy.get('.password').type(password)
+  cy.contains('.button', 'Sign In').click()
+  cy.wait('@signIn')
+})
```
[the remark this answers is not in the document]
mccataldo: I'm doing that for the purpose of brevity in a different example that references `signIn` but I'll find a better way.

content/api/commands/session.md
```diff
@@ -0,0 +1,273 @@
...
+- form data
+
+## Usage & Syntax
+
+A session must be first be defined:
+
+##### **Define the Session**
+
+```js
+cy.session(name, setupFn, sessionOptions)
```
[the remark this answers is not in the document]
mccataldo: Thanks for pointing that out. I fixed it.

cowboy:
Closing in favor of #4004

PR #3932, opened 2021-05-18
feat: add fullstory to documentation site

admah:
Closes [DX-220](https://linear.app/cypress/issue/DX-220/add-fullstory-to-docs-site).

PR adds `@fullstory/browser` so that we can see user sessions in FS. Before merging, a new env variable will need to be added `FULLSTORY_ORG_ID=ZVWG5`

### DIFF
--- a/plugins/fullstory.client.js
+++ b/plugins/fullstory.client.js
@@ -1,0 +1,3 @@
+import * as FullStory from '@fullstory/browser'
+
+FullStory.init({ orgId: process.env.FULLSTORY_ORG_ID })
--- a/plugins/fullstory.client.js
+++ b/plugins/fullstory.client.js
@@ -1,3 +1,5 @@
 import * as FullStory from '@fullstory/browser'
 
-FullStory.init({ orgId: process.env.FULLSTORY_ORG_ID })
+if (process.env.CONTEXT === 'production') {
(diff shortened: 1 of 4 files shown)


plugins/fullstory.client.js
```diff
@@ -0,0 +1,3 @@
+import * as FullStory from '@fullstory/browser'
+
+FullStory.init({ orgId: process.env.FULLSTORY_ORG_ID })
```
matthamil, line 3: LGTM, probably only need to call this init when the `process.env.CONTEXT === 'production'`, looks like the tests are failing because this `init` is expecting an `orgId`

matthamil: the `CONTEXT` env var is set by Netlify btw, which is why I recommend using a check like so:

```js
if (process.env.CONTEXT === 'production') {
  FullStory.init(// ...
}
```

admah: @matthamil great call out. I do need to add that check.

amirrustam:
@admah Are there any remaining todos for this PR?

matthamil (approved):
🚢 it

matthamil:
We just need to make sure that `FULLSTORY_ORG_ID` is set in Netlify now @amirrustam then it's good to go

PR #3949, opened 2021-05-25
fix: [DX-220] update fullstory plugin to init correctly

admah:
Updated fullstory to correctly initialize.


matthamil (approved):
LGTM ship it

robertguss (approved):

PR #3951, opened 2021-05-26
Add examples for how to upload videos for retried tests

jennifer-shehane:
I’ve had people open this issue a few times now, asking for a feature to also upload videos for retried tests:

- https://github.com/cypress-io/cypress/issues/16377
- https://github.com/cypress-io/cypress/issues/16672
- https://github.com/cypress-io/cypress/issues/16681

This is actually already possible, but I don’t think people are finding it since it’s in the [‘after:spec’ API page](https://docs.cypress.io/api/plugins/after-spec-api#Examples). Why would anyone look there?

- This adds the example to the ‘Screenshots & Videos’ page as well as ‘Test Retries’. 
- It also frames the example as ‘upload video on test retries’ instead of ‘delete videos that didn’t pass’, which is less direct on the intent. 
- **Question:** Is there a way to reuse snippets of the docs? We have this same exact code sample in the docs in 3 places now and I would prefer if it could be edited in one place.

### DIFF
--- a/content/guides/guides/screenshots-and-videos.md
+++ b/content/guides/guides/screenshots-and-videos.md
@@ -71,6 +71,42 @@ If you are an FFmpeg pro and want to see all the settings and debug messages dur
 
 </Alert>
 
+### Control which videos to keep and upload to Dashboard
+
+You may want to have more control over which videos you want to keep and upload to the Dashboard. Deleting videos after the run can save resource space on the machine as well as skip the time used to process, compress, and upload the video to the [Dashboard Service](/guides/dashboard/introduction).
+
+To only process videos in the case that a tests fail, you can set the [`videoUploadOnPasses`](/guides/references/configuration#Videos) configuration option to `false`.
+
+For more fine grained control, you can use Cypress's [`after:spec`](/api/plugins/after-spec-api) event listener that fires after each spec file is run and delete the video when certain conditions are met.
+
+#### Only upload videos for specs with failing or retried tests
+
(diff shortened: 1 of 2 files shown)


content/guides/guides/screenshots-and-videos.md
```diff
@@ -71,6 +71,42 @@ If you are an FFmpeg pro and want to see all the settings and debug messages dur
 
 </Alert>
 
+### Control which videos to keep and upload to Dashboard
+
+You may want to have more control over which videos you want to keep and upload to the Dashboard. Deleting videos after the run can save resource space on the machine as well as skip the time used to process, compress, and upload the video to the [Dashboard Service](/guides/dashboard/introduction).
+
+To only process videos in the case that a tests fail, you can set the [`videoUploadOnPasses`](/guides/references/configuration#Videos) configuration option to `false`.
```
robertguss, line 78:
```suggestion
To only process videos in the case that a test fails, you can set the [`videoUploadOnPasses`](/guides/references/configuration#Videos) configuration option to `false`.
```

robertguss:
@jennifer-shehane I am looking into how to create a re-usable markdown component. I think we can leverage this: https://www.npmjs.com/package/@nuxtjs/markdownit

I will submit a PR if I get it things working and keep you posted.

robertguss:
I wasn't able to get that markdown library working. I will see if I can find another way.

jennifer-shehane:
@robertguss Very cool. We used to have this with the old hexo framework, although it was very flawed. So I know there are a few places it could be used.

robertguss (approved):

PR #3979, opened 2021-06-14
feat: Added explanations on objects as stub response body.

### LINKED ISSUE #3976: cy.intercept doesn't use "empty" JSON object as stubbed response
### Current behavior

As per the docs for the `cy.intercept()` [routeHandler](https://docs.cypress.io/api/commands/intercept#routeHandler-lt-code-gtstring-object-Function-StaticResponselt-code-gt) argument:

> If an object with no StaticResponse keys is passed, it will be sent as a JSON response body. For example, passing { foo: 'bar' } is equivalent to passing { body: { foo: 'bar' } }.

This does not seem to be the case for an empty JS object stub.

So eg. `cy.intercept('/my-data', {})` should be equivalent to a routeHandler `{ body: {} }` instead, the log reports the stubbed response to be an empty string (which you can see in the `cy.wait` command log):

image

Same bug is in all 4 `cy.intercept` command signatures (which I assume is logical as they probably all boil down to one of them eventually).

### Desired behavior

The empty JS object should be used as the JSON response body instead of the empty string.

### Test code to reproduce

https://github.com/verheyenkoen/cypress-intercept-test

(I added a 30s requestTimeout here, because sometimes the jsonplaceholder JSON response times out for some reason so that gives you some time to manually click the button.)

### Versions

Tested in Cypress 7.0.1 and 7.1.0
Electron 89 and Chrome 89.
Node.js v12.18.3
Mac OS Big Sur 11.2.3

sainthkh:
closes #3976

With the `cy.intercept` doc update on #3944, the explanation about passing an object as stub response body was removed entirely from the documentation. I kind of revived it with examples.

If you intentionally removed it and are planning to deprecate the feature in the future, just close this PR and #3976.

### DIFF
--- a/content/api/commands/intercept.md
+++ b/content/api/commands/intercept.md
@@ -342,6 +342,18 @@ Stub a response with a JSON body:
 cy.intercept('/projects', {
   body: [{ projectId: '1' }, { projectId: '2' }],
 })
+
+// If you want to stub only body, you can directly set it as an argument:
+cy.intercept(
+  '/projects', 
+  [{ projectId: '1' }, { projectId: '2' }]
+)
+
+// But empty object doesn't mean an empty JSON body, it's an empty StaticResponse:
+cy.intercept(
+  '/projects', 
+  {} // => StaticResponse
+)
 ```
 
 Stub headers, status code, and body all at once:


flotwig (commented):
@sainthkh I think the behavior of {} should be updated, see https://github.com/cypress-io/cypress/issues/17062#issuecomment-866277254

amirrustam:
@flotwig is this PR still valid?

PR #3980, opened 2021-06-14
DX-85 improving examples

mccataldo:
Closes DX-85 DX-361

### DIFF
--- a/content/api/commands/intercept.md
+++ b/content/api/commands/intercept.md
@@ -269,7 +269,7 @@ cy.intercept({
 cy.wait('@search')
 ```
 
-```ts
+```js
 cy.intercept({
   pathname: '/search',
   query: {
@@ -279,7 +279,9 @@ cy.intercept({
 // this 'cy.wait' will only resolve once a request is made to '/search'
 // with the query paramater 'q=some+terms'
 cy.wait('@searchForTerms')
+```
 
+```js
 cy.intercept(
   {
     // this RegExp matches any URL beginning with 'http://api.example.com/' and ending with '/edit' or '/save'
--- a/content/api/commands/intercept.md
+++ b/content/api/commands/intercept.md
@@ -73,10 +73,15 @@ cy.intercept('/users*', { hostname: 'localhost' }, (req) => {
 
 ### Arguments
 
-#### **<Icon name="angle-right"></Icon> method** **_(String, Glob, RegExp)_**
+#### **<Icon name="angle-right"></Icon> method** **_(String)_**
 
-Specify the [HTTP method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) (`GET`, `POST`, `PUT`, `PATCH`, `DELETE`, etc.) to match on. This is an optional argument, but if no `method` is provided, Cypress will match _any_ HTTP method (`*`) by default.
+Match the route to a specific [HTTP method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) (`GET`, `POST`, `PUT`, etc).
 
+<Alert type="bolt">
+
+If no method is defined Cypress will match all requests by default.
+
+</Alert>
 #### **<Icon name="angle-right"></Icon> url** **_(String, Glob, RegExp)_**
 
 Specify the URL to match. See [Matching `url`](#match-url) for examples.
--- a/content/api/commands/intercept.md
+++ b/content/api/commands/intercept.md
@@ -73,7 +73,7 @@ cy.intercept('/users*', { hostname: 'localhost' }, (req) => {
 
 ### Arguments
 
-#### **<Icon name="angle-right"></Icon> method** **_(String)_**
+#### **<Icon name="angle-right"></Icon> method** **_(String, Glob, RegExp)_**
 
 Specify the [HTTP method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) (`GET`, `POST`, `PUT`, `PATCH`, `DELETE`, etc.) to match on. This is an optional argument, but if no `method` is provided, Cypress will match _any_ HTTP method (`*`) by default.
 
@@ -197,7 +197,7 @@ cy.intercept('GET', '**/users')
 ### Pattern Matching
 
 ```js
-// match updates to the `/users` endpoint
+// match updates to the `/users` endpoint using glob matching
 cy.intercept({
   method: '+(PUT|PATCH)',
   url: '**/users/*',
--- a/content/api/commands/intercept.md
+++ b/content/api/commands/intercept.md
@@ -129,6 +129,8 @@ By passing in a `StaticResponse` as the last argument, you can [statically defin
 
 See [Stubbing a response with a `StaticResponse` object](#With-a-StaticResponse-object) for an example.
 
+See also [`StaticResponse` objects](#StaticResponse-objects).
+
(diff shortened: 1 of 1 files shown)


content/api/commands/intercept.md
```diff
@@ -212,9 +214,41 @@ For more guidance around aliasing requests with GraphQL, see [Working with Graph
 
 </Alert>
 
+### Matching `url` and `method` with [RouteMatcher](#routeMatcher-RouteMatcher)
```
cowboy, line 217: These "Matching" example should immediately follow the other two "Matching" examples.

content/api/commands/intercept.md
```diff
@@ -212,9 +214,41 @@ For more guidance around aliasing requests with GraphQL, see [Working with Graph
...
+cy.intercept({ method: 'GET', url: '**/users' })
+cy.intercept('GET', '**/users')
+```
+
+### Pattern Matching
+
+```js
+// match updates to the `/users` endpoint
+cy.intercept({
+  method: '+(PUT|PATCH)',
```
cowboy, line 232: I see a mention in the the latter example `method: '/PUT|PATCH/'` that it's a regex, but I don't see an explanation here of what format `method: '+(PUT|PATCH)'` is.

Additionally, in reading Arguments -> method, I don't either of these usages explained. We'll need to document this there.

content/api/commands/intercept.md
```diff
@@ -356,13 +394,23 @@ cy.intercept('/not-found', {
 })
 ```
 
-See ["`StaticResponse` objects"][staticresponse] for more information on `StaticResponse`s.
+See also [`StaticResponse` objects][staticresponse].
+
+### Using the **`routeHandler`** function
```
cowboy, line 399: Instead of calling this section "Using the routeHandler function," maybe it should be named similarly to the preceding section, "Stubbing a response," where we focus on what the user is trying to do, instead of how they're trying to do it.

For example, instead of seeing "Using the routeHandler function" in the sidebar, it may be more useful for the user to see examples like "Asserting on a request" and "Modifying an outgoing request" there.

"Modifying an outgoing request" should probably contain sub-examples "Adding a header to an outgoing request" and "Add, modify or delete a header to all outgoing requests"

"Dynamically stubbing a response" should probably be a sub-example under "Stubbing a response"

cowboy (changes requested):

content/api/commands/intercept.md
mccataldo: Yes, they should 😄  That was my intention. Good catch!

content/api/commands/intercept.md
mccataldo: Superb. I updated the comment in the snippet and the types in the heading for `method`.

cowboy: > Superb. I updated the comment in the snippet and the types in the heading for `method`.

If you push up a commit with that change, I'll review it. No need to wait until everything is done to push.

content/api/commands/intercept.md
mccataldo: I considered exactly that when I made this change. The `Using the routeHandler function` section encapsulates the examples, but they're hidden from the side nav. The reason I decided not to do what you're suggesting is because I was already going outside the scope of this PR. `Intercepting a request` doesn't make sense because the mere fact that we're spying is "intercepting". This was a source of confusion when I read the original docs and has been a thorn in my side ever since so I decided to change it to something at least marginally better.

content/api/commands/intercept.md
cowboy: In the heading for `method` I see that you added `Glob` and `RegExp` but I see a few problems with this.

1. `Glob` is not actually a type. It's just a way to format a string. I do see [GlobPattern](https://github.com/cypress-io/cypress/blob/7192a78ef273a0d0a0811721cf69a88da54cf276/packages/net-stubbing/lib/external-types.ts#L205) but that doesn't really explain to an end user what a glob is. Instead of listing `Glob` as a type in the header, mention/explain it in the description.
2. `RegExp` is a type, but the only `method` examples I see use `String` values. Is `RegExp` actually supported? eg. `cy.intercept(/PUT/PATCH/, '/my/url')`? If it is, we may want to show an example. If not, we should mention/explain the regex-as-string format in the description.

content/api/commands/intercept.md
cowboy: It's difficult for me to find what I'm looking for by using the right-side nav. I don't feel like the example titled "Using the routeHandler function" in the nav adds any value, because a user won't what it means. However, they will know what "Asserting on a request" and "Modifying an outgoing request" mean (for example). If you consider it to be out of the scope of this PR, can you make sure there's a follow-up issue to optimize the Examples nav items?

content/api/commands/intercept.md
mccataldo: I'm being consistent with the rest of the docs. If you want to change it everywhere, create an issue.

content/api/commands/intercept.md
mccataldo: Yep

content/api/commands/intercept.md
mccataldo: These are great thoughts but let's get this in and address it in a future issue.

cowboy: I'm trying to understand which docs this is consistent with. The [`cy.route` Arguments](https://docs.cypress.io/api/commands/route#Arguments) `match` only mentions `String` as a type (although `url` mentions `Glob` and `RegExp`, which makes sense to me).

For consistency's sake, how about we change `method` to exactly match what `cy.route` has, with the one difference being, in the callout, saying "all" instead of "GET"?

mccataldo: It would be really helpful if you made code suggestions.

cowboy: Sure, let me give that a try! I'm still not used to that feature.

content/api/commands/intercept.md
```diff
@@ -73,7 +73,7 @@ cy.intercept('/users*', { hostname: 'localhost' }, (req) => {
 
 ### Arguments
 
-#### **<Icon name="angle-right"></Icon> method** **_(String)_**
+#### **<Icon name="angle-right"></Icon> method** **_(String, Glob, RegExp)_**
 
 Specify the [HTTP method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) (`GET`, `POST`, `PUT`, `PATCH`, `DELETE`, etc.) to match on. This is an optional argument, but if no `method` is provided, Cypress will match _any_ HTTP method (`*`) by default.
```
cowboy, line 79:
```suggestion
#### **<Icon name="angle-right"></Icon> method** **_(String)_**

Match the route to a specific [HTTP method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) (`GET`, `POST`, `PUT`, etc).

<Alert type="bolt">

If no method is defined Cypress will match all requests by default.

</Alert>
```

cowboy (changes requested):

content/api/commands/intercept.md
cowboy: Once you make a follow-up issue, please link it here and I will resolve this comment. Thanks!

mccataldo: https://linear.app/cypress/issue/DX-399/optimize-examples

cowboy (approved):

PR #3992, opened 2021-06-23
DX-338 Update `as` `get` and `wait`

mccataldo:
Closes DX-338

### DIFF
--- a/content/api/commands/as.md
+++ b/content/api/commands/as.md
@@ -58,13 +58,20 @@ it('disables on click', () => {
 
 ### Route
 
-Aliasing a route and then using [`cy.wait()`](/api/commands/wait) to wait for the aliased route.
+Aliasing an intercepted route defined with [`cy.intercept()`](/api/commands/intercept) and then using [`cy.wait()`](/api/commands/wait) to wait for the aliased route.
 
 ```javascript
-cy.intercept('PUT', '/users', { fixture: 'user' }).as('putUser')
+// `PUT` requests on the `/users` endpoint will be stubbed with
+// the `user` fixture and be aliased as `editUser`
+cy.intercept('PUT', '/users', { fixture: 'user' }).as('editUser')
+// we'll assume submitting `form` triggers a matching request
 cy.get('form').submit()
-cy.wait('@putUser').its('url').should('contain', 'users')
+// once a response comes back from the `editUser`
+// this `wait` will resolve with the subject containing `url`
+cy.wait('@editUser').its('url').should('contain', 'users')
 ```
+
+More examples of aliasing routes can be found [here](/api/commands/intercept#Aliasing-an-intercepted-route).
 
 ### Fixture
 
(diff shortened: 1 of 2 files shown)


content/api/commands/as.md
```diff
@@ -58,14 +58,21 @@ it('disables on click', () => {
...
+// `PUT` requests on the `/users` endpoint will be stubbed with
+// the `user` fixture and be aliased as `editUser`
+cy.intercept('PUT', '/users', { fixture: 'user' }).as('editUser')
+// we'll assume submitting `form` triggers a matching request
 cy.get('form').submit()
-cy.wait('@putUser').its('url').should('contain', 'users')
+// once a response comes back from the `editUser`
+// this `wait` will resolve with the subject containing `url`
+cy.wait('@editUser').its('url').should('contain', 'users')
 ```
```
cowboy, line 72: Would it be possible to put linebreaks before the comments on line 67 and 71? Interestingly enough, GH won't let me create a suggestion here.

content/api/commands/as.md
```diff
@@ -58,14 +58,21 @@ it('disables on click', () => {
 
 ### Route
```
cowboy, line 59: Should we change the header from `Route` to `Intercept`?

cowboy (changes requested):

content/api/commands/as.md
mccataldo: OK I added whitespace

content/api/commands/as.md
mccataldo: It depends on their definitions which are currently unclear.

mccataldo:
@cowboy FYI, this is still a WIP.

mccataldo:
@cowboy We don't appear to be documenting using `get` with route or intercept aliases so I'm not sure there's a use case. Do you want to explore that as part of this issue or a separate one?

cowboy:
> @cowboy We don't appear to be documenting using `get` with route or intercept aliases so I'm not sure there's a use case. Do you want to explore that as part of this issue or a separate one?

Are you talking about my "Should we change the header from Route to Intercept?" comment? If so, it was just a question, I'm not sure any changes are necessary!

content/api/commands/wait.md
```diff
@@ -122,17 +123,18 @@ cy.wait('@getBooks')
 // responded with an empty array first
 cy.get('#book-results').should('be.empty')
 
-// now re-define the /books response
-cy.intercept('/books', [{ name: 'Emperor of all maladies' }])
+// now the request (aliased again as `getBooks`) will return one book
```
cowboy, line 126: This comment and code line are causing horizontal scrolling at large viewport sizes. Can you tweak the text and book name just a little to help them fit?

mccataldo:
> > @cowboy We don't appear to be documenting using `get` with route or intercept aliases so I'm not sure there's a use case. Do you want to explore that as part of this issue or a separate one?
> 
> Are you talking about my "Should we change the header from Route to Intercept?" comment? If so, it was just a question, I'm not sure any changes are necessary!

I've already made the changes anyway. I'm talking about updating `cy.get` with an intercept example. We never demonstrated this with `cy.route` so I'm not sure there's a use case, but we could explore outside of this PR.

cowboy (changes requested):

content/api/commands/wait.md
mccataldo: Done

cowboy (approved):

PR #3993, opened 2021-06-23
feat: add copy button to code blocks

admah:
copy-button

This PR adds a copy button to code blocks within API and Guides pages. The button will not appear within API blocks that demonstrate incorrect usage.

### DIFF
--- a/components/AppCopyButton.vue
+++ b/components/AppCopyButton.vue
@@ -2,7 +2,7 @@
   <button
     ref="copy"
     :class="$style.copy"
-    class="copy rounded-t-md rounded-b-md text-sm text-white font-normal relative flex flex-row"
+    class="copy absolute rounded-md text-xs text-white font-normal flex flex-row self-end items-center"
     data-test="copy-button"
   >
     <Icon :name="state === 'copied' ? 'check-circle' : 'paste'" color="white" />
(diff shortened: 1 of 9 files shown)


jennifer-shehane:
Had to comment that I love this. 😍 Extra props for not showing it on broken examples.

The copy button is a little weird when the scrollbar is present. I know y'all were discussing the scrolling with long code in another PR.

admah:
@jennifer-shehane thank you! I agree that it's not great UX when there's a scrollbar. I'll take another pass at it. 

Maybe I can just position the button top-right absolute in that box and see what it looks like.

admah:
https://user-images.githubusercontent.com/2184114/123290010-8b4b7480-d4d6-11eb-82d8-10772706600c.mov

Updated the way the code element flexes when there is a copy button. Blocks without copy buttons remain as-is - full-width in the parent container.

jennifer-shehane:
I don't think this is ideal, to reduce the width of all code blocks available. It's also not obviously that there may be more content hidden with the content hidding 85% of the width in. Maybe the copy button could sit outside of the code content box? Like, static positioned in the black margin somewhere? Just my observations. Up to y'all.

admah:
@jennifer-shehane thanks so much for the feedback! Here is an iteration where the button sits on top and does not limit the width of the code block below:

jennifer-shehane:
Yah, still a little weird with the long examples, but this is an improvement overall I think. 👍  Can iterate on more improvements for the scrolling experience if needed.

admah:
Closing in favor of #4015

PR #3996, opened 2021-06-24
DX-190 API: Intercept - Update section, 'Request/Response Modification with `routeHandler`' with links to examples

mccataldo:
Closes DX-190

### DIFF
--- a/content/api/commands/intercept.md
+++ b/content/api/commands/intercept.md
@@ -617,10 +617,10 @@ From here, you can do several things with the intercepted request:
 
 <!-- TODO DX-190 add links to examples -->
 
-- modify and make assertions on the request like its body, headers, URL, method, etc. ([example](/api/commands/intercept#Asserting-on-a-request-1))
-- stub out the response without interacting with a real back-end ([example](/api/commands/intercept#Stubbing-a-response))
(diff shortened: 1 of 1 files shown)


content/api/commands/intercept.md
```diff
@@ -617,10 +617,10 @@ From here, you can do several things with the intercepted request:
 
 <!-- TODO DX-190 add links to examples -->
 
-- modify and make assertions on the request like its body, headers, URL, method, etc.
-- stub out the response without interacting with a real back-end
-- pass the request through to its destination and modify or make assertions on the real response on its way back
-- attach listeners to various events on the request
+- modify and make assertions on the request like its body, headers, URL, method, etc. ([example](/api/commands/intercept#Asserting-on-a-request-1))
```
cowboy, line 620: There's a fair amount of jumping around in the examples with these links. Are we concerned about the user losing their place? Does it make sense to move any of these underneath the "Request/Response Modification with routeHandler" section?

mccataldo: OK the links refer to content below but I can't link to specific examples because of an issue with L5 headings.

cowboy: The only solution I can think of right now would be to reorganize the examples a bit, but I don't want to hold up this work for that. We can revisit later if necessary.

cowboy (approved):

PR #4004, opened 2021-06-29
DX-439 write session api docs

cowboy:
### Do not merge until https://github.com/cypress-io/cypress/pull/14350 is merged.

This PR is dependent on the sessions work being merged first.

- https://deploy-preview-4004--cypress-docs.netlify.app/api/commands/session
- https://deploy-preview-4004--cypress-docs.netlify.app/api/cypress-api/session
- https://deploy-preview-4004--cypress-docs.netlify.app/guides/references/experiments

TODO:
- [x] Update experiments history entry, blog post to reference 8.2.0. Update blog post link to reference 2021/08/02 date.


content/api/commands/session.md
```diff
@@ -0,0 +1,451 @@
...
+---
+
+Cache and restore [cookies](/api/cypress-api/cookies),
+[`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage),
+and
+[`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)
+in order to reduce test setup times.
+
+<Alert type="warning">
+<p>The <code>cy.session</code> API is currently experimental, and can be enabled by
```
mccataldo, line 12: If you want to stay consistent, change `cy.session` to `cy.session()`

content/api/commands/session.md
```diff
@@ -0,0 +1,451 @@
...
+<p>Enabling this flag does two things:</p>
+<ul>
+  <li>It enables the <code>cy.session</code> API for use in tests.
+  <li>It adds a new behavior (that will be the default in a future major update
+  of Cypress), where the page is cleared (by setting it to <code>about:blank</code>)
+  and all stored session data (cookies, <code>localStorage</code> and
+  <code>sessionStorage</code>) are cleared at the beginning of each test.</li>
+</ul>
+<p>You must explicitly call call <code>cy.visit</code> in each test to visit a page
+in your application.</p>
```
mccataldo, line 25: You shouldn't need to use HTML.

cowboy: I did some digging, and apparently as long as you keep an empty line between HTML/MDX tags and the content, it will format the content as markdown. I'll fix this in the next commit.

content/api/commands/session.md
cowboy: Fixed in d474bf0

content/api/commands/session.md
cowboy: Fixed in d474bf0

content/api/commands/session.md
```diff
@@ -0,0 +1,729 @@
...
+Cache and restore [cookies](/api/cypress-api/cookies),
+[`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage),
+and
+[`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)
+in order to reduce test setup times.
+
+<Alert type="warning">
+
+The `cy.session()` API is currently experimental, and can be enabled by setting
+the `experimentalSessionSupport` flag to `true` in the Cypress config or by
```
amirrustam, line 14: Two points here:
1. We also need to add the `experimentalSessionSupport` to the experiments table in https://deploy-preview-4004--cypress-docs.netlify.app/guides/references/experiments#Configuration
2. Within that "description" column of the table we should link back to the cy.session doc.

robertguss (approved):

content/api/commands/session.md
```diff
@@ -0,0 +1,729 @@
...
+function hasn't changed.
+
+### Explicitly clearing sessions
+
+When running the Test Runner in "open" mode, you can explicitly clear all
+sessions and re-run the spec file by clicking the "Clear All Sessions" button in
+the [Instrument Panel](#The-Instrument-Panel).
+
+For debugging purposes, all sessions can be cleared with the
+`Cypress.session.clearAllSavedSessions()` method.
```
amirrustam, line 647: This is ancillary, but `Cypress.session` also needs to be documented under the "Cypress API" section. Similar to how `Cypress.spec` is documented for example: https://deploy-preview-4004--cypress-docs.netlify.app/api/cypress-api/spec.

The `Cypress.session` doc page should also indicate this feature is still experimental and link to `cy.session` doc as well.

content/api/commands/session.md
```diff
@@ -0,0 +1,729 @@
...
+always create a new session with a different `id`.
+
+In order to reduce development time, when running the Test Runner in "open"
+mode, sessions will be cached _between spec file runs_ as long as the `setup`
+function hasn't changed.
+
+### Explicitly clearing sessions
+
+When running the Test Runner in "open" mode, you can explicitly clear all
+sessions and re-run the spec file by clicking the "Clear All Sessions" button in
```
jennifer-shehane, line 643: Would be great to have an image of this as I'd have to search for where this  is just reading it from nothing.

amirrustam (changes requested):
Exemplary! The thorough examples and additional notes section are great. 
I made some small suggestions and pointed out a few additional things that need documenting, but overall the core content is ready for the experimental release.

jennifer-shehane (changes requested):
I opened a PR with direct suggestions here: https://github.com/cypress-io/cypress-documentation/pull/4031 But I have some other feedback that I did not address in code edits below.

1. I feel like I'm missing the  overall 'why' of using cy.session. If you read the document from start to finish, it never seems to really answer this question fully. I start to get an idea on why I might use this when the examples start. It'd be great to just be really upfront after the first sentence of the page on why you want to use this. - This is great for testing logins, it will save you time, you won't have to manually juggle cookies, session data, etc... Maybe there's a blog post to point to, not  sure.
2. I'm a bit confused  about the Object and Array usage for ids. Can you provide a clear example  or passing it and how  it's used. It says it's used to 'generate a unique id', is this just implementation details that Cypress does that  the user doesn't necessarily need to know? Or can I later use this uniquely generated id?

content/api/commands/session.md
jennifer-shehane: My PR has this addition

cowboy: Done in 4d36c81ef92abd80e606fe20daf9cb2083c47636

content/api/commands/session.md
cowboy: > * I feel like I'm missing the  overall 'why' of using cy.session. If you read the document from start to finish, it never seems to really answer this question fully. I start to get an idea on why I might use this when the examples start. It'd be great to just be really upfront after the first sentence of the page on why you want to use this. - This is great for testing logins, it will save you time, you won't have to manually juggle cookies, session data, etc... Maybe there's a blog post to point to, not  sure.

Does this (draft) blog post help explain it? https://cypress-io.ghost.io/blog/p/32959945-4e57-40d7-8a47-8e1ba67784cf/ I didn't want to get too much into the why in the API docs page, because I was concerned that it would make using the page as a reference more difficult, but instead focused on the how.

content/api/commands/session.md
```diff
@@ -0,0 +1,731 @@
+---
+title: session
+---
+
+Cache and restore [cookies](/api/cypress-api/cookies),
+[`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage),
+and
+[`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)
+in order to reduce test setup times.
```
amirrustam, line 9: One point of improvement would be to think about how to better talk about the value of the API beyond stating "test setup times." I anticipate that this doesn't really convey the immediate value and could be misinterpreted. 

I have no actionable edit at the moment, but it's important to think more about this.

content/api/commands/session.md
cowboy: > Would be great to have an image of this as I'd have to search for where this is just reading it from nothing.

I have the image in the [Instrument Panel](https://deploy-preview-4004--cypress-docs.netlify.app/api/commands/session#The-Instrument-Panel) section of the docs, which is linked to in the section you've called out. Should I duplicate the image there as well?

content/api/commands/session.md
cowboy: If there's anything from the [(draft) blog post](https://cypress-io.ghost.io/blog/p/32959945-4e57-40d7-8a47-8e1ba67784cf/) that you think will add value here, let me known. I just want to make sure that we don't make this page more difficult to consume as API documentation!

cowboy:
> 2\. I'm a bit confused  about the Object and Array usage for ids. Can you provide a clear example  or passing it and how  it's used. It says it's used to 'generate a unique id', is this just implementation details that Cypress does that  the user doesn't necessarily need to know? Or can I later use this uniquely generated id?

@jennifer-shehane Imagine someone had `name` and `password` vars that were used inside the session setup function, and they wanted to cache them, so they had to create an `id`. They might do something like this:

```js
const login = (name, password) => {
  cy.session(`${name}-${password}`, () => {
    // code that uses both name and password
  })
}
```

But let's say they had an `options` object with many properties, because they were doing something more complex inside the setup function. How would they generate that `id`?

This doesn't work because `options` is stringified to `[object Object]`:

```js
`${name}-${password}-${options}`
```

This is smarter, but isn't reliable for a cache key:

```js
JSON.stringify([name, password, options])
```

Because objects aren't necessarily serialized deterministically:

```js
obj = {a: 1}; obj.b = 2; JSON.stringify(obj) // {"a":1,"b":2}
obj = {b: 2}; obj.a = 1; JSON.stringify(obj) // {"b":2,"a":1}
```

So, instead of having to worry about any of that, `cy.session()` does deterministic stringification with the [json-stable-stringify](https://www.npmjs.com/package/json-stable-stringify) module. You just pass in any stringifiable value, and an `id` will be generated for you.

If you have thoughts on how I can make that more clear, please let me know! Perhaps I could add something about this into Notes or into the [Choosing the correct id to cache a session
](https://deploy-preview-4004--cypress-docs.netlify.app/api/commands/session#Choosing-the-correct-id-to-cache-a-session) example?

jennifer-shehane:
@cowboy Ok, that explains the implementation of how this id is generated. That answers my initial question. Reading more, now I am confused over this statement below.

>Any variable that is used inside the setup function that could possibly change throughout the spec file should be specified in the id argument. This includes objects that may be mutated.

I'm not understanding why I need to be concerned about this ID I guess is what it boils down to. Can't I just make up a string name to have this always be unique? Why does what I pass in as the id have any concern with whether the variable  will change latter in the spec file? Does the id have some significance to how my session is captured? Or maybe it's strange to call this an id when it should be viewed as params that are passed into the session?

I have hints of how this works from my memory of earlier discussions, but I don't see this explained anywhere in the document.

content/api/commands/session.md
```diff
@@ -0,0 +1,731 @@
...
+```
+
+**<Icon name="exclamation-triangle" color="red"></Icon> Incorrect Usage**
+
+```javascript
+// visiting before calling cy.session() is redundant, it needs to
+// be done inside the setup function
+cy.visit('/login')
+cy.session(name, () => {
+  // need to call cy.visit() here because the page is blank here
```
jaffrepaul, line 77: Verrry subtle, but probably one too many `here`s in this line. 
Perhaps subbing in '...page is blank at this point' or 'page is blank right now' or '...page hasn't loaded yet' or something similar?

jaffrepaul (dismissed):
I added one very small comment to clarify a code comment. The only question I have is around the search. Should this populate as a result after the Algoloia scrape once it's live?

cowboy:
> I'm not understanding why I need to be concerned about this ID I guess is what it boils down to. Can't I just make up a string name to have this always be unique? Why does what I pass in as the id have any concern with whether the variable will change latter in the spec file? Does the id have some significance to how my session is captured? Or maybe it's strange to call this an id when it should be viewed as params that are passed into the session?

If I have custom `login` code that uses multiple parameters (in this example, a name, a token, and a password), in order to be able to log in many different users, but the `id` only included one of them (in this example, `name`):

```js
const login = (name, token, password) => {
  cy.session(name, () => {
    cy.visit('/login')
    cy.get('[data-test=name]').type(name)
    cy.get('[data-test=token]').type(token)
    cy.get('[data-test=password]').type(password)
    cy.get('#submit').click()
  })
}
```

If I ran this, `user1` would be logged in with `token1` and `p4ssw0rd`, and a session would be created and cached using `"user1"` as the `id`.

```js
login('user1', 'token1', 'p4ssw0rd')
```

Now let's say I wanted to try to log in the same user, but with a different token and/or password, and expect a different session to be created and cached. I run this, but because `cy.session()` is only being passed `name` as its `id`, it won't create a new session, but will instead load the saved session for `"user1"`.

```js
login('user1', 'different-token', 'p4ssw0rd')
```

In order sessions to be cached uniquely, I'd need to create an `id` from all the parameters that are used inside the `setup` function, otherwise `id` values may collide and create unexpected results.

admah:
> I added one very small comment to clarify a code comment. The only question I have is around the search. Should this populate as a result after the Algoloia scrape once it's live?

@jaffrepaul yes, the search will be updated after this is merged. It only scrapes/reindexes when things are merged into master.

content/api/commands/session.md
cowboy: Fixed in 6805e78

content/api/commands/session.md
amirrustam: I'll take another look at the post. Sidenote, we should also link to he blog post in the "See Also" section.

cowboy:
> > I'm not understanding why I need to be concerned about this ID I guess is what it boils down to. Can't I just make up a string name to have this always be unique? Why does what I pass in as the id have any concern with whether the variable will change latter in the spec file? Does the id have some significance to how my session is captured? Or maybe it's strange to call this an id when it should be viewed as params that are passed into the session?

Explaining this has reminded me that I wanted to create an example that was missing. I know it's a bit of a roundabout answer, but does 23fa71e79337890ff75425cc510d7695690e52f9 help to explain this at all?

cowboy:
> > I'm not understanding why I need to be concerned about this ID I guess is what it boils down to. Can't I just make up a string name to have this always be unique? Why does what I pass in as the id have any concern with whether the variable will change latter in the spec file? Does the id have some significance to how my session is captured? Or maybe it's strange to call this an id when it should be viewed as params that are passed into the session?

Explaining this has reminded me that I wanted to create an example that was missing. I know it's a bit of a roundabout answer, but does 23fa71e79337890ff75425cc510d7695690e52f9 help to explain this at all?

cowboy:
@jennifer-shehane ok, I think I've got it. See 8f6895f!

content/api/commands/session.md
cowboy: Link added in ecefd272a47a77cdb1391356c3f9ee3d344ec6b9

content/api/commands/session.md
cowboy: Done in 8e499112dc0bbc4bfdd66014fc96eb3cf48e2863

content/api/commands/session.md
jennifer-shehane: Yah, probably just duplicate the image. Since I was reading straight through or if you're linked here directly, you may not know to scroll down. I must have missed the link.

jennifer-shehane:
@cowboy I think I fully understand now and I think that the explanation around the id generation may be more complicated than is necessary. Here's a suggestion for that intro paragraph on the unique ID paragraph:

---

The `id` for `cy.session()` requires being unique for each new session created. In order to make this id generation simpler, Cypress generates an id for you by stringifying any parameters passed to the `id` argument as an Object or Array. For example, if you pass `['Jane', '123', 'admin']`, we would generate an id of `Jane-123-admin`. You'll want to make sure the id that'll be generated for each new session is unique.

cowboy:
> The `id` for `cy.session()` requires being unique for each new session created. In order to make this id generation simpler, Cypress generates an id for you by stringifying any parameters passed to the `id` argument as an Object or Array. For example, if you pass `['Jane', '123', 'admin']`, we would generate an id of `Jane-123-admin`. You'll want to make sure the `id` that'll be generated for each new session is unique.

I've addressed this in fb1161f46bbce5e921860ed6c6b70ba991c02576, thanks!

content/api/commands/session.md
cowboy: Image added in fb1161f46bbce5e921860ed6c6b70ba991c02576.

cowboy:
@jaffrepaul @jennifer-shehane @amirrustam I believe I have addressed all outstanding issues! Please resolve any open conversations and approve the PR if you feel it's good, thanks!

content/api/commands/session.md
```diff
@@ -0,0 +1,847 @@
...
+Clicking a session `id` in the Instrument Panel or clicking the first line under
+an expanded session group in the command log will print that session's details
+to the console. This information contains the `id` along with any cached session
+data, including cookies, `localStorage` and `sessionStorage`.
+
+<DocsImage src="/img/api/session/print-session-to-console.png" alt="Session console output"></DocsImage>
+
+## See also
+
+- [Authenticate faster in tests with the cy.session command](https://cypress-io.ghost.io/blog/authenticate-faster-in-tests-cy-session-command/)
```
amirrustam, line 844:
```suggestion
- [Authenticate faster in tests with the cy.session command](https://cypress.io/blog/2021/06/19/authenticate-faster-in-tests-cy-session-command/)
```

amirrustam (changes requested):

jennifer-shehane (approved):

amirrustam (approved):
LGTM. @cowboy let's fix the failing CI job before merging.

jennifer-shehane:
### Do not merge until https://github.com/cypress-io/cypress/pull/14350 is merged.

This PR is dependent on the sessions work being merged first. If it's merged in today for 8.0, we need to update the base branch of this PR to 8.0-release before merging. 

If it is not merged today, this will go into the next release.

cowboy:
TODO: update blog post link (and blog post) before merging.

jennifer-shehane:
Changing base to 8.1.0 release branch. 🤞

jennifer-shehane:
Changed the base branch to 8.2.0 release

PR #4010, opened 2021-07-05
fix: correct regression in requirement wording

JessefSpecialisterren:
A few months ago, commit 9bcdd2574c replaced "when" with "or" in a requirement of the `request` command, making it incorrect. This commit changes the wording back to the old one. Old (correct) wording (notice the "when" after "3xx"):

https://github.com/cypress-io/cypress-documentation/blob/fa40ceb26c466be7a5aa8007921219032da0f78b/lib/tags/requirements.js#L137

Incorrect wording introduced in 9bcdd2574c (notice the "or" after "3xx":

https://github.com/cypress-io/cypress-documentation/blob/9bcdd2574cd472203f854bb52cbf05618bc59962/scripts/requirements.js#L80


jennifer-shehane (approved):
Thanks!

PR #4022, opened 2021-07-13
Always wrap markdown prose

cowboy:
By setting [`proseWrap`](https://prettier.io/docs/en/options.html#prose-wrap) to `always`, saving a file (with prettier formatting enabled) now results in a much easier to edit file if you use vim, because now you can actually use line-based commands!

Eg.

## Before

image

## After

image


admah (approved):
🎉

PR #4023, opened 2021-07-14
fix: replaced missing font awesome icons

robertguss:
While attempting to build and serve a static version of the docs site, several errors appear in the console regarding various font-awesome icons that cannot be found. It looks like some of these icons are referencing an older version of FA. I have replaced them with icons from FA v5 which is what we are currently using.


admah (approved):
Looks good!

PR #4024, opened 2021-07-14
Fix accessibility issues flagged by axe

jacobherrington:
I took a quick peak at the documentation site with an automated
accessibility tool and it flagged a few issues.

This commit fixes two of those:
1) An aside shouldn't use the aria-role navigation
2) An ul shouldn't contain anything except li elements


robertguss:
Hi @jacobherrington thank you so much for this. Do you mind sharing a link to the tool you used to find these. It would be helpful for us going forward.

jacobherrington:
@rob-balfre Yup! :smile: It was the Firefox version, but the same tool!

PR #4027, opened 2021-07-15
fixed hydration errors

robertguss:
There were 3 `hydration` errors that were being thrown because a `<svg>` element was nested inside of a `<button>` element for each link in the sidebar. I managed to fix these errors by moving the `<svg>` element outside of the `<button>` and adding some additional tailwind classes to fix the styling. 

The elements which were causing this issue are in the sidebar. The arrow icon is the `<svg>` icon that was the issue. 
Screen Shot 2021-07-15 at 1 56 36 PM

You will not see the hydration errors in production, most likely. However, locally vue will output warnings to the console like so:
Screen Shot 2021-07-15 at 1 58 46 PM

If you pull down this branch locally, these errors are now gone. I share all of this as you most likely will not see these errors and issues on Prod or in the Netlify preview.

### DIFF
--- a/components/sidebar/CollapsibleSidebarSection.vue
+++ b/components/sidebar/CollapsibleSidebarSection.vue
@@ -104,12 +104,11 @@ export default {
 
 <template>
   <div
-    :class="
-      depth > 1
-        ? `ml-${depth + 3} pl-2 border-l-2 border-gray-200`
-        : 'px-4 pb-4'
-    "
-    class="space-y-1"
+    :class="[
+      'space-y-1', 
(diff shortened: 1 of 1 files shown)


robertguss:
This is the blog article that helped me figure out how to debug the hydration issues: https://blog.lichter.io/posts/vue-hydration-error/#solving-the-hydration-failure

kevinold:
@robertguss Use https://image.nuxtjs.org/components/nuxt-img/ to call the svg and see if it resolves the issue.

PR #4029, opened 2021-07-15
Fix various nuxt issues

robertguss:
- Fixed link in `AppHeader.vue` to the proper plugins directory url
- Fixed bad links `/api/api/table-of-contents` in markdown files. This URL does not exist.
- Removed explicit `trailingSlash: undefined` since this is the nuxt default, which makes it redundant
- Fixed issues with 404 page creation and updated npm build script accordingly
- Added in three default nuxt npm scripts per their documentation. This is primarily for testing and local dev.
- Removed redirect `/guides/dashboard/dashboard-introduction` in `redirects.js` since this was causing nuxt to try to generate a route that does not exist.
- added analyze npm script for webpack performance and profiling (https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-build#analyze)


robertguss:
We are going to wait until Nuxt 3 is released to see if it will fix most of these issues for us.

PR #4030, opened 2021-07-15
docs: Update client certificates docs

jennifer-shehane:
- Fix links that did not go to client-certificates doc
- Add History to client certs and config pages to specify when this feature is added
- Fix reference of where this should live (should be cypress.json, not cypress.env.json)
- Rearrange Syntax

I’m on the fence about my updates to the Syntax section. I like tables for the properties, but maybe this isn’t clearer than  the  list before? Thoughts?

### Before

Screen Shot 2021-07-15 at 1 28 27 PM

### After

localhost_3000_guides_references_client-certificates (1)


cowboy (approved):
This is WAY better than what I had! 💯  🎉  🥳

cowboy:
I'm so impressed by your use of tables that I'm going to start using them instead of CSS for [all my site layouts](https://youtu.be/dQw4w9WgXcQ)!

jennifer-shehane:
Do not bring my memories back to those days. 😑

PR #4031, opened 2021-07-16
Edits for session suggestions

jennifer-shehane:
- Add sessions to experiments page
- emove language that makes writing the code sound easy - this minimizes people's experience who may be struggling writing the new API we wrote. https://github.com/cypress-io/cypress-documentation/issues/2120 We used to use a markdown linter to remove  works like 'just'  and 'simple',but it was removed in the migration  to the new framework. I'd love to see this added back as this shouldn't rely on manual reviews.


admah:
@jennifer-shehane thank you for surfacing the second point. I wasn't aware that we had something like that in place, but I do know we have talked about many improvements that we can make to how we handle markdown in the new docs and this will be added to that discussion.

cowboy:
This is great stuff, thanks!

jennifer-shehane:
@admah This file actually still exists, but I don't think it's being run https://github.com/cypress-io/cypress-documentation/blob/develop/.textlintrc

PR #4033, opened 2021-07-19
Disable video recording in order to (maybe) fix the build

cowboy:
I didn't want to make a mess in https://github.com/cypress-io/cypress-documentation/pull/4004 so I'm doing this here to try to get the failing checks to pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,12 +126,7 @@ workflows:
           requires:
             - build
             - Setup Linux
-          command: |
-            if [ -n "$DOCS_RECORD_KEY" ]; then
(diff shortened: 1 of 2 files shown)


jennifer-shehane:
@cowboy So, I would suggest trying to update to 8.0 of Cypress. It may solve your performance issues with the screen size being back to 1280x720. Shouldn't really be any breaking changes for this project.

jennifer-shehane:
I could make some other suggests possibly also - but it's probably more likely because the Cypress command log is printed A LOT of things in a single test here.

cowboy:
Since we have a little time before merging, I'll work on these issues back in #4004

PR #4034, opened 2021-07-20
fix: fix the code block that was not rendering on the CT page

robertguss:


jennifer-shehane (approved):
I'm not a fan of this auto formatting with line breaks, as an aside. The issue seems to be fixed.

robertguss:
@jennifer-shehane the auto formatting was done in a different PR and it was a modification to the prettier config that will now wrap at 80 characters I believe. So this change will be performed on all changes going forward, and not just this PR.

PR #4038, opened 2021-07-21
refactor: formatted all .md files to conform to prettier wrapping

robertguss:


robertguss:
@cowboy This is the new PR for the markdown formatting. I spoke to Amir about turning on percy for UI testing and we think that we can probably just spot check the netlify deploy preview. This would basically be the same thing as reviewing all of the Percy tests. Also, if for some reason there is an issue after we push this to prod, we can easily log into Netlify and click a button to point to the previous deploy.

amirrustam (approved):

cowboy:
> @cowboy This is the new PR for the markdown formatting. I spoke to Amir about turning on percy for UI testing and we think that we can probably just spot check the netlify deploy preview. This would basically be the same thing as reviewing all of the Percy tests. Also, if for some reason there is an issue after we push this to prod, we can easily log into Netlify and click a button to point to the previous deploy.

I've never used Percy before, but I was hoping that instead of needing to review all the Percy tests, it would just compare all the before and after images and tell us if there were any visual changes, and since we'd expect zero changes, it would be very easy to tell if there were any problems.

Either way, if @amirrustam is OK with it just getting a manual spot check, that's fine with me.

cowboy (approved):

PR #4042, opened 2021-07-23
docs(api): add information on cy.intercept and request caching

flotwig:
Adding these docs in response to repeated inquiries about this subject.


cowboy (approved):
This makes a lot of sense to me!

PR #4048, opened 2021-07-28
Refactor integration tests

cowboy:
This is based on my work in #4004 to attempt to get the build to work

* Before: one gigantic single test (per spec file) with a gazillion assertions. After: per-page tests, grouped into categories
* The test code feels significantly more readable to me, and we no longer have inconsistent assertions and logic across spec files
* I'm pretty sure this covers every case we were testing previously
* The build passes now, whereas pre-refactor it was failing in #4004

@kevinold asked me some questions, so I thought I'd answer them here for anyone to read:

* The reason these tests are dynamic is because they were already dynamic, and I can't spent a ton of time refactoring them. However, due to the repetitive nature of what's being tested, I think the dynamic tests make sense and add value.
* The reason the refactored tests are in 4 spec files is because I had them all in one file and it was taking 6 minutes, splitting them back into 4 files allows them to run in parallel (and keeps them as far away from collectively timing out as possible).
* The reason the dynamic code isn't duplicated in each of the 4 spec files is because that's what we had before, and each spec file was being edited inconsistently, even though all 4 spec files were doing the exact same thing. It was _very_ hard to reason about. Repetition is valuable for manually-defined tests, but it's not valuable for dynamically defined tests.
* The reason I changed the `testFiles` glob pattern and created an `integration/test-utils` directory for the shared logic is because I felt it made it more clear how the `test-utils` file was being used. I thought about putting it in the `/cypress/support` directory, but that seems like it's for things to be loaded before every test. I thought about putting it in a new `/cypress/helpers` directory, but I was concerned that would be confusing to someone just browsing the directory structure for the first time. It also seemed like way too large a function to be made into a custom command, especially since it's running outside of tests. Since we don't have an established pattern for helper files, I felt that putting it inside `integration` makes it most clear where it's being used. And the `testFiles` glob pattern had to be changed at that point to prevent the helper file from showing up in the test runner.


kevinold (approved):
@amirrustam I had a call with @cowboy to chat about the updates in this PR and to gain my own knowledge of these tests directly from him.  I asked for additional comments around the the why's of the overrides to the best of his understanding, which he has added to the PR description.

In the end, I think these changes put us in a better position for optimization and parallelization of the these tests.

amirrustam (approved):

PR #4077, opened 2021-08-30
Add decode option to cy.url() doc

sainthkh:
Related to cypress-io/cypress#17399


content/api/commands/url.md
```diff
@@ -55,6 +56,15 @@ cy.url().should('include', '/users/1/edit') // => true
 cy.url().should('eq', 'http://localhost:8000/users/1/edit') // => true
 ```
 
+### `decode` option
+
+When the URL contains non-ASCII characters, use the `decode` option.
+
+```javascript
+// For the curious, '사랑' means 'love' in Korean.
```
jennifer-shehane, line 64: ❤️

jennifer-shehane (approved):

PR #4090, opened 2021-09-15
Refactor ::include directive to facilitate adding other directives

cowboy:
Refactor the `::include` [remark directive](https://github.com/remarkjs/remark-directive) to facilitate adding other directives in the future.

See the [Contributing to Cypress Documentation](https://github.com/cypress-io/cypress-documentation/blob/master/CONTRIBUTING.md#partials) for an explanation of how `::include` works (TL;DR: it's basically partials).

# Example Future Usage

In addition to using partials throughout the docs to reuse deprecation alerts, I plan on using this system to allow code like this:

    :::cypress-config-example

    ```js
    {
      includeShadowDom: true,
      experimentalSessionSupport: true,
      viewportWidth: 1280,
      viewportHeight: 1000,
    }
    ```

    :::

to generate tabbed interfaces like this
image

<details>
  <summary>via code like this</summary>

```js
const vm = require('vm')

const replacer = (key, value) => {
  if (typeof value === 'function') {
    throw new Error(`Function values not supported in json: ${key}`)
  }

  return value
}

function processNode(node, { error, warn }) {
  const { children = [] } = node
  const [{ type, value } = {}] = children

  if (type !== 'code') {
    return error(`Expected a code block inside directive`)
  }

  const configStr = value.trim()
  let configObj

  try {
    configObj = vm.runInNewContext(`(${configStr})`, {})
  } catch (err) {
    return error(`Unable to parse code`, err)
  }

  let jsonCodeBlock = ''

  try {
    jsonCodeBlock = `<code-block label="cypress.json (deprecated)">

\`\`\`json
${JSON.stringify(configObj, replacer, 2)}
\`\`\`

</code-block>`
  } catch (err) {
    warn(`${err.message} (skipping cypress.json tab)`)
  }

  return `<code-group>
<code-block label="cypress.config.js" active>

\`\`\`js

module.exports = defineConfig(${configStr})
\`\`\`

</code-block>
<code-block label="cypress.config.ts">

\`\`\`ts
import { defineConfig } from 'cypress'

export default defineConfig(${configStr})
\`\`\`

</code-block>
${jsonCodeBlock}
</code-group>`
}

module.exports = {
  type: 'containerDirective',
  name: 'cypress-config-example',
  processNode,
}
```

</details>


jennifer-shehane (commented):
Very interesting. I think this is a good solution for all the places this config exists. 

We used to have partials for the docs, but the migration didn't support them anymore, so there's a few places of duplicated code (like the "Allowed config values"....which I am seeing are not matching content anymore). https://github.com/cypress-io/cypress-documentation/blob/master/content/guides/core-concepts/writing-and-organizing-tests.md#allowed-config-values

Is this ready for a full review or is there going to be some uses of the new syntax included with tests around it?

cowboy:
> Is this ready for a full review or is there going to be some uses of the new syntax included with tests around it?

Do you have any suggestions for how I can test this? It should behave exactly the same as before, I just tried to better organize the code.

cowboy:
Ok, unit tests have been added. @jennifer-shehane please review when you can, thanks!

jaffrepaul (approved):
Conceptually this should be a nice addition. 

The full execution is a little fuzzy to me, but due to things like all the hoops `processor()` has to jump thru...not the stuff that you've augmented. I'm still getting up to speed on this repo so with the new tests passing, I feel good to give it a go.

amirrustam (commented):
@cowboy did this add any additional performance overhead when iterating on content?

cowboy:
> @cowboy did this add any additional performance overhead when iterating on content?

To reiterate, I didn't _add_ anything, I just refactored existing code to be easier to extend, and added tests.

PR #4102, opened 2021-09-21
UNIFY-327 Update all plugin file examples

jaffrepaul:
https://cypress-io.atlassian.net/browse/UNIFY-327


content/api/commands/task.md
```diff
@@ -306,8 +321,9 @@ are using
 [cypress-skip-and-only-ui](https://github.com/bahmutov/cypress-skip-and-only-ui)
 plugin and want to install your own task to read a file that might not exist:
 
-```javascript
-// in plugins/index.js file
+:::cypress-plugin-example
```
jaffrepaul, line 324: Let's confirm this is what we want here since the example changed a bit with the new directive.

content/guides/component-testing/framework-configuration.md
```diff
@@ -39,18 +39,20 @@ Create React App. We can do this easily using the `react-scripts` plugin
 provided by Cypress. Place the following in `cypress/plugins/index.js`, creating
 the relevant directories.
 
-```js
-// cypress/plugins/index.js
+:::cypress-plugin-example
```
cowboy, line 42: The directive needs to be updated to support an attribute that allows us to specify the config property that `setupNodeEvents` displays in. Right now, it's hard-coded to `e2e` but some examples will need it to be `component`. We'll also need an attribute to suppress the `setupNodeEvents can be defined in either the e2e or component configuration` comment.

See the `include` directive source for an example of using attributes.

So changing the above line to this:

```
:::cypress-plugin-example {configProp=component noComment=true}
```

should render like this:

```js
const injectDevServer = require('@cypress/react/plugins/react-scripts')

  component: {
    setupNodeEvents(on, config) {
      injectDevServer(on, config)
      
      return config
    }
  }
})
```

But any existing examples without those attributes should continue to render as-is.

Once that's done, please update all the examples in this file and the content/guides/component-testing/introduction.md file to use the `{configProp=component noComment=true}` attributes.

content/guides/references/configuration.md
```diff
@@ -260,23 +260,24 @@ While this may take a bit more work than other options - it yields you the most
 amount of flexibility and the ability to manage configuration however you'd
 like.
 
+:::cypress-plugin-example
```
cowboy, line 263: Please revert the changes in this file for now and add a note to UNIFY-366.

cowboy (changes requested):
Overall, looks good. Take a look at the comments and LMK if you have any questions!

content/api/cypress-api/env.md
```diff
@@ -161,16 +161,17 @@ that's been
...
 
+:::cypress-plugin-example
+
 ```js
-// cypress/plugins/index.js
-module.exports = (on, config) => {
-  config.env.sharedSecret =
-    process.env.NODE_ENV === 'qa' ? 'hoop brick tort' : 'sushi cup lemon'
+config.env.sharedSecret =
+  process.env.NODE_ENV === 'qa' ? 'hoop brick tort' : 'sushi cup lemon'
```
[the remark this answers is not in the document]
jaffrepaul: Got it. This sort of thing will be handled in UNIFY-366. Just to be clear does as-is refer to the original version or this update?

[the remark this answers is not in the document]
cowboy: As it currently is, with the updates you've already applied in this branch. Otherwise I'd ask you to revert your changes!

content/api/commands/task.md
jaffrepaul: Wrapped the comments up in these fixes cecc092f4f4d04ec2bc24b0b48bbcf4968720671

jaffrepaul:
In case the update got lost in the comments....

Wrapped the comments up in these fixes cecc092

cowboy (approved):
LGTM!

PR #4111, opened 2021-09-24
Allow reloading directives while developing with `yarn start`

cowboy:
Because the alternative is to wait a really long time between pressing `Ctrl-C` and waiting for `yarn start` to complete


jaffrepaul (approved):
LGTM. Extra comments and logging, yas 🎉

PR #4126, opened 2021-09-29
UNIFY-366 update special configuration refs

jaffrepaul:
https://cypress-io.atlassian.net/browse/UNIFY-366

Small updates made after discussing nuanced implementation + assorted feedback.

### DIFF
--- a/content/api/plugins/configuration-api.md
+++ b/content/api/plugins/configuration-api.md
@@ -139,9 +139,7 @@ between them like:
 How you choose to organize your configuration and environment variables is up to
 you.
 
-:::cypress-plugin-example
-
-```js
+```javascript
 // promisified fs module
 const fs = require('fs-extra')
 const path = require('path')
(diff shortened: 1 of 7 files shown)


content/api/plugins/configuration-api.md
```diff
@@ -129,6 +129,14 @@ in the **Settings** tab of the Test Runner.
 
 ### Switch between multiple configuration files
 
+<Alert type="warning">
```
cowboy, line 132: Let's add the `::include{file=partials/warning-plugins-file.md}` warning here instead. It'll be usable once https://github.com/cypress-io/cypress-documentation/pull/4129 gets merged

cowboy (changes requested):

content/api/plugins/configuration-api.md
cowboy: @jaffrepaul you should be good to make this change now!

cowboy (approved):

PR #4128, opened 2021-09-30
Add link to the cypress-data-session plugin

bahmutov:
Flexible data setup
- https://github.com/bahmutov/cypress-data-session

### DIFF
--- a/content/_data/plugins.json
+++ b/content/_data/plugins.json
@@ -595,6 +595,13 @@
           "link": "https://github.com/BrandedEntertainmentNetwork/cypress-rest-graphql",
           "keywords": ["api", "rest", "graphQL"],
           "badge": "community"
+        },
+        {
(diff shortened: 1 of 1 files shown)


cowboy:
I'm concerned that users may be confused about what this plugin might be used for, since has the name "session" in it, but seems to have nothing to do with sessions.

cowboy:
@bahmutov have you considered a name like `cypress-persistent-data` or `cypress-memoize`?

cowboy:
> Nah, the users will be confused anyway, and confusion can be solved by the readme and examples

If it's likely to to confuse users, we're not going to be able to add it here. If you can rename the plugin to something that doesn't have `session` in the name and better expresses what the plugin is doing, feel free to open a new PR, and we'll re-review.

Thanks!

PR #4140, opened 2021-10-05
Add --component and --e2e CLI args

BlueWinds:
Adds --component and --e2e CLI args to open and run commands, and replaces all references to run-ct and open-ct with them.

Related to https://github.com/cypress-io/cypress/pull/18399


content/guides/references/migration-guide.md
```diff
@@ -470,14 +470,14 @@ module.exports = (on, config) => {
 To run your component tests you _must_ use the dedicated component testing
 subcommands.
 
-- `cypress open-ct`
-- `cypress run-ct`
+- `cypress open --component`
```
cowboy, line 473: I don't think we want to change usage instructions for older versions of Cypress, since those version won't have the new functionality. However, it may be valuable for there to be a "Migrating to Cypress ___" section added to the top of the migration guide that explains these changes. Would that be done as part of this PR, or as part of some separate process?

cowboy (changes requested):

PR #4141, opened 2021-10-06
8.6.0 Release

chrisbreiding:
- [ ] 8.6.0 changelog
- [x] Add note about unselecting all selected options with cy.select([]) (#4134)
- [x] Add "overwrite" option to cy.screenshot() (#4122)


brandon-leapyear:
I see `browser-versions.json` on the `v8.6.0` tag says the chrome version is `94.0.4606.71`, but the docs say `93.0.4577.58`; should the docs be fixed?
https://github.com/cypress-io/cypress/blob/v8.6.0/browser-versions.json

PR #4142, opened 2021-10-06
Rewrite that horrible CodeGroup component to actually work in dev

cowboy:
Test locally with `yarn start`

And let's make sure it works in the build preview as well:
https://deploy-preview-4142--cypress-docs.netlify.app/

Notes:
- Should actually render tabs + code in both `yarn dev` and `yarn build`
- Should highlight the `active` tab and show the `active` code block by default
- Should highlight the correct tab and show the correct code block when tab clicked
- Should not break page scrolling when visiting a page with an anchor, eg. the **Argument should be serializable** heading should be at the top of the page and not jump around when visiting https://deploy-preview-4142--cypress-docs.netlify.app/api/commands/task#Argument-should-be-serializable

### DIFF
--- a/components/global/CodeBlock.vue
+++ b/components/global/CodeBlock.vue
@@ -10,11 +10,26 @@ export default {
       default: false,
     },
   },
+  data() {
+    return {
+      activeState: this.active,
+    }
+  },
+  beforeMount() {
+    if (this.$parent && this.$parent.registerCodeBlock) {
+      this.$parent.registerCodeBlock(this)
+    }
+  },
+  methods: {
+    setActiveState(state) {
+      this.activeState = state
+    },
+  },
 }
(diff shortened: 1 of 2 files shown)


amirrustam (commented):
@cowboy I'm seeing this on CT Framework configuration page:
CleanShot 2021-10-06 at 12 20 00@2x

Location of issue: https://deploy-preview-4142--cypress-docs.netlify.app/guides/component-testing/framework-configuration#React-Create-React-App

jaffrepaul (commented):
I'm seeing the same thing as Amir on both CT pages with no errors in the console or otherwise. Other than that it looks good. 

The not break page scrolling tho 🥇

amirrustam (approved):

PR #4179, opened 2021-10-26
UNIFY-498-add-E2E-callout-to-pages

jaffrepaul:
https://cypress-io.atlassian.net/browse/UNIFY-498

### DIFF
--- a/content/guides/testing-strategies/working-with-graphql.md
+++ b/content/guides/testing-strategies/working-with-graphql.md
@@ -1,6 +1,5 @@
 ---
 title: Working with GraphQL
-e2eSpecific: true
 ---
 
 <Alert type="info">
--- a/content/guides/testing-strategies/working-with-graphql.md
+++ b/content/guides/testing-strategies/working-with-graphql.md
@@ -1,5 +1,6 @@
 ---
 title: Working with GraphQL
+e2eSpecific: true
 ---
 
 <Alert type="info">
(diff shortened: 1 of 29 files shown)


pages/examples/_.vue
```diff
@@ -142,7 +144,7 @@ export default {
       />
       <div v-if="mediaObjectIsEmpty" class="main-content-article-wrapper">
         <article class="main-content-article hide-scroll">
-          <h1 class="main-content-title">{{ exampleItem.title }}</h1>
+          <MainContentHeader :title="exampleItem.title" />
```
jaffrepaul, line 147: Using the component on this page template actually updated the page title size of the sub pages under 'media' to match the rest of the pages. Which I think is what we'd want anyway.

content/guides/testing-strategies/working-with-graphql.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Working with GraphQL
+e2eSpecific: true
```
cowboy, line 3: It looks like this page isn't e2e specific, it's just this specific example needs to have a tabbed example https://docs.cypress.io/guides/testing-strategies/working-with-graphql#Modifying-a-Query-or-Mutation-Response

Please revert this change and I'll move the line in the spreadsheet to UNIFY-506.

cowboy (changes requested):

cowboy (approved):

PR #4191, opened 2021-10-31
Change `run` to `open` Cypress command

kubk:
Since it's section about `cypress open` the `open` command should be used instead of `run`. Related issue: https://github.com/cypress-io/cypress/issues/18712


jennifer-shehane (approved):
Thanks!

PR #4208, opened 2021-11-13
docs: add hover as a custom command

mysticdevx:
closes #18887


BlueWinds:
Thanks for putting something in. We've discussed this a bit internally, and agree that it's confusing as things stand - we've opened https://github.com/cypress-io/cypress/issues/18941 to fix it in a more comprehensive way at a later date. For now, just updating the docs seems like a good interim step.

BlueWinds (approved):

cowboy (approved):
LGTM :shipit:

PR #4212, opened 2021-11-15
DX-739: Migrated and converted root content to next-docs

AGS1130:
DX-739


robertguss:
@AGS1130 Hey Greg, for this PR we actually want to get all of the content inside of the `_cypress-next-js-docs` into the root of the docs site, not copy all of the content into that folder. If we copy everything into that folder, we will loose all of the git history of those files. 

Also, this PR is massive with over 400 files changed. so maybe we can limit the scope a bit. The content folder should not move and remain in place and all of the `next.js` files need to ultimately replace the `nuxt` config files, etc. 

I apologize that this was not clear as my putting the `_cypress-next` directory may have been confusing. My apologies. If you want to jump on a call with me, we can pair program some of this stuff together. Just hit me up in slack.

AGS1130:
@robertguss is this what you're asking me for first steps?

AGS1130:
Removing to work in main repo.

PR #4216, opened 2021-11-16
UNIFY-520 Add component testing content to `cy.session()` docs

cowboy:
JIRA
https://cypress-io.atlassian.net/browse/UNIFY-520

Deploy preview
https://deploy-preview-4216--cypress-docs.netlify.app/api/commands/session

### DIFF
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -32,7 +32,7 @@ Enabling this flag does the following:
   [`Cypress.Cookies.defaults()`](/api/cypress-api/cookies#Defaults) methods.
 
 Because the page is cleared at the beginning of each test,
-[`cy.mount()`](/api/commands/mount) or [`cy.visit()`](/api/commands/visit) must
+[`cy.visit()`](/api/commands/visit) or [`cy.mount()`](/api/commands/mount) must
 be explicitly called at the beginning of each test.
 
 </Alert>
@@ -735,8 +735,8 @@ to explicitly log out.
 | After `cy.session()` | <Icon name="check-circle" color="green"></Icon> |                                                 |
 
 Because calling `cy.session()` clears the current page in addition to restoring
-cached session data, [`cy.mount()`](/api/commands/mount) or
-[`cy.visit()`](/api/commands/visit) must always be explicitly called afterwards
+cached session data, [`cy.visit()`](/api/commands/visit) or
+[`cy.mount()`](/api/commands/mount) must always be explicitly called afterwards
 to ensure a page is visited or a component is mounted.
 
 ### Session caching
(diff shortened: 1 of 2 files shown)


content/api/commands/session.md
```diff
@@ -31,9 +31,9 @@ Enabling this flag does the following:
   [`Cypress.Cookies.preserveOnce()`](/api/cypress-api/cookies#Preserve-Once) and
   [`Cypress.Cookies.defaults()`](/api/cypress-api/cookies#Defaults) methods.
 
-Because the page is cleared before each test,
-[`cy.visit()`](/api/commands/visit) must be explicitly called in each test to
-visit a page in your application.
+Because the page is cleared at the beginning of each test,
+[`cy.mount()`](/api/commands/mount) or [`cy.visit()`](/api/commands/visit) must
```
jaffrepaul, line 35: I believe most if not all other references list visit before mount. This doesn't have any deeper meaning other than mount came later and has been plugged into content in that order.

content/api/commands/session.md
```diff
@@ -636,8 +735,9 @@ to explicitly log out.
 | After `cy.session()` | <Icon name="check-circle" color="green"></Icon> |                                                 |
 
 Because calling `cy.session()` clears the current page in addition to restoring
-cached session data, [`cy.visit()`](/api/commands/visit) must always be
-explicitly called afterwards to ensure a page is visited.
+cached session data, [`cy.mount()`](/api/commands/mount) or
+[`cy.visit()`](/api/commands/visit) must always be explicitly called afterwards
```
jaffrepaul, line 739: Same here on the order for consistency.

jaffrepaul (changes requested):
Seems solid. I left one small comment. 

In general, this is def a dense page and I'm going to spend more time reading through it. I've done that in the past but may have a different/better perspective now-a-days.

Also, one other thing that stuck out to me is the new badge that simple reads "Component Only". It feels like that may not have enough context to be effective. Although I don't love the alternative either, i.e real long badges like End-to-End Testing Only & Component Testing Only. Gonna think on that one.

elylucas (approved):
we chatted through any concerns i had, so lgtm

content/api/commands/session.md
cowboy: I actually did this deliberately. I think we should actually reword all of our other examples to put mount before visit because if we don't want to appear like we're favoring one over the other, the most sensible way to list things would be alphabetically. But I'm happy to change this to visit followed by mount to be more consistent. We could always re-order them in a separate story if we wanted. @amirrustam, thoughts?

cowboy:
> Also, one other thing that stuck out to me is the new badge that simple reads "Component Only". It feels like that may not have enough context to be effective. Although I don't love the alternative either, i.e real long badges like _End-to-End Testing Only & Component Testing Only_. Gonna think on that one.

Agreed. I want us to be more consistent with our wording, so I think the badges should either say "Component Only" and "End-to-End Only" or "Component Testing Only" and "End-to-End Testing Only" but perhaps @amirrustam has some thoughts

content/api/commands/session.md
cowboy: I put it back in b7f6e7298130cdfef7d3529f0f26a1dd7c2cf2d4 and we can revisit later if necessary.

PR #4218, opened 2021-11-16
UNIFY-525-stub-onBeforeLoad-directive

jaffrepaul:
https://cypress-io.atlassian.net/browse/UNIFY-525

### DIFF
--- a/scripts/directives/cypress-stub-example.js
+++ b/scripts/directives/cypress-stub-example.js
@@ -1,0 +1,30 @@
+function processNode(node, { _require, error }) {
+  const helpers = _require(__dirname, './helpers/example-helpers')
+  const { children } = helpers.getNodeProperties(node)
+  const { errorArgs, parts } = helpers.getCodeBlocks(children, { count: 2 })
+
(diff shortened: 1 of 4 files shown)


scripts/directives/cypress-stub-example.js
```diff
@@ -0,0 +1,30 @@
...
+      label: 'Component Test',
+      language: 'js',
+      body: componentExample,
+    }
+  )
+}
+
+module.exports = {
+  type: 'containerDirective',
+  name: 'cypress-stub-example',
```
jaffrepaul, line 28: I referred to it as the stub example because `cy.stub()` is used in both docs refs. And in the CT example there is no `onBeforeLoad` hook.

cowboy: It seems like this should just be called `cypress-e2e-component-example`.

cowboy (changes requested):

jaffrepaul:
Moved updates to over to https://github.com/cypress-io/cypress-documentation/pull/4221

PR #4219, opened 2021-11-17
DX-742: Replace nuxt root level config files with next config files.

AGS1130:
DX-742


kevinold:
@AGS1130 @robertguss Thinking we might want to leave the `circleci/.config.yml` as is and add the Next.js version of that file at the bottom, but commented out.  Thoughts?

robertguss:
> @AGS1130 @robertguss Thinking we might want to leave the `circleci/.config.yml` as is and add the Next.js version of that file at the bottom, but commented out. Thoughts?

@kevinold I don't want to loose anything from the current nuxt ci config, or the next either. But do we actually need CI to run at this stage? What if we combine both configs, but commented out everything. I don' think it really makes sense to have all of these jobs running when we know most, if not all, will fail. We can then turn on jobs and portions of the config incrementally as they make sense.

AGS1130:
> > @AGS1130 @robertguss Thinking we might want to leave the `circleci/.config.yml` as is and add the Next.js version of that file at the bottom, but commented out. Thoughts?
> 
> @AGS1130 @kevinold I don't want to loose anything from the current nuxt ci config, or next either. But do we actually need CI to run at this stage? What if we combine both configs, but commented out everything? I don' think it really makes sense to have all of these jobs running when we know most, if not all, will fail. We can then turn on jobs and portions of the config incrementally as they make sense.

@kevinold @robertguss That makes sense to me. We can combine them and comment out Nuxt related configurations for Circle CI.

kevinold (approved):

robertguss (approved):

AGS1130:
@kevinold @robertguss Have a look again at the [config file](https://github.com/cypress-io/cypress-documentation/blob/d0632ca4a37b041974719ee914cf5641fe4f57cc/.circleci/config.yml) I commented out Nuxt related jobs/builds and retained the new config for Next. I'll create a separate issue so we will update the CI config and Cypress tests for working project at this moment.

kevinold:
@AGS1130 Please update so that the full contents of the [existing Circle Config](https://github.com/cypress-io/cypress-documentation/blob/nextjs-docs/.circleci/config.yml) are listed first, complete commented out, then add the contents of the config from the `nextjs-docs` Next folder (https://github.com/cypress-io/cypress-documentation/blob/nextjs-docs/_cypress-nextjs-docs/.circleci/config.yml) below that and comment it out.

The result should be a config.yml with all the lines commented out for the moment.

PR #4223, opened 2021-11-19
UNIFY-504-test-type-badge-links

jaffrepaul:
https://cypress-io.atlassian.net/browse/UNIFY-504

### DIFF
--- a/components/global/Badge.vue
+++ b/components/global/Badge.vue
@@ -1,7 +1,9 @@
 <template>
-  <div :class="[$style.badge, $style[type]]">
-    <slot />
-  </div>
+  <router-link :to="{path}" style="border-style: none">
+    <div :class="[$style.badge, $style[type]]">
+      <slot />
+    </div>
+  </router-link>
 </template>
 
 <script>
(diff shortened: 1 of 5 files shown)


components/global/E2EOnlyBadge.vue
```diff
@@ -9,5 +9,7 @@ export default {
 </script>
 
 <template>
```
jaffrepaul, line 11: I originally made this a path prop to pass into the `<Badge />` component but there are edge cases to account for with conditional CSS, etc that felt like a lot since we only need this on these 2 test type badges atm.

cowboy:
The "Component Only" is now in the `unification` branch, so if you rebase, you should be able to add a link to it.

components/global/Badge.vue
```diff
@@ -1,7 +1,9 @@
 <template>
-  <div :class="[$style.badge, $style[type]]">
-    <slot />
-  </div>
+  <router-link :to="{path}" style="border-style: none">
```
cowboy, line 2: Do this kind of thing

```vue
<template>
  <router-link v-if="path" :to="{path}">
    <div>some text</div>
  </router-link>
  <div v-else>some text</div>
</template>
```

cowboy (approved):

PR #4240, opened 2021-11-29
UNIFY-474-sidebar-nav

jaffrepaul:
https://cypress-io.atlassian.net/browse/UNIFY-474

### DIFF
--- a/components/sidebar/CollapsibleSidebarSection.vue
+++ b/components/sidebar/CollapsibleSidebarSection.vue
@@ -161,8 +161,8 @@ export default {
           </nuxt-link>
         </li>
         <li v-else>
-          <div class="pl-4 pr-2 py-0 text-md font-bold bg-gray-200">
-            {{ child.title }}
+          <div class="pl-4 py-0 mt-4 text-xs font-semibold text-gray-400 grid grid-cols-1">
+            <span class="border-b border-gray-300 border-dashed">{{ child.title }}</span>
           </div>
         </li>
       </div>


components/sidebar/CollapsibleSidebarSection.vue
```diff
@@ -161,8 +161,9 @@ export default {
           </nuxt-link>
         </li>
         <li v-else>
-          <div class="pl-4 pr-2 py-0 text-md font-bold bg-gray-200">
-            {{ child.title }}
+          <div class="pl-4 py-0 mt-4 text-xs font-semibold text-gray-400 grid grid-cols-1 divide-y divide-gray-300 divide-dashed">
+            <span>{{ child.title }}</span>
+            <span></span>
```
cowboy, line 166: Why the extra `span`?

cowboy (changes requested):

components/sidebar/CollapsibleSidebarSection.vue
jaffrepaul: Yeah, experimenting with the best way to handle the dashed horizontal border across the full sidebar width, not just the content. Tailwind has [this approach](https://tailwindcss.com/docs/divide-width#add-borders-between-stacked-children) with adds the separator between children, which there is only really one of. It's not ideal but the thought was this is likely a placeholder for a convo starter with design and can be updated.

cowboy: You should be able to put the border on the span with something like `border-b border-gray-300 border-dashed`

jaffrepaul: Cool, that worked. Tried something similar yesterday but I think I was tacking on unnecessary integers as px value which made for wonk. There are still mobile considerations that need to be added but I'll see where the design convo goes before getting into that detail.

cowboy (approved):
This looks so much better! :shipit:

PR #4245, opened 2021-11-30
UNIFY-616-specPattern

jaffrepaul:
https://cypress-io.atlassian.net/browse/UNIFY-616


content/api/plugins/before-run-api.md
```diff
@@ -74,7 +74,7 @@ module.exports = (on, config) => {
     //     // ... more specs
     //   ],
     //   specPattern: [
-    //     '**/*_spec.js'
+    //     '**/*.cy.{js,jsx,ts,tsx}'
```
cowboy, line 77: The `specPattern` won't default to that value if the `specs` array contains a spec like `relative: 'cypress/integration/login_spec.js'`. This example might need to be double-checked.

content/guides/component-testing/introduction.md
```diff
@@ -114,10 +114,11 @@ look like this:
 
 ```js
 {
-  testFiles: "cypress/integration/*.spec.js",
+  e2e: {
```
cowboy, line 117: Shouldn't this file actually be deleted? See https://github.com/cypress-io/cypress-documentation/blob/unification/netlify.toml#L56-L60

content/guides/getting-started/component-framework-configuration.md
```diff
@@ -51,17 +51,15 @@ const { devServer } = require('@cypress/react/plugins/react-scripts')
 {
   component: {
     devServer,
-    componentFolder: 'src',
-    testFiles: '**/*.test.{js,ts,jsx,tsx}'
+    specPattern: '**/*.cy.{js,jsx,ts,tsx}'
```
cowboy, line 54: It doesn't entirely make sense to recommend that the user specify a `specPattern` option with the default value. Also, the text before the example doesn't match the pattern. 

Perhaps change the text in the paragraph above to say something like this (and maybe link `specPattern` to the place in the configuration guide where it is explained):

> By default, Cypress looks for spec files anywhere in your project with an extension of either `.cy.js`, `.cy.jsx`, `.cy.ts`, or `.cy.tsx`. However, you can change this behavior with a custom `specPattern` value. In the following example, we've configured Cypress to look for spec files with those same extensions, but only in the `src` folder or any of its subdirectories.

And then change the `specPattern` to `'src/**/*.cy.{js,jsx,ts,tsx}'`

This comment applies to all 5 examples on this page.

content/guides/references/configuration.md
```diff
@@ -238,7 +235,7 @@ object:
...
 | `setupNodeEvents` | `null`                     | Function in which node events can be registered and config can be modified. Takes the place of the (deprecated) plugins file. [Please read the notes for examples on using this.](#setupNodeEvents) |
 | `supportFile`     | `cypress/support/index.js` | Path to file to load before test files load. This file is compiled and bundled. (Pass `false` to disable)                                                                                           |
-| `testFiles`       | `**/*.*`                   | A String or Array of glob patterns of the test files to load                                                                                                                                        |
+| `specPattern`     | `**/*.cy.{js,jsx,ts,tsx}`  | A String or Array of glob patterns of the test files to load                                                                                                                                        |
```
cowboy, line 238: Please add a note here to this effect: https://www.notion.so/cypressdx/testFiles-specPattern-glob-patterns-295d545a0af942ebb5b71eda2a6d422b#1a893de703d14709bc159c17559d2a97

content/guides/references/migration-guide.md
```diff
@@ -366,8 +366,7 @@ required for component testing
 ```json
 {
   "experimentalComponentTesting": true,
-  "componentFolder": "src",
-  "testFiles": "**/*spec.{js,jsx,ts,tsx}"
+  "specPattern": "**/*spec.{js,jsx,ts,tsx}"
```
cowboy, line 369: Revert all changes to the Migration Guide, the `specPattern` change won't be retroactively applied to older versions of Cypress

cowboy (changes requested):

content/guides/references/migration-guide.md
jaffrepaul: Yep 👍 I had been filtering out this page as to not change it. Must have slipped in there during one of the project wide searches.

content/guides/component-testing/introduction.md
jaffrepaul: Yeah you're right. It's still in `unification`. IDK if a previous PR was to handle that. I'll pull it from here.

content/api/plugins/before-run-api.md
jaffrepaul: Agreed. I wasn't certain on this one and forgot to add that note to the PR. I'll check it out when we get the dev build.

content/guides/getting-started/component-framework-configuration.md
```diff
@@ -36,10 +36,11 @@ Configure the dev server to use the same Webpack configuration used by Create
 React App. You can do this using the `react-scripts` plugin provided by the
 `@cypress/react` module.
 
-Also, you need to tell Cypress where to find your component tests. The following
-example configuration assumes that all the test files are somewhere in the `src`
-folder and end with a `.test.js`, `.test.jsx`, `.test.ts` or `.test.tsx`
-extension.
+By default, Cypress looks for spec files anywhere in your project with an
```
cowboy, line 39: You need to make an equivalent change to the text preceding the 4 other examples where `specPattern` is used. People are most likely going to jump to the example that is specific to their needs, and won't see this one. Double-check the examples at https://github.com/cypress-io/cypress-component-testing-examples to figure out what the best non-default `specPattern` value should be for each example.

content/guides/references/configuration.md
```diff
@@ -232,13 +229,13 @@ object:
...
+| `devServerConfig` | `null`                     | Optional dev server configuration. Config options will be determined by the dev server. [Please read the notes for examples on using this.](#devServer-devServerConfig)                                                                      |
+| `setupNodeEvents` | `null`                     | Function in which node events can be registered and config can be modified. Takes the place of the (deprecated) plugins file. [Please read the notes for examples on using this.](#setupNodeEvents)                                          |
+| `supportFile`     | `cypress/support/index.js` | Path to file to load before test files load. This file is compiled and bundled. (Pass `false` to disable)                                                                                                                                    |
+| `specPattern`     | `**/*.cy.{js,jsx,ts,tsx}`  | A String or Array of glob patterns of the test files to load. <br><br>\*Note on matching: component tests will be found using the component.specPattern value but any tests found matching the e2e.specPattern value will be **excluded.\*** |
```
cowboy, line 238: This doesn't seem to be rendering correctly. Can you please double-check the formatting?

cowboy (changes requested):

content/guides/getting-started/component-framework-configuration.md
jaffrepaul: @cowboy We can chat when you review the latest if need be. There's some nuance where I kept the Next.js (Webpack 4) example mostly in tact while it's testing an index component in the `pages` directory. But we also have a note in the Nuxt section about that sort of content being better used as part of e2e tests. I left these in and we can talk through if changes are required to make things more clear.

content/guides/getting-started/component-framework-configuration.md
```diff
@@ -401,17 +401,15 @@ const { getWebpackConfig } = require('nuxt')
         webpackConfig,
       })
     },
-    componentFolder: 'components',
-    testFiles: '**/*.spec.js'
+    specPattern: 'components/**/*.cy.{js,jsx,ts,tsx}'
```
cowboy, line 404: Like with Next.js, I think for Nuxt, we'll want to use a `specPattern` like `'{pages,components}/**/*.cy.{js,jsx,ts,tsx}'`.

content/guides/references/configuration.md
```diff
@@ -232,13 +229,13 @@ object:
...
+| `devServerConfig` | `null`                     | Optional dev server configuration. Config options will be determined by the dev server. [Please read the notes for examples on using this.](#devServer-devServerConfig)                                                                        |
+| `setupNodeEvents` | `null`                     | Function in which node events can be registered and config can be modified. Takes the place of the (deprecated) plugins file. [Please read the notes for examples on using this.](#setupNodeEvents)                                            |
+| `supportFile`     | `cypress/support/index.js` | Path to file to load before test files load. This file is compiled and bundled. (Pass `false` to disable)                                                                                                                                      |
+| `specPattern`     | `**/*.cy.{js,jsx,ts,tsx}`  | A String or Array of glob patterns of the test files to load. <br><br>\*Note on matching: component tests will be found using the `component.specPattern` value but any tests found matching the `e2e.specPattern` value will be **excluded.** |
```
cowboy, line 238:
```suggestion
| `specPattern`     | `**/*.cy.{js,jsx,ts,tsx}`  | A String or Array of glob patterns of the test files to load. <br><br>Note that any files found matching the `e2e.specPattern` value will be automatically **excluded.** |
```

cowboy (changes requested):

cowboy (approved):

PR #4252, opened 2021-12-03
Update List of Valid Test Configuration Config Values 

davidmunechika:
Related issue: https://github.com/cypress-io/cypress/issues/6407
Linked PR: https://github.com/cypress-io/cypress/pull/18896

Updates the list of valid test configuration options that can be updated during test time in `configuration.md`.


content/guides/references/configuration.md
```diff
@@ -329,17 +329,21 @@ configuration:
 
 - `animationDistanceThreshold`
 - `baseUrl`
-- `browser` **note:** filters whether the tests or a suite of tests runs
```
emilyrohrbough, line 332: You are correct that this value cannot be updated , however this is a unique configuration option since it does allow the user to filter if their test should run. For example:
```
if('my test only for chrome', { browser: 'chrome' }, () => {})
```
Might be worth break this into its own section about filtering tests. The above example is used in quite a few places where we describe the test config concept.

emilyrohrbough (approved):

jennifer-shehane:
@davidmunechika I think we should update the Cypress.config file to mention this. I think it does have a note about some values being readonly but it doesn’t specify what values. Can you look into this?

davidmunechika:
@jennifer-shehane Would you prefer I include the list of writeable configuration options on the Cypress.config page as well? Since it already links to the other page, I was thinking of just updating that link to go straight to the list on the Configuration docs page.

Screenshot 2021-12-17 at 8 55 22 AM

jennifer-shehane (approved):

content/api/cypress-api/config.md
```diff
@@ -140,10 +140,10 @@ Cypress.config() // => {defaultCommandTimeout: 10000, viewportHeight: 900, ...}
 
 ### Not all config values can be changed at all times
 
-Some configuration values are readonly and cannot be changed while running a
-test. Anything that's not directly under Cypress's control - like timeouts,
-`userAgent`, or environment variables - will be ignored at run-time. Be sure to
-review the list of
+Some configuration values are readonly and cannot be changed during running a
```
emilyrohrbough, line 143: https://github.com/cypress-io/cypress-documentation/pull/4252/commits/268373ffa9323ad2d9d80eeba27d89b6f3f59b06#diff-915c4bed144b351513cdcaa774106a940b56dc71f5a53116990509f9353c46a1R143

PR #4253, opened 2021-12-03
Dx 761 create a script to transform sidebarjson

robertguss:


scripts/sidebarJsonToYaml.js
```diff
@@ -0,0 +1,35 @@
+const fs = require('fs')
+const { indent } = require('../utils/indentString')
+const sidebarJSON = require('../content/_data/sidebar.json')
+const keys = Object.keys(sidebarJSON)
+let yamlArray = []
```
cowboy, line 5: Please use a YAML parser to convert your data structure to YAML. Don't stringify manually.

https://www.npmjs.com/package/yaml

cowboy (changes requested):

package.json
```diff
@@ -19,11 +19,13 @@
     "lint:js": "eslint --ext .js,.vue .",
     "lint:style": "stylelint **/*.{vue,css}",
     "lint": "yarn lint:js && yarn lint:style && yarn lint:markdown",
+    "sidebar-json-to-yaml": "node scripts/sidebarJsonToYaml.js",
```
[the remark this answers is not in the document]
flotwig: Any time we need to synchronize the sidebar to https://github.com/cypress-io/cypress-services/tree/develop/packages/on - so whenever the sidebar is updated in the main branch, which usually coincides with a release of the test runner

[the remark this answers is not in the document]
cowboy: What consumes the `static/manifest.yml` file that gets generated?

content/_data/sidebar.json
```diff
@@ -372,7 +372,8 @@
           "children": [
             {
               "title": "Table of Contents",
-              "redirect": "/api/table-of-contents"
+              "redirect": "/api/table-of-contents",
+              "slug": "table-of-contents"
```
cowboy, line 376: I don't think it's valid for a sidebar item to have both a `slug` and a `redirect`. [This code](https://github.com/cypress-io/cypress-documentation/blob/master/components/sidebar/CollapsibleSidebarSection.vue) always shows them as mutually exclusive. I think it would be confusing and/or problematic for a given node to have both properties.

cowboy (changes requested):

package.json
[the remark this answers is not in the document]
robertguss: @cowboy this script will run anytime the docs are build. There is a `postbuild` script which runs the `sidebar-json-to-yaml` script. This way the most update version of this file will be available. 

This work is for the Cypress services "on" package which creates a bunch of "on" links for various resources in the docs. The sidebar file was apparently a `.yml` originally but now we are using `.json` and so we need to convert it for the on package to consume it properly.

scripts/sidebarJsonToYaml.js
robertguss: @cowboy I tried use both js-yaml and yaml but none of them are able to output in the specific format we need. https://github.com/cypress-io/cypress-services/blob/develop/packages/on/data/manifest.yml

Which is why I did this manually.

content/_data/sidebar.json
robertguss: @cowboy thanks for pointing this out. I will take a look and make changes as needed.

scripts/sidebarJsonToYaml.js
cowboy: The code that is reading `manifest.yml` is just `yaml.load()` from `js-yaml`, so anything that outputs valid YAML should work, including one of the above libraries.

https://github.com/cypress-io/cypress-services/blob/1386378e4d05e38268bc8708a802b4e53af9aeb9/packages/on/lib/manifest.js#L5

package.json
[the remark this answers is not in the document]
cowboy: Makes sense!

scripts/sidebarJsonToYaml.js
robertguss: @cowboy I don't think that is accurate as if you look at the manifest.yml I linked above there needs to be `.html` extension added to each line. I think the formatting is too specific to use a 3rd party library. Believe me I wish I could, as it would make things much easier, but the format they are expecting is too unique.

content/_data/sidebar.json
robertguss: @cowboy I have reverted the sidebar.json and updated the script to handle these 2 children accordingy.

scripts/sidebarJsonToYaml.js
cowboy: When you create the data structure that will be serialized into YAML, add the `".html"` suffix to the slug, and then serialize the object.

For example, this code:

```js
const input = [
  { slug: 'a' },
  { slug: 'b' },
  { slug: 'c', children: [ { slug: 'd' }, { slug: 'e', children: [ { slug: 'f' } ] } ] },
  { slug: 'g' }
]

const process = arr => arr.reduce((acc, { slug, children }) => ({
  ...acc,
  [slug]: children ? process(children) : `${slug}.html`
}), {})

const output = process(input)

fs.writeFileSync('data/manifest.yml', yaml.dump({ output }), 'utf8')
```

Will generate a `data/manifest.yml` file that looks like:

```yaml
output:
  a: a.html
  b: b.html
  c:
    d: d.html
    e:
      f: f.html
  g: g.html
```

robertguss: @cowboy installed `js-yaml` and used it for the final output, i.e., `yaml.dump()`

cowboy: @robertguss this code is still serializing the YAML data manually. Not only that, but it's now outputting a YAML string, and not a YAML object, like before, so it's likely not working, since the [`normalizeNestedPaths()`](https://github.com/cypress-io/cypress-services/blob/develop/packages/on/lib/manifest.js#L7) function expects an object, and not a string.

We shouldn't ever be writing code that constructs YAML-looking (or JSON-looking) output via string concatenation. It's an anti-pattern, is generally prone to errors, and in most cases, results in significantly more code than is necessary.

Please rewrite this script to generate the proper data structure, which is an object with sub-objects, and then use `yaml.dump()` to serialize that data structure to YAML. Ideally, you'll do this recursively, so that the generated `data/manifest.yml` file will be correct, even if another level of nesting is added (which is actually very possibly going to happen).

Additionally, please don't hard-code in checks against the `title` being `Table of Contents` or `All Assertions`. In those cases, you need to ask yourself if there is something more general you can be checking, to help generate the correct output. However, stepping back for a moment, It's not clear what the correct output should even be in those cases. Have you looked at what's being done with the data in the generated file?

scripts/sidebarJsonToYaml.js
```diff
@@ -0,0 +1,25 @@
...
+    return {
+      ...acc,
+      [slug ? slug : slugify(title).toLowerCase()]: children
+        ? process(children)
+        : `${slug ? slug : slugify(title).toLowerCase()}`,
+    }
+  }, {})
+}
+
+sidebarKeys.map((key) => {
```
cowboy, line 19: Check this out: instead of manually iterating over all the top-level keys of `sidebarJSON` and then processing the corresponding array value, which you then manually merge into the `formattedSidebar` object, why not just get an array of all those array values with [`Object.values()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values) and then flatten that array using [`Array.prorotype.flat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) into something that `process()` understands?

```js
const formattedSidebar = process(Object.values(sidebarJSON).flat())
```

cowboy:
We've been changing the IA and sidebar quite a bit as part of the Unification / CT GA work. As a result, the manifest that will be generated after that branch is merged is going to be quite different than the current manifest. There will be new entries, some old entries will be removed, and others will be renamed.

The entire point of this generated manifest file seems to be to create "implicit" on-links that track the sidebar content, but what happens when entries are removed or renamed? They will no longer be in the manifest, and those on-links will be broken.

You can probably see where I'm going with this. Implicit on-links are a bad idea, and we should avoid them at all costs.

I recommend that we generate a list of all existing on-links generated in this way, look at our analytics / CDN logs to see which actually get used, and then explicitly add them to [links.yml](https://github.com/cypress-io/cypress-services/blob/develop/packages/on/data/links.yml), and stop trying to generate these links dynamically.

jaffrepaul:
Closing this as we won't need after https://github.com/cypress-io/cypress-documentation/pull/4590

PR #4257, opened 2021-12-06
Upgrade cypress to 9.1.1

chrisbreiding:


chrisbreiding:
Not sure what's up with the netlify build failures. I don't have access in netlify to view the errors. Are we using Vercel instead of netlify now? Can the netlify checks be removed?

cowboy:
> Not sure what's up with the netlify build failures. I don't have access in netlify to view the errors. Are we using Vercel instead of netlify now? Can the netlify checks be removed?

We're looking into switching to Vercel, but we haven't done so yet. The Netlify build has been having problems, I just restarted it for you. You might need to get someone to add you to Netlify so you can do so yourself!

PR #4271, opened 2021-12-14
Adding info for CYPRESS_VERIFY_TIMEOUT

tbiethman:
New docs are associated with: https://github.com/cypress-io/cypress/issues/18290

### DIFF
--- a/content/guides/guides/command-line.md
+++ b/content/guides/guides/command-line.md
@@ -657,6 +657,19 @@ cypress verify
 ✔  Verified Cypress! /Users/jane/Library/Caches/Cypress/3.0.0/Cypress.app
 ```
 
+To change the default timeout of 30 seconds, you can set the environment
+variable `CYPRESS_VERIFY_TIMEOUT`:
+
+```shell
+export CYPRESS_VERIFY_TIMEOUT=60000 # wait for 60 seconds
+cypress verify
+```
+
+Note that the `cypress verify` command is executed as part of the `cypress open`
+and `cypress run` commands. The `CYPRESS_VERIFY_TIMEOUT` environment variable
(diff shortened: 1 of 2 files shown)


emilyrohrbough (approved):

mjhenkes (approved):

content/guides/guides/command-line.md
```diff
@@ -657,6 +657,19 @@ cypress verify
...
 
+To change the default timeout of 30 seconds, you can set the environment
+variable `CYPRESS_VERIFY_TIMEOUT`:
+
+```shell
+export CYPRESS_VERIFY_TIMEOUT=60000 # wait for 60 seconds
+cypress verify
+```
+
+Note that the `cypress verify` command is executed as part of the `cypress open`
```
emilyrohrbough, line 668: Is this always ran on these commands or only in certain scenarios? I might be helpful to add this note in reverse under the `cypress open` & `cypress run` sections to add a link to this section

jennifer-shehane: @tbiethman I think it's only run in certain conditions - like if we can't immediately find it installed? Would have to look at code.

tbiethman: The timeout only comes into play when the verify smoke test executes. We run the smoke test when:
* We run the `verify` command directly with --dev option
  * this bypasses the binary check and forces the smoke test to run
* We run the `run`/`open`/`verify` commands and the binary has not been previously validated
  * if not previously validated, the smoke test is run as part of the binary validation process. 
  * there is logic for a --force option in here to force the validation, but it's no longer supported and errors if provided.

emilyrohrbough (approved):

PR #4275, opened 2021-12-15
Ensure prettier --check is run on yarn lint

cowboy:
We were running `prettier --write` but not `prettier --check` in CI, which wasn't causing the build to fail when markdown files weren't correctly formatted. This PR fixes that.


jaffrepaul (approved):
LGTM 🎉

admah (approved):
Looks great!

robertguss (approved):

PR #4277, opened 2021-12-15
Mount command updates

elylucas:


content/api/commands/mount.md
```diff
@@ -49,7 +49,33 @@ Cypress.Commands.overwrite('mount', (jsx, options) => {
...
+<template #vue2>
+
+```js
+import { mount } from '@cypress/vue'
+
+Cypress.Commands.overwrite('mount', (comp, options = {}) => {
+  // Setup options object
+  options.extensions = options.extensions || {}
+  options.extensions.plugins = options.extensions.plugins || []
+  options.extensions.components = options.extensions.components || {}
```
cowboy, line 61: I feel like this kind of thing is better written like:

```js
  if (!options.extensions) {
    options.extensions = {}
  }
  if (!options.extensions.plugins) {
    options.extensions.plugins = []
  }
  if (!options.extensions.components) {
    options.extensions.components = {}
  }
```

Which is especially true in the "Vue Router" example, where we have:

```js
  options.router =
    options.router ||
    createRouter({
      routes: routes,
      history: createMemoryHistory(),
    })
```

Which could be:

```js
  if (!options.router) {
    options.router = createRouter({
      routes: routes,
      history: createMemoryHistory(),
    })
  }
```

content/api/commands/mount.md
```diff
@@ -302,117 +327,17 @@ it('User profile should display user name', () => {
 })
 ```
 
+> The `getStore` method is a factory method that initializes a new Redux store.
```
cowboy, line 330: Instead of `>` use an `<Alert type="info"></Alert>`

content/api/commands/mount.md
```diff
@@ -589,6 +514,172 @@ it('login link should be active when url is "/login"', () => {
...
+  options.store = options.store || getStore()
+
+  // Add Vuex plugin
+  options.extensions.plugins.push(Vuex)
+
+  return mount(comp, options)
+})
+```
+
+> The `getStore` method is a factory method that initializes Vuex and creates a
```
cowboy, line 545: Instead of `>` use an `<Alert type="info"></Alert>`

content/api/commands/mount.md
```diff
@@ -589,6 +514,172 @@ it('login link should be active when url is "/login"', () => {
...
+    install(app) {
+      app.use(store)
+    },
+  })
+
+  return mount(comp, mountOptions)
+})
+```
+
+> The `getStore` method is a factory method that initializes Vuex and creates a
```
cowboy, line 627: Instead of `>` use an `<Alert type="info"></Alert>`

cowboy (changes requested):
I also have a few general comments:
- can we standardize the name of the first argument to `cy.mount()` and call it `component`? Right now, I see `jsx`, `comp` and `component`.
- In the "React Examples" section, "relies on provider" -> "relies on context" and "in that provider" -> "in a provider"

content/api/commands/mount.md
elylucas: i agree with the vue router sample, but in the other ones it triples the lines of code to wrap them in if statements, and some samples have more options set up to do than others. Im cautious of making the commands too long. Thoughts?

elylucas:
> 

done in #30dbc56

content/api/commands/mount.md
elylucas: done in #7888195

content/api/commands/mount.md
elylucas: done in 7888195

content/api/commands/mount.md
elylucas: done in 7888195

cowboy (approved):

PR #4278, opened 2021-12-15
chore: clarify cy.intercept fixture encoding behavior

emilyrohrbough:
https://github.com/cypress-io/cypress/pull/18534 unintentionally introduced a non-passive change in [9.0.0.](https://docs.cypress.io/guides/references/changelog#9-0-0) The new functionality added was intended to be passive, but 
 there was breaking change in `cy.intercept()`. when a fixture was read when passed as a static response.

When a fixture path was specified in cy.intercept() without an encoding value:
- before 9.0.0 it would read as a Buffer
- after 9.0.0 it would use the default `utf8` encoding option . This change aligned the default encoding behavior to `cy.readFile()` and `cy.fixture()`.

These docs changes add the breaking change log & updates the cy.intercept() documentation to expose how to pass the encoding for a fixture in a static response.


mjhenkes (approved):

amirrustam (approved):

content/api/commands/intercept.md
```diff
@@ -1326,30 +1315,33 @@ from propagating to the next matching response handler in line. See
...
+| statusCode        | HTTP response status code                                                                                                                                                                                                         |
+| headers           | HTTP response headers                                                                                                                                                                                                             |
+| body              | Serve a static response body (`object`, `string`, `ArrayBuffer`) (when `fixture` is omitted).                                                                                                                                     |
+| fixture           | Serve a fixture as the HTTP response body (allowed when `body` is omitted). Read the contents with an encoding other than the [default for the fie type](/api/commands//fixture#Encoding), pass the fixture like `path,encoding`. |
```
jaffrepaul, line 1328: this should read 'file type', right? Also, some rogue forward slashes in the slug which will be problematic.

emilyrohrbough: @jaffrepaul yes good eye

PR #4294, opened 2021-12-31
docs: add new vs code snippets.

stu01509:
Add new vs code snippets.


w3bdesign (changes requested):
cypress should be spelled with a capital D

robertguss (approved):

PR #4315, opened 2022-01-19
Minor updates to address lint warnings

mjhenkes:
A pair of minor updates to address lint warnings blocking some PRS from merging. 

Screen Shot 2022-01-19 at 11 38 09 AM

### DIFF
--- a/components/PluginsList.vue
+++ b/components/PluginsList.vue
@@ -24,11 +24,13 @@ export default {
           >{{ pluginType.name }}</a
         >
       </h2>
+      <!-- eslint-disable vue/no-v-html -->
       <p
         v-if="pluginType.description"
         class="my-8"
         v-html="pluginType.description"
       ></p>
+      <!--eslint-enable-->
       <ul :class="$style.pluginsList">
         <li v-for="plugin in pluginType.plugins" :key="plugin.name" class="m-3">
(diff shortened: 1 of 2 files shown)


components/PluginsList.vue
```diff
@@ -24,11 +24,13 @@ export default {
           >{{ pluginType.name }}</a
         >
       </h2>
+      <!-- eslint-disable vue/no-v-html -->
       <p
         v-if="pluginType.description"
         class="my-8"
         v-html="pluginType.description"
```
mjhenkes, line 31: The assumption is that this is intentional.

components/global/DocsImage.vue
```diff
@@ -11,6 +11,7 @@ export default {
     },
     title: {
       type: String,
+      default: '',
```
mjhenkes, line 14: same default as 'alt'

components/PluginsList.vue
cowboy: It's not clear to me if this is intentional. Let me check!

emilyrohrbough (approved):

components/PluginsList.vue
cowboy: It is!

cowboy (approved):

PR #4316, opened 2022-01-19
Updates to selectFile migration guide

BlueWinds:
We've discovered an oversight with selectFIle(), and want to add the fact that it's not supported to the migration guide. For more details, see https://github.com/cypress-io/cypress/issues/19751.

This is a quick-fix while we work on an actual solution (supporting mimeType)

### DIFF
--- a/content/guides/references/migration-guide.md
+++ b/content/guides/references/migration-guide.md
@@ -32,7 +32,7 @@ When the first argument is an object:
   binary file handling in Cypress 9.0.
 - `mimeType`: This property is unsupported in Cypress 9.3.0. If you need
   mimeType support, continue using cypress-file-upload until we add support for
-  this parameter.
+  this parameter. Support will be added in a future update.
 
 In the second argument:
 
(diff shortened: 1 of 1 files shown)


content/guides/references/migration-guide.md
```diff
@@ -30,15 +30,18 @@ When the first argument is an object:
   rather than `Cypress.Blob`.
 - `encoding`: Remove this property. It is no longer needed due to improved
   binary file handling in Cypress 9.0.
-- `mimeType`: Remove this property. It is no longer needed due to improved
-  binary file handling in Cypress 9.0.
+- `mimeType`: This property is unsupported in Cypress 9.3.0. If you need
```
cowboy, line 33: Are we prioritizing adding support for `mimeType` in the near future? if so, maybe we could say something like "Support for mimeType will be added in a future update."

cowboy (changes requested):

content/guides/references/migration-guide.md
BlueWinds: Added "Support will be added in a future update."

emilyrohrbough (approved):

cowboy (approved):

jennifer-shehane (approved):

PR #4321, opened 2022-01-21
Issue-19803: Use filename on disk when referring to fixtures

BlueWinds:
Matching code PR: https://github.com/cypress-io/cypress/pull/19820

I struggled a bit to find a good way to document this change, mostly because it "just works" the way you expect, whereas the previous behavior was the surprising bit.

### DIFF
--- a/content/api/commands/selectfile.md
+++ b/content/api/commands/selectfile.md
@@ -216,7 +216,7 @@ cy.get('input[type=file][multiple]')
     expect(files[1].name).to.eq('file.png')
     expect(files[1].type).to.eq('image/png')
 
-    // But an explicitly specified MIME type is always used.:
+    // But an explicitly specified MIME type is always used.
     expect(files[2].name).to.eq('file.png')
     expect(files[2].type).to.eq('text/plain')
 
--- a/content/api/commands/selectfile.md
+++ b/content/api/commands/selectfile.md
@@ -186,15 +186,45 @@ This will fail unless the file input has the `multiple` property.
 
 </Alert>
 
-### Selecting a file with custom filename, mimeType and lastModified
+### Custom fileName, mimeType and lastModified
 
 ```javascript
-cy.get('input[type=file]').selectFile({
-  contents: 'path/to/file.yml',
-  fileName: 'custom-name.json',
-  mimeType: 'text/plain',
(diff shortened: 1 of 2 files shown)


cowboy:
> I struggled a bit to find a good way to document this change, mostly because it "just works" the way you expect, whereas the previous behavior was the surprising bit.

What if you showed an example like this? It's based on your [allows users to override the inferred filenames and mimetypes](https://github.com/cypress-io/cypress/pull/19820/files#diff-ec360a32960eb1e176699c03725d936d96afe5c5111b083b2a4e6d3324739347R267) test.

```js
// The file name matches that of the file on disk, and the MIME type
// is inferred from the file name, whether you specify a file:
cy.get('input[type=file]').selectFile('cypress/fixtures/example.json').then(($input) => {
  expect($input[0].files[0].name).to.eq('example.json')
  expect($input[0].files[0].type).to.eq('application/json')
})

// Or use an alias to a file:
cy.fixture('example.json').as('myFixture')

cy.get('input[type=file]').selectFile('@myFixture').then(($input) => {
  expect($input[0].files[0].name).to.eq('example.json')
  expect($input[0].files[0].type).to.eq('application/json')
})

// It's also possible to override the file name and MIME type.
cy.get('input[type=file][multiple]').selectFile([
  // file 0
  {
    contents: 'cypress/fixtures/example.json',
    fileName: '1.png',
  },
  // file 1
  {
    contents: '@myFixture',
    fileName: '2.png',
    mimeType: 'text/plain',
  }
])
.then(($input) => {
  // If fileName is specified, it will override the filename on disk:
  expect($input[0].files[0].name).to.eq('1.png')
  expect($input[0].files[1].name).to.eq('2.png')

  // The MIME type will be inferred from the fileName, if specified:
  expect($input[0].files[0].type).to.eq('image/png')

  // But an explicitly specified mimeType will override that:
  expect($input[0].files[1].type).to.eq('text/plain')
})
```

(note: untested)

AtofStryker (approved):

BlueWinds:
@cowboy - Added an example inspired by the above.

Also realized that I forgot to add mimeType to the migration guide in the last docs PR, so adding it here instead.

cowboy:
> @cowboy - Added an example inspired by the above.

This is great!

cowboy (approved):
LGTM

PR #4347, opened 2022-02-10
update docs for vue 3 official release

marktnoonan:


content/guides/component-testing/framework-configuration.md
```diff
@@ -177,8 +177,8 @@ without opening a browser with `npx cypress run-ct`.
 
 The installation and configuration is the same as Vue 2 with the Vue CLI as
 described above. The only difference is the Vue adapter should be installed
-using `npm install @cypress/vue@next` - `@cypress/vue` targets Vue 2, and the
-`next` branch targets Vue 3.
+using `npm install @cypress/vue@latest`. `@cypress/vue@2` targets Vue 2, and the
```
ZachJW34, line 180: Do we need to specify `@latest` since `npm install @cypress/vue` will install the vue 3 version by default? It looks like our standard is to specify a modifier when it is not the default, such as `html-webpack-plugin@4` otherwise we say `html-webpack-plugin`

marktnoonan: For some reason, yesterday `npm install @cypress/vue` still gave me the vue 2 version. Maybe it was a cache thing. I'd like to reorder the whole page a little bit so that we start with Vue 3 and make Vue 2 the footnote at the end, but for today I'll settle for the commands being accurate.

elylucas (approved):

ZachJW34 (approved):

PR #4348, opened 2022-02-11
Oss

conversayShawn:
Updating verbiage for OSS to align with current process

### DIFF
--- a/content/guides/dashboard/organizations.md
+++ b/content/guides/dashboard/organizations.md
@@ -79,14 +79,15 @@ to another organization before you can delete the organization.
 ### Open Source Plan
 
 To support the community, we provide the Open Source (OSS) plan to take
-advantage of our Dashboard Service with unlimited test runs. To qualify, your
-Organization needs the following:
+advantage of our Dashboard Service with unlimited test runs. To qualify, all
+Projects in your Organization must meet the following criteria:
 
-- Project is a non-commercial entity
-- Project is [public](https://on.cypress.io/projects#Change-project-access)
-- Your project source code is available in a public location with an
+- Be a non-commercial entity
+- Be [Public](https://on.cypress.io/projects#Change-project-access) in the
+  Dashboard
+- Your Project source code is available in a public location with an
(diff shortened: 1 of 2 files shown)


elylucas (approved):

content/guides/dashboard/organizations.md
```diff
@@ -78,24 +78,27 @@ to another organization before you can delete the organization.
...
-To support the community, we provide the Open Source (OSS) plan for public
-projects to take advantage of our Dashboard Service with unlimited test runs. To
-qualify, your project needs two things:
+To support the community, we provide the Open Source (OSS) plan to take
+advantage of our Dashboard Service with unlimited test runs. To qualify, your
+Organization needs the following:
 
-- Your project is a non-commercial entity
-- Source code for your project is available in a public location with an
+- Project is a non-commercial entity
```
jaffrepaul, line 85: The points are clear, but the new text reads a little awkwardly. Maybe something like:

- A project that is a non-commercial entity
- A project that is public
- Your project source code is available in a public location with an OSI-approved license
- _Hit the Free Plan limit of 500 Test Recordings_

I'm not sure how to tweak the last one because I'm not sure what we're getting at here. Is the point that they can't exceed that limit of 500 tests?

jaffrepaul (changes requested):
Good stuff. I added one comment about the verbiage.

content/guides/dashboard/organizations.md
jaffrepaul: Cool. That feels better. I'm still fuzzy on the last bullet though and this is more for my own edification...

A project qualifies for the OSS tier, if it hits a limit of test 500 runs? Or is it more like the project exists within/under the Free plan limit of 500 runs in order to be on this plan? It may just be the use of _hit_ that seems confusing.

conversaShawn: Another great point. I changed it to reach. The goal is to see some significant recording activity in the Dashboard.

jaffrepaul (approved):
LGTM 🎉

PR #4375, opened 2022-03-02
UDX-132-fix-subheader-styles

jaffrepaul:
https://cypress-io.atlassian.net/browse/UDX-132

Reset CSS that is squishing h3 & h4 subheaders that include content wrapped in back ticks.

### DIFF
--- a/styles/content.css
+++ b/styles/content.css
@@ -185,17 +185,11 @@
   @apply text-2xl;
   @apply font-bold;
   @apply mt-8;
-  @apply flex;
-  @apply flex-wrap;
-  @apply items-center;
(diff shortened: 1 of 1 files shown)


styles/content.css
```diff
@@ -185,17 +185,11 @@
   @apply text-2xl;
```
jaffrepaul, line 185: I had added these during the work to add the test-type banner in various locations. It was addressing specific content that was wrapping and is no longer relevant. Worse, the byproduct was borking in other places. I reviewed the list we created for that audit and looked through these as a confirmation that we're 👍

PR #4388, opened 2022-03-14
Update docs to scope slowTestThreshold by testing type

BlueWinds:
Code PR: https://github.com/cypress-io/cypress/pull/20569

`slowTestThreshold` is now testing-type specific, and cannot be set globally (except via command line, where we interpret what the user means).


content/guides/guides/module-api.md
```diff
@@ -45,29 +45,29 @@ node e2e-run-tests.js
...
+| `reporterOptions`   | _object_  | Specify [Mocha reporter](/guides/tooling/reporters) options                                                                                                        |
+| `slowTestThreshold` | _number_  | Time, in milliseconds, to consider a test "slow" during `cypress run`. A slow test will display in orange text in the default reporter.                            |
+| `spec`              | _string_  | Specify the specs to run, see examples below                                                                                                                       |
+| `tag`               | _string_  | Identify a run with a tag or tags                                                                                                                                  |
+| `testingType`       | _string_  | Specify the type of tests to execute; either `e2e` or `component`. Defaults to `e2e`                                                                               |
 
 ### Examples
```
emilyrohrbough, line 72: at the bottom of this page, Do we need to add a history entry about this value changing to be testing-type spcific in 10.x?

BlueWinds: I wasn't sure, because nothing else about 10.0 had been added. The fact that no one else has before isn't a good reason not to, though, so added it in.

It seems possible that we'll end up supersceding the list of specific changes with a "Reworked everything; See 10.x changelog for details" link by the time we're through.

elylucas: I created a story for DX to add the other 10 config changes to the history of these files

elylucas (approved):

Issue #4187, opened 2021-10-27
URL Hashes are not working after the dependency upgrade


AGS1130:
### Description

When clicking on a link that contains a URL Hash it does not navigate to the section

### URL of Issue(s)

https://deploy-preview-4183--cypress-docs.netlify.app/guides/references/changelog

### Steps to replicate

1. Go to the url
2. Click the links on the right hand side where all the release versions are
3. See that you do not navigate to the release version.

### Browser

Firefox 93

### Device

- [ ] PC
- [X] Mac
- [ ] iPhone
- [ ] iPad
- [ ] Android Phone
- [ ] Android Tablet

PR #4395, opened 2022-03-18
chore: updating package versions to support m1

JessicaSachs:
The latest **master** and **unification** branches don't build on m1 machines. This PR fixes master (but the changes will also fix the unification branch, too).

To fix the build, I needed to update a couple of modules. `@percy/cypress` needed to be bumped an entire major version and Nuxt needed to be updated so that the native module `sharp` would build correctly on Apple hardware. After updating Nuxt itself, other ecosystem tooling (namely tailwind and postcss) needed to update as well so that media queries would apply. Without updating the tailwind + postcss packages, the styles wouldn't render correctly.

Before I started trying to fix the m1 issue, @elylucas pointed out an issue we had about a year ago when we upgraded Nuxt where the hash routing was broken (reported here https://github.com/cypress-io/cypress-documentation/issues/4187). This issue doesn't seem to exist in the latest versions of Nuxt -- at least not locally.

I want to test the /changelog#0-0-0 hash routing manually against staging before merging this PR in.


content/guides/component-testing/framework-configuration.md
```diff
@@ -369,9 +369,9 @@ Finally, add a component and test:
       }
     },
     async fetch() {
-      this.mountains = await fetch(
-        'https://api.nuxtjs.dev/mountains'
-      ).then((res) => res.json())
+      this.mountains = await fetch('https://api.nuxtjs.dev/mountains').then(
+        (res) => res.json()
+      )
```
JessicaSachs, line 374: Prettier decided a handful of lines needed to update -- I'm not quite sure why.

.circleci/config.yml
```diff
@@ -6,11 +6,11 @@ orbs:
 executors:
   node-executor:
     docker:
-      - image: cimg/node:14.15.1
+      - image: cimg/node:14.16.0
   with-chrome-and-firefox:
     resource_class: medium+
     docker:
-      - image: 'cypress/browsers:node12.14.1-chrome85-ff81'
+      - image: 'cypress/browsers:node16.13.2-chrome97-ff96'
```
JessicaSachs, line 13: Latest versions of Nuxt required running on 14+.

content/guides/tooling/typescript-support.md
```diff
@@ -169,13 +169,13 @@ declare global {
...
-) => {
-  return cy.wrap(subject).type(generateRandomWords(count), options)
-})
+Cypress.Commands.add(
+  'typeRandomWords',
+  { prevSubject: 'element' },
+  (subject /* :JQuery<HTMLElement> */, count = 3, options?) => {
+    return cy.wrap(subject).type(generateRandomWords(count), options)
+  }
+)
```
JessicaSachs, line 178: Prettier re-ran after an update. I'm really unsure **why** these rules are triggering.

nuxt.config.js
```diff
@@ -58,7 +58,7 @@ export default {
   /*
    ** Nuxt.js dev-modules
    */
-  buildModules: ['@nuxtjs/tailwindcss', '@nuxtjs/fontawesome', '@nuxt/image', 'nuxt-build-optimisations'],
+  buildModules: ['@nuxtjs/tailwindcss', '@nuxtjs/fontawesome', '@nuxt/image', 'nuxt-webpack-optimisations'],
```
JessicaSachs, line 61: Nuxt renamed their build optimization library (probably bc they're gonna support Vite **and** Webpack now)

package.json
```diff
@@ -20,7 +20,7 @@
     "lint:style": "stylelint **/*.{vue,css}",
     "lint": "yarn lint:js && yarn lint:style && yarn lint:markdown",
     "test": "./node_modules/.bin/cypress open",
-    "test:unit": "jest",
+    "test:unit": "NODE_OPTIONS=--experimental-vm-modules jest",
```
JessicaSachs, line 23: `rehype-stringify` is now an ESM-only package. We need to invoke jest with experimental module support.

scripts/remark-directives.test.js
```diff
@@ -6,15 +6,15 @@ const unified = require('unified')
 const remarkParse = require('remark-parse')
 const remarkDirective = require('remark-directive')
 const remarkRehype = require('remark-rehype')
-const rehypeStringify = require('rehype-stringify')
-
```
JessicaSachs, line 10: Rehype Stringify's ESM update broke this line, forced the async import below.

patches/github-slugger+1.4.0.patch
```diff
@@ -0,0 +1,51 @@
...
++    .replace(greaterThan, '-gt')
++    .replace(specials, '-')
++    .replace(/\s/g, '-')
++    .replace(twoOrMoreDashes, '-')
++
++  if (str.endsWith('-')) {
++    str = str.slice(0, str.length - 1)
++  }
++
++  return str
```
JessicaSachs, line 48: All of this logic was previously happening in the older patch-package. It's now been carried over.

pages/guides/references/changelog.vue
```diff
@@ -151,7 +151,7 @@ export default {
           <Footer />
         </article>
       </div>
-      <TableOfContents :toc="tableOfContents" :has-banner="Boolean(banner)" />
+      <TableOfContents :toc="tableOfContents" :has-banner="banner" />
```
JessicaSachs, line 154: Saw this when I was debugging -- you don't need to use `Boolean` in your templates.

cowboy (changes requested):
The code changes seem reasonable. I ran `yarn start` locally on my non-m1 mac, and everything appeared to work as-expected. The netlify deploy preview seems good, as well. No idea why Prettier is

cowboy:
> Prettier is reformatting some markdown files because the `yarn fix:markdown` command wasn't run on a previous merge for some reason, which is strange, because the CSS/Markdown check is required. Can you run `yarn fix:markdown` to fix those issues, either in a separate commit or a separate PR? Thanks!

I fixed the Prettier issues in https://github.com/cypress-io/cypress-documentation/pull/4399

package.json
```diff
@@ -83,15 +84,16 @@
     "make-empty-github-commit": "^1.2.1",
     "netlify-cli": "^3.5.0",
     "patch-package": "^6.4.5",
-    "postcss": "^7",
+    "postcss": "8.4.12",
     "postinstall-postinstall": "^2.1.0",
     "prettier": "^2.0.5",
```
cowboy, line 89: The reason prettier changed a few files is because your changes to the lockfile resulted in prettier being upgraded to `2.6.0`. Which is totally fine, but can you update this dep version? By removing the `^` we should no longer see random prettier issues like this unless we're explicitly updating it.

```suggestion
    "prettier": "2.6.0",
```

cowboy (changes requested):

package.json
elylucas: I'll do this @JessicaSachs

cypress/plugins/index.js
```diff
@@ -8,9 +8,6 @@
 // You can read more here:
 // https://on.cypress.io/plugins-guide
 // ***********************************************************
-
-let percyHealthCheck = require('@percy/cypress/task')
-
```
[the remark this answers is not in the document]
elylucas: going to go ahead and merge this PR. If percy causes an issue we can update later.

PR #4428, opened 2022-04-08
docs: update mount api doc for ct changes

elylucas:


components/global/CodeGroupReactVue2Vue3.vue
```diff
@@ -4,13 +4,13 @@
...
     </CodeBlock>
-    <CodeBlock label="Vue 2">
-      <template #alert><slot name="vue2-alert"></slot></template>
-      <slot name="vue2"></slot>
-    </CodeBlock>
     <CodeBlock label="Vue 3">
       <template #alert><slot name="vue3-alert"></slot></template>
       <slot name="vue3"></slot>
     </CodeBlock>
+    <CodeBlock label="Vue 2">
```
elylucas, line 11: switching vue2 and vue3 spots since vue3 is now default

components/global/CodeGroupVue2Vue3.vue
```diff
@@ -1,12 +1,12 @@
...
   <CodeGroup sync-group="vue2-vue3">
-    <CodeBlock label="Vue 2" active>
-      <template #alert><slot name="vue2-alert"></slot></template>
-      <slot name="vue2"></slot>
-    </CodeBlock>
     <CodeBlock label="Vue 3">
       <template #alert><slot name="vue3-alert"></slot></template>
       <slot name="vue3"></slot>
     </CodeBlock>
+    <CodeBlock label="Vue 2" active>
```
elylucas, line 7: switching vue2 and vue3 spots since vue3 is now default

content/api/commands/mount.md
```diff
@@ -179,13 +184,6 @@ include the `cypress.d.ts` file in all your `tsconfig.json` files like so:
 "include": ["./src", "cypress.d.ts"]
 ```
 
-## Additional Mount Commands
```
jaffrepaul, line 182: Is removing this and normalizing the custom command names just for simplicity?

jaffrepaul (commented):
Adding one clarifying question.

content/api/commands/mount.md
elylucas: Feedback from Jess was that we shouldn't have "additional" mount commands, that a dev should only have one mount command that takes care of all the setup needed. So I removed the title and updated the examples to all by `cy.mount` instead of `cy.mountXXX`.

jaffrepaul: 👍

jaffrepaul (approved):

content/api/commands/mount.md
```diff
@@ -5,8 +5,13 @@ componentSpecific: true
 
 <Alert type="warning">
 
-Cypress does not have a `cy.mount()` command out-of-the-box. See below for info
-on how to craft your own.
+Cypress does not have a built-in `cy.mount()` command. The command must be set
+up in your `/cypress/support/component.{js/ts}` file. By default, when you use
```
cowboy, line 9: Is there any way we can refer to this file without including the `/cypress/support/component.{js/ts}` bit? There was a lot of that in the pre-10.0 docs, which resulted in a lot of cleanup work when the default value changed. What if we called it "your component testing `supportFile`" and link that to `/guides/references/configuration#component` ?

content/api/commands/mount.md
```diff
@@ -127,7 +132,7 @@ example as a starting point for customizing your own command:
 <template #react>
 
 ```ts
-import { MountOptions, MountReturn } from '@cypress/react'
+import { MountOptions, MountReturn } from 'cypress/react'
```
cowboy, line 135: Do we know that `MountOptions` and `MountReturn` will still be importable in the same way from `cypress/react`?

cowboy (changes requested):

content/api/commands/mount.md
elylucas: I asked the team to verify

elylucas: They verified that all exports should be the same in the new bundled package.

content/api/commands/mount.md
elylucas: 👍

cowboy (approved):

PR #4432, opened 2022-04-08
docs: update migration guide for ct changes

elylucas:


jaffrepaul (approved):
LGTM 🎉

content/guides/references/migration-guide.md
```diff
@@ -267,12 +261,11 @@ export default defineConfig({
 <Badge type="danger">Before</Badge>
 
 ```js
-const injectDevServer = require('@cypress/react/plugins/craco')
```
cowboy, line 270: CRACO is a valid before/after scenario. Is there a reason this example has been changed to React?

cowboy (changes requested):

content/guides/references/migration-guide.md
elylucas: CRACO fell through the cracks in CT's technical specs and it wasn't yet accounted for in the new architecture. They are aware of it now but not sure it will make the beta release. So I updated the sample to be plain CRA. They also think that they might be able to handle CRACO in the `create-react-app` framework, so the setup between CRA and CRACO might be the same anyways.

cowboy (approved):

cowboy (approved):

PR #4438, opened 2022-04-13
Bugfix: Position is not an option but an argument to click()

k9ert:
In the [click-documentation](https://docs.cypress.io/api/commands/click#Arguments), `position` is listed as an argument, not a s an option. Unfortunately i can't judge whether one or the other works as this seems to be flaky in our case.
I just wanted to do a PR instead of an issue in order to be more constructive in the case that i'm right but maybe i'm wrong here for some reason?


k9ert:
The reason my test was flaky had nothing to do with this workaround. So the workaround did work for me. 
Feel free to close this issue but maybe a clarification of this would be good in the docs.

PR #4441, opened 2022-04-14
Updating CI Integrations in docs - LayerCI is now webapp.io :)

joshdsouza8:
LayerCI is now webapp.io, changing the reference in the docs here & adding some more documentation on how to use webapp.io with cypress.


admah (approved):
LGTM

PR #4455, opened 2022-04-26
feat(plugins): add cypress chrome recorder to plugins page

jordanpowell88:
This PR adds the Cypress Chrome Recorder that was released today to our plugins page


admah (approved):
🚀

Issue #4227, opened 2021-11-23
Documentation examples should follow best practices


marcandre:
In particular, selecting arguments by ID/class is [explicitly recommended](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) to be used only "sparingly", yet [many](https://docs.cypress.io/guides/overview/why-cypress#End-to-end) [examples](https://docs.cypress.io/guides/getting-started/writing-your-first-test#Step-4-Make-an-assertion), including the [best practices](https://docs.cypress.io/guides/references/best-practices#1-Combine-into-one-test) use those.

Docs should follow best practices and *always* use `contains` or `get("[data-attr-cy='...']")` where appropriate.

jaffrepaul:
1. Using IDs or classes in the examples should be updated for consistency with data-*, per the best practices docs.
2. There is a larger conversation kicking off to revise the best practice docs around selectors to consider things like `cy.contains()` in examples, accessibility (using aria roles,labels,etc) and RTL style selectors.

jennifer-shehane:
This seems to have been addressed

PR #4467, opened 2022-05-02
UDX-162-guides-example-selector-updates

jaffrepaul:
https://cypress-io.atlassian.net/browse/UDX-162

Additional pass at updating selectors in code examples to align with our best practices. Specifically, https://github.com/cypress-io/cypress-documentation/issues/4227.

### NOTE: this effort is only to become consistent with our best practices. A follow up convo will take place on revising our best practices and these will be easier to find and replace once they're consistent.

### DIFF
--- a/content/guides/guides/debugging.md
+++ b/content/guides/guides/debugging.md
@@ -39,7 +39,7 @@ cy.mount(<MyComponent />)
 it('let me debug like a fiend', () => {
   __VISIT_MOUNT_PLACEHOLDER__
 
-  cy.get('.selector-in-question')
+  cy.get('[data-testid=selector-in-question]')
 
   debugger // Doesn't work
 })
@@ -66,7 +66,7 @@ execution and add a `debugger` at the appropriate time:
 ```js
 cy.visit('/my/page/path')
 
-cy.get('.selector-in-question').then(($selectedElement) => {
+cy.get('[data-testid=selector-in-question]').then(($selectedElement) => {
   // Debugger is hit after the cy.visit
   // and cy.get commands have completed
   debugger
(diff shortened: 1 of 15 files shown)


content/guides/guides/debugging.md
```diff
@@ -66,7 +66,7 @@ execution and add a `debugger` at the appropriate time:
 ```js
 cy.visit('/my/page/path')
 
-cy.get('.selector-in-question').then(($selectedElement) => {
+cy.get('[data-testid=selector-in-question]').then(($selectedElement) => {
```
admah, line 69:
```suggestion
cy.get('[data-testid="selector-in-question"]').then(($selectedElement) => {
```

I think the attribute values need quotes around them like [this example](https://docs.cypress.io/api/commands/get#Find-5-elements-with-the-given-data-attribute).

admah (changes requested):
Correct me if I'm wrong, but I think all of these types of selectors:

[data-testid=completed]

should have quotes around the attribute values like:

[data-testid="completed"]

content/guides/guides/debugging.md
jaffrepaul: Technically, I think that's more correct. I used the example that we show at the bottom of the table here as a guide https://docs.cypress.io/guides/references/best-practices#How-It-Works and confirmed this works. 

I was going for consistency with these edits but totally open to making a larger scale find and replace if we decide we want something else.

jaffrepaul: Or...maybe we should update that table to be more technically correct and the examples follow that.

admah: @jaffrepaul my vote would be to update the table to avoid similar questions in the future.

jaffrepaul: I can dig that. I'll need to make another pass and update some other selectors. Easy find and replace tho.

admah (approved):
LGTM 🚀

PR #4487, opened 2022-05-22
update single origin warnings to mention experimental cy.origin

conversayShawn:


elylucas (changes requested):
Hi @conversaShawn content looks good, thanks! Since it seems to be the same in each area, could we put it in a reusable partial file and include it in all the places?

Theres an example here: https://github.com/cypress-io/cypress-documentation/blob/5d32342e8e31e5768b4768633a2a39ff36583767/content/api/commands/mount.md?plain=1#L30

elylucas (approved):

PR #4527, opened 2022-06-01
UDX-165: Correct usage of end-to-end vs E2E across docs

debrisapron:
This PR attempts to regularize usage of the terms "end-to-end" and "E2E" across the docs. This required some subjective and possibly tendentious linguistic calls. My reasoning is as follows:

- The proper noun referring to the **practice** is "E2E Testing"
- The name of the Cypress **feature** is also "E2E Testing"
- The adjective used to denote the type of a test or tests is "end-to-end"
- "end-to-end" in title case is "End-to-End"
- When explaining the concept to people who may not have heard it, prefer "end-to-end", possibly with the acronym in brackets

Examples of correct usage:

- "Choose the E2E Testing option"
- "Writing your first end-to-end test"
- "What is End-to-End (E2E) Testing?"

There are some exceptions, but these are the principles I've tried to follow. I've also avoided changing any slugs because I'm already nervous about the amount of churn we're introducing to URLs.

### DIFF
--- a/content/_data/sidebar.json
+++ b/content/_data/sidebar.json
@@ -37,7 +37,7 @@
           "slug": "end-to-end-testing",
           "children": [
             {
-              "title": "Introduction to End-to-End Testing"
+              "title": "Introduction to E2E Testing"
             },
             {
               "title": "Writing Your First End-to-End Test",
--- a/content/guides/component-testing/component-framework-configuration.md
+++ b/content/guides/component-testing/component-framework-configuration.md
@@ -188,8 +188,8 @@ While you could pass in props directly to the page component in a component
 test, that would leave these server-side methods untested. However, an
 end-to-end test would execute and test a page entirely.
 
-Because of this, we recommend using E2E testing over component testing for
-Next.js pages and component testing for individual components in a Next.js app.
+Because of this, we recommend using E2E Testing over Component Testing for
+Next.js pages and Component Testing for individual components in a Next.js app.
 
 #### Sample Next.js Apps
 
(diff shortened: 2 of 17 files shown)


content/guides/component-testing/component-framework-configuration.md
```diff
@@ -188,9 +188,8 @@ While you could pass in props directly to the page component in a component
 test, that would leave these server-side methods untested. However, an
 end-to-end test would execute and test a page entirely.
 
-Because of this, we recommend using end-to-end testing over component testing
-for Next.js pages and component testing for individual components in a Next.js
-app.
+Because of this, we recommend using E2E testing over component testing for
+Next.js pages and component testing for individual components in a Next.js app.
```
jaffrepaul, line 192:
```suggestion
Because of this, we recommend using E2E Testing over Component Testing for
Next.js pages and Component Testing for individual components in a Next.js app.
```

debrisapron: Good spot!

content/_data/sidebar.json
```diff
@@ -37,7 +37,7 @@
           "slug": "end-to-end-testing",
```
jaffrepaul, line 37: The title prop here can be "E2E Testing" as well so the nav reflects this change too.

jaffrepaul (commented):
Added a couple of changes

content/_data/sidebar.json
debrisapron: Talked about this with Brian, I feel that showing users the acronym in the top-level nav presumes too much that they already know what it means. I think users shouldn't need to be educated before understanding our navigation, does that make sense? Anyway, going to mark this as resolved for now and we can always revisit later.

jaffrepaul (approved):
LGTM 🎉

content/_data/sidebar.json
jaffrepaul: Got it.

PR #4542, opened 2022-06-01
Upgrade Cypress tests to v10

debrisapron:


cypress.config.js
```diff
@@ -0,0 +1,14 @@
+const { defineConfig } = require('cypress')
+
+module.exports = defineConfig({
+  projectId: 'ma3dkn',
+  retries: {
+    runMode: 3,
+    openMode: 0,
+  },
+  e2e: {
+    setupNodeEvents() {},
```
jaffrepaul, line 10: Is this line intentional?

debrisapron: The migrator added it, but it's not needed. Removing

PR #4545, opened 2022-06-02
docs: fix incorrect link for react with vite

admah:
Fixing an incorrect link. Previously it went to a Next app.


jordanpowell88 (approved):
Nice catch!

PR #4553, opened 2022-06-06
fix: update gh action version to match latest

admah:
Updated refs on GH Action CI page to reference `v4` instead of `v2`


elylucas (approved):

jaffrepaul (approved):
LGTM 🎉

PR #4558, opened 2022-06-07
fix: update to vue ct samples, closes #4555

### LINKED ISSUE #4555: Component testing: Vue - Spies: `called.with` should be `calledWith`
### Subject

guides/Component Testing / Vue

### Description

The example given asserts spy-calls with the following code
```
it('clicking + fires a change event with the incremented value', () => {
  const onChangeSpy = cy.spy().as('changeSpy')
  cy.mount(Stepper, { props: { onChange: onChangeSpy } })
    .get(incrementSelector)
    .click()
    .get('@changeSpy')
    .should('have.been.called.with', 1)
})
```
I spent a while trying to assert something similar with my own component, but the assertion was always true, no matter what I gave as the second parameter. To correct my test, I had to change the assertion from `have.been.called.with` to `have.been.calledWith`

elylucas:


jaffrepaul (approved):
LGTM 🎉

PR #4566, opened 2022-06-09
Update test case for action-email input

kserjey:
As I understand, there is no more data-testid="action-email" for the input on example page. Somewhere in the repository I found another example with similar input which is queried by classname.


jaffrepaul:
Thanks for the contribution @kserjey! This change was made elsewhere so I'm closing this PR.

PR #4572, opened 2022-06-13
fix: change steps order for github actions example

monuelo:
### Description

In the GitHub Actions example, the artifact upload was performed before the build, and it leads to an error. The correct order should be:
- **Cypress install** (that runs `yarn build`)
- And then, **Save build folder**


hericlesme:
@jaffrepaul @elylucas @mjhenkes  ☝🏽

monuelo:
@elylucas @jaffrepaul @mjhenkes, is this even a real open source project?

admah (approved):
@monuelo thank you for taking the time to make this change. Sorry that it has taken us a bit to get it reviewed. 

In the future, please be mindful of our [Code of Conduct](https://github.com/cypress-io/cypress/blob/develop/CODE_OF_CONDUCT.md) when interacting in our issues.

PR #4603, opened 2022-07-08
rewrite(guides): update step 4 string selector for the input example to class

arthus05:
**Where**
[Here](https://docs.cypress.io/guides/end-to-end-testing/writing-your-first-end-to-end-test#Adding-more-commands-and-assertions), in the `cy.get()` selector.

**What**:
Change the String alias of the `cy.get()` selector to the class of the element and adapt the rest of the section accordingly. 

**Why**:
The input element which we are aiming for, available in the example page `https://example.cypress.io/commands/actions`, does not have the `data-testid` attribute mentioned, so the test fails even following the given steps. Also, in the tutorial video it is used the class attribute instead, which works.


elylucas (approved):

elylucas:
Thanks @arthus05! I added a small section talking about our best practices as updating the kitchen sink to have data-testids would be a larger effort.

PR #4608, opened 2022-07-12
docs(guides/contains): note limitation on default labels

flotwig:
As noted [here](https://github.com/cypress-io/cypress/pull/21285/files#r873417446), this is an existing undocumented quirk of `cy.contains()`, so it doesn't need to go in a release branch.

### DIFF
--- a/content/api/commands/contains.md
+++ b/content/api/commands/contains.md
@@ -343,6 +343,22 @@ yielded before the `.contains()`.
 cy.get('#main').contains('Jane Lane')
 ```
 
+### Default `<input type="submit">` labels
+
+When `value` is omitted in a `<input type=submit>`, the default label is used
+and can be locale-dependent.
+[More info at MDN.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit#omitting_the_value_attribute)
+
+When this happens, the `value` attribute is an empty string, and there is no
+programmatic way for Cypress to filter elements by the label displayed by the
+user agent. It can cause unexpected failures when trying to use `cy.contains()`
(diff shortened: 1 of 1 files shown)


content/api/commands/contains.md
```diff
@@ -343,6 +343,22 @@ yielded before the `.contains()`.
 cy.get('#main').contains('Jane Lane')
 ```
 
+### Default `<input type="submit">` labels
+
+When `value` is omitted in a `<input type=submit>`, the default label is used
```
jaffrepaul, line 348:
```suggestion
When the `value` attribute is omitted from an `<input type="submit">`, the default label is used
and can be locale-dependent.
[More info at MDN.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit#omitting_the_value_attribute)

When this happens, the `value` is an empty string, and there is no
programmatic way for Cypress to filter elements by the label displayed by the
user agent. This can cause unexpected failures when using `cy.contains()`
with submit buttons.

The solution in this case is to:

```js
 // assert the empty string
 cy.get('input').should('have.value', '')
 
 // ---or---
 
 // if possible, set the `value` attribute
 <input type=submit value="Submit" />
 ```
```

jaffrepaul (changes requested):
Nice find! I did not know that. 

I tweaked the verbiage a lil and included a code block/styles for consistency with the other content.

content/api/commands/contains.md
flotwig: edb7a9274

jaffrepaul (approved):
LGTM 🎉

PR #4610, opened 2022-07-13
Updates to `cy.session()` and the `experimentalSessionAndOrigin` configuration option

### LINKED ISSUE #4490: [sessions] Update documentation to not use password in the session id
### Subject

api

### Description

The `cy.session` documentation currently uses the password in the session id. However, since the session id is displayed in the command log, the password should not be used as part of the id.

emilyrohrbough:
Closes #4490

Some small documentation changes around `cy.session()` and the `experimentalSessionAndOrigin` configuration option.

- We should not recommend using sensitive information like tokens/passwords as session identifiers. This information is displayed in the reporter and can be observes/leaked in a video recording and is logged to the DevTools when the command is clicked. It's better we recommend name-spacing or using a uuid to ensure uniqueness, not tokens/passwords.
- The `experimentalSessionAndOrigin` configuration option is not a valid suite-/test-specific configuration option. Remove from docs.

### DIFF
--- a/content/api/commands/session.md
+++ b/content/api/commands/session.md
@@ -778,7 +778,7 @@ const login = (name, token, password, uniqueKey) => {
 ```
 
 The [`uuid`](https://www.npmjs.com/package/uuid) npm package can be used to
-generate random unique ids if a arbitrary name-space does not meet your needs.
+generate random unique ids if an arbitrary name-space does not meet your needs.
 
 ### Common Questions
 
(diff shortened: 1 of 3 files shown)


content/api/commands/session.md
```diff
@@ -850,19 +755,19 @@ You run this, but because `cy.session()` is only being passed `name` as its
 login('user1', 'different-token', 'p4ssw0rd')
 ```
 
-In summary, you need to ensure that the `id` is created from all the parameters
+In summary, you need to ensure that the `id` is unique from all the parameters
 that are used inside the `setup` function that may change, otherwise `id` values
 may collide and create unexpected results.
```
chrisbreiding, line 760:
```suggestion
In summary, you need to ensure that the `id` is unique. Create it from all the
parameters used inside the `setup` function that may change, otherwise `id` values
may collide and create unexpected results.
```

chrisbreiding (approved):

PR #4630, opened 2022-07-30
docs: updating single origin to a temporary trade-off.

modern-sapien:
I updated the permanent trade offs section to have **single origin** in the temporary trade-offs section with some additional context. 

`- Currently, we are working to address each test being bound to a [single origin](#Same-origin), with the use of our experimental [origin command](/api/commands/origin).`


robertguss (approved):

AtofStryker (commented):
I think prettier needs to be run against the markdown file but other than that LGTM.

PR #4631, opened 2022-08-01
docs: add note about react 18

lmiller1990:


elylucas (approved):

debrisapron (approved):

content/guides/component-testing/mounting-react.md
```diff
@@ -19,7 +19,11 @@ components within Cypress's sandboxed iframe and handling and framework-specific
 cleanup.
 
 ```js
+// React 16, 17
 import { mount } from 'cypress/react'
+
+// React 18
```
emilyrohrbough, line 25: Do we have a matrix / list of supported react versions anywhere beyond this code snipped that needs updated?

lmiller1990: There are so many sources of truth for what we support at this point... I feel like we probably should make the docs **THE** source of truth, since anything internal gets out of date quickly.

@elylucas @rockindahizzy  thoughts on how best to communicate/maintain "what we support"? The nice thing about making the docs the source of truth is when making a PR, it has the "have the docs been updated" step, which everyone should be consciously thinking about before ticking and marking a PR as ready for review.

lmiller1990: We have this table we could use: https://docs.cypress.io/guides/component-testing/component-framework-configuration (which says React 16+).

emilyrohrbough: oh nice!! we should prob move that to page one of ct testing??

lmiller1990: Yeah, agreed, let me make this change

rockindahizzy: In my opinion, it makes sense to move away from the `+` when saying which versions we support. I think in theory the goal would be to support new versions as they are released, but it feels a little too broad. React 18 is a great example. The docs imply we support React 18 right now, but its not necessarily officially supported (at least from my understanding).

lmiller1990: I agree, we can't assume forward compatibility. I will update this once the fix for React 18 (https://github.com/cypress-io/cypress/issues/23081) is fixed.

astone123 (approved):

PR #4636, opened 2022-08-03
Maximizing Sessions in Cypress

emilyrohrbough:
VERY very rough draft -- quit a few grammatical errors at this point

### DIFF
--- a/content/api/commands/Maximizing Sessions in Cypress.md
+++ b/content/api/commands/Maximizing Sessions in Cypress.md
@@ -1,0 +1,325 @@
+# Maximizing Sessions in Cypress
+
+As you likely know, Cypress ensures consistent results through test isolation,
+where test isolation is the practice of resetting the state _before_ each test.
+
+Cleaning up state ensures the operations of one test does not affect another
+test downstream. When writing tests, the goal for the test is reliably pass when
+ran standalone or in a randomized order. If you have set up your tests in a way
+that requires the state of a previous test, this could potentially break your
+tests.
+
+A known downside of test isolation is the need to rebuild test and application
+state before each test kicks off. This in in itself takes time to setup before
+the real heavy lifting can be performed to ensure your application is truly
+performing as expected.
+
+This can lead to two issues:
+
+1. longer tests time
+
+If a spec is structured correctly, each concept should be it's own test for
+multiple reasons, readability, easier debugging and requirement traceability.
+However, if a spec contains multiple tests that required authentication and
+random data generate to properly conduct assertions on an application, its
+noticeable, and possibly acceptable to increase test times to ensure test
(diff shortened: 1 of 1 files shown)


content/api/commands/Maximizing Sessions in Cypress.md
```diff
@@ -0,0 +1,325 @@
+# Maximizing Sessions in Cypress
+
+As you likely know, Cypress ensures consistent results through test isolation,
+where test isolation is the practice of resetting the state _before_ each test.
+
+Cleaning up state ensures the operations of one test does not affect another
+test downstream. When writing tests, the goal for the test is reliably pass when
+ran standalone or in a randomized order. If you have set up your tests in a way
+that requires the state of a previous test, this could potentially break your
+tests.
```
mjhenkes, line 10:
```suggestion
Cleaning up state ensures the operation of one test does not affect another
tests downstream. When writing tests, the goal for the test is reliably pass when
run standalone or in randomized order. If you have set up your tests in a way
that requires the state of a previous test, this could potentially break your
tests.
```

content/api/commands/Maximizing Sessions in Cypress.md
```diff
@@ -0,0 +1,325 @@
...
+
+A known downside of test isolation is the need to rebuild test and application
+state before each test kicks off. This in in itself takes time to setup before
+the real heavy lifting can be performed to ensure your application is truly
+performing as expected.
```
mjhenkes, line 15:
```suggestion
A known downside of test isolation is the need to rebuild test and application
state before each test kicks off. This in itself takes time to setup before
the real heavy lifting can be performed to ensure your application is truly
performing as expected.
```

content/api/commands/Maximizing Sessions in Cypress.md
```diff
@@ -0,0 +1,325 @@
...
+1. longer tests time
+
+If a spec is structured correctly, each concept should be it's own test for
+multiple reasons, readability, easier debugging and requirement traceability.
+However, if a spec contains multiple tests that required authentication and
+random data generate to properly conduct assertions on an application, its
+noticeable, and possibly acceptable to increase test times to ensure test
+isolation. The real problem is when multiple specs need this same setup in order
+for the application to be testable and when these tests need to scale for form
+factor testing.
```
mjhenkes, line 28:
```suggestion
If a spec is structured correctly, each concept should be its own test for
multiple reasons, readability, easier debugging and requirement traceability.
However, if a spec contains multiple tests that require authentication and
random data generation to properly conduct assertions on an application, it is
noticeable, and possibly acceptable to increase test times to ensure test
isolation. The real problem is when multiple specs need this same setup in order
for the application to be testable and when these tests need to scale for form
factor testing.
```

PR #4639, opened 2022-08-04
updated syntax highlighting theme for both dark and light themes

robertguss:
- updated syntax highlighting theme to "nightOwl" for both light and dark themes.


jaffrepaul (changes requested):
I was messing with these themes as well. It looks like the accessibility took a little hit with nightowl. I say we go **dracula** for both light & dark and call it a day.

PR #4640, opened 2022-08-04
updated docusaurus to latest version

robertguss:
- updated all docusaurus packages to their latest version per the instructions in the console when running `yarn start`


jaffrepaul (approved):
LGTM 🎉

Looks like some CSS stuff I updated came back. Must be some new class names or something. I'll take another pass at that.

PR #4641, opened 2022-08-04
added baseRegex to guides nav link to fix active highlighting bug

robertguss:
- fixe bug where the guides link in the nav was always active and selected.

### DIFF
--- a/docs/guides/overview/why-cypress.mdx
+++ b/docs/guides/overview/why-cypress.mdx
@@ -1,7 +1,6 @@
 ---
 title: Why Cypress?
 sidebar_position: 10
-slug: /
 ---
 
 :::info
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -89,12 +89,13 @@ const config = {
         title: '',
         style: 'dark',
         logo: {
+          href: '/guides/overview/why-cypress',
           alt: 'My Site Logo',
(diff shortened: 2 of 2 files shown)


jaffrepaul (changes requested):
What does this property end up doing? Wasn't sure how this is working. 

The only thing now is that the "guides" nav element never has a selected state, even when it's the only thing selected.

robertguss:
@jaffrepaul it doesn't have a selected state when you are on the "home page" or why why cypress page. If you click on any of the docs under the guides you will see it get highlighted. I hope that makes sense.

jaffrepaul:
> @jaffrepaul it doesn't have a selected state when you are on the "home page" or why why cypress page. If you click on any of the docs under the guides you will see it get highlighted. I hope that makes sense.

Got it. I hadn't clicked around enough. It feels like a bit of weird UX that it doesn't highlight. Like it's the complete opposite that it used to be "selected" too much and now, not enough 😆 I'm curious to know if this is a bi product of our weird hierarchy or an actual technical issue.

robertguss:
@jaffrepaul I am honestly not sure what the issue is, it is very strange. It seems to work just fine as far as I can tell, the only time it does not activate is when you are on the root route ie "/" which is the why cypress page.

PR #4644, opened 2022-08-04
Heading ids

robertguss:
- used https://docusaurus.io/docs/cli#docusaurus-write-heading-ids-sitedir to manually create heading ID's to keep the same case


robertguss:
@jaffrepaul I manage to find in their docs that they have a CLI to manually create custom heading IDs. I ran it to preserve the case which should match what our current docs site does. The only problem is that some of the headings do not match exactly like prod, but the majority of them should work now with this. The ones that don't work, we can manually update and fix, but this should get us more than 90% there I think.

jaffrepaul (approved):
LGTM 🎉

This is sweet. Fixing up any that get borked because a "/" in the header got removed should be easy enough to fix.

robertguss:
@elylucas I have spoken with Paul about this PR already, but wanted your eyes on it as well. Docusaurus has a CLI that allows us to give headings specific ID's and so I ran the CLI to create custom ID's for every header but maintain the case, like our docs site does currently. This should hopefully solve 90% of the links. Happy to talk more about this when you get back.

PR #4649, opened 2022-08-08
Updated writing your first test doc page

robertguss:
- Removed outdated video at top of the page
- added a link to the testing your first app course in the next steps section at the bottom.


jaffrepaul (commented):
Certainly solves the problem of having old video content. 

Was there any discussion with marketing to re do a video like this but in v10?

Feels like we could do more to promote the Learn site here in additional to the small mention below. Maybe one of those color callouts near the top where the video was to draw more attention to it?

robertguss:
@jaffrepaul yeah we could do a lot more to this page honestly. However, if we do put a call out at the top, to me it kinda renders the content below it obsolete and it's would be a bit confusing to link out to another tutorial first and then have a different tutorial just below it.

My goal with this PR was to simply get this old video out ASAP. 

I personally think we should be pointing people to the learn site for tutorials rather than having tutorials both in our docs and the learn site. So this page could basically act as a "landing page" for the testing your first app course. 

What do you think?

jaffrepaul:
> @jaffrepaul yeah we could do a lot more to this page honestly. However, if we do put a call out at the top, to me it kinda renders the content below it obsolete and it's would be a bit confusing to link out to another tutorial first and then have a different tutorial just below it.
> 
> My goal with this PR was to simply get this old video out ASAP.
> 
> I personally think we should be pointing people to the learn site for tutorials rather than having tutorials both in our docs and the learn site. So this page could basically act as a "landing page" for the testing your first app course.
> 
> What do you think?

Makes sense. I think we can definitely iterate on the content here and how we leverage the Learn site overall. But that will probably come over time.

robertguss:
@jaffrepaul I think we can iterate on it as well. My goal was to get the outdated video off the page in the meantime.

emilyrohrbough (approved):

PR #4653, opened 2022-08-08
feat(plugins): add cypress-graphql to plugins page

RcKeller:
Submitting [cypress-graphql](https://github.com/Shopify/cypress-graphql) as a plugin, which allows users to stub/spy/mock GraphQL operations like you would be able to for a standard REST request. It's fully tested [here](https://github.com/Shopify/cypress-graphql/runs/5664486638?check_suite_focus=true). Thanks, and LMK if you have any questions!


RcKeller:
@admah @elylucas Could one of you please review this?

elylucas (approved):

PR #4662, opened 2022-08-11
docs: add `cypress-vite` to preprocessors

mammadataei:
This PR adds [cypress-vite](https://github.com/mammadataei/cypress-vite) to plugins page!

The `cyrpess-vite` is a cypress preprocessor that compiles specs using vite. It enables users to use the same vite configuration both for development and testing, and also makes it possible to use vite features like `import.meta.*` in e2e tests.


elylucas (approved):
Awesome plugin! Thanks for contributing.

PR #4665, opened 2022-08-12
Patch slugger to maintain heading case

elylucas:


robertguss (approved):

jaffrepaul (changes requested):
LGTM.

We just need to document this before it becomes one of those weird patched things like we have a number of on the current site without the historical context.

elylucas:
> LGTM.
> 
> We just need to document this before it becomes one of those weird patched things like we have a number of on the current site without the historical context.

Great point, i'll add something

jaffrepaul (approved):
👍

Issue #4660, opened 2022-08-10
Reorder dashboard sidebar to move important features closer to the top


debrisapron:
### Subject

Dashboard

### Description

Reorder sidebar

debrisapron:
Closed by #4674

PR #4674, opened 2022-08-16
Reorder Dashboard section sidebar

debrisapron:
Addresses #4660. My aims here are broadly:

- Make section flow better narratively
- Move higher-value features up in hierarchy
- Move more in-the-weeds technical details down

So we have the intro first obviously, followed by the key structural elements, projects and runs. Then we have USP features, Smart Orchestration, Flake & Analytics. Following this, the integrations in what I believe are roughly order of use (based on a brief analysis on FullStory). Finally some more mundane information about Organization & User settings. Note that while these last two are clearly key to the usage of the Dashboard, they are more of a "one-time" thing and also familiar enough concepts that I think only a minority of users will really need to access these pages.


pstakoun (approved):
lgtm

PR #4675, opened 2022-08-17
Add blurb for create from Vue component

astone123:
This PR adds a small section under `mounting-vue` "Your First Component Test" about the new create from component card for Vue components.


content/guides/component-testing/mounting-vue.md
```diff
@@ -80,6 +80,21 @@ describe('<Stepper>', () => {
 
 Here, we have a single test that ensures that our component mounts.
 
+If you already have a component that you want to create a test for and you don't
+have a spec pattern configured for your project, we can scaffold the test for
```
marktnoonan, line 84: >and you don't have a spec pattern configured for your project

The restriction around needing to "not have a spec pattern" will only exist until next week, wonder if we should hold off on the docs until then? If not, I'd suggest this should be more of a warning format up top, it's a bit easy to skim over here, but makes the difference between this feature being present or not.

> we can scaffold the test for you via the Cypress App

Should we explain where this will be scaffolded? 

One minor thing to confirm: I think we refer to the files as _specs_ in the docs and only use "test" to talk about actual tests contained in the spec file (`it('whatever')`). That's probably the right wording to use here - "create a spec for", "scaffold the spec". The precision is probably more important for dashboard customers, where what we mean by "test" is a billing thing.

marktnoonan (changes requested):
Requesting some minor changes, but also I'd suggest we hold off on publishing until next release when it is working with custom spec patterns, and document for that situation, instead of publishing this and republishing next week.

We can still review and merge if we change the base branch to `develop` to get it squared away before release, iirc.

content/guides/component-testing/mounting-vue.md
elylucas: correct, we refer to the collection of tests as a spec.

elylucas:
could we remove the black borders from around the images? The `<DocsImage>` component will put a slight border and a drop shadow and it will look uniform with the rest of the images.

jordanpowell88 (approved):
LGTM. If the `specPattern` restriction gets resolved in the next release we can merge this now and then come back to it or we can wait. Either way is fine

content/guides/component-testing/mounting-vue.md
```diff
@@ -80,6 +80,21 @@ describe('<Stepper>', () => {
 
 Here, we have a single test that ensures that our component mounts.
 
+If you already have a component that you want to create a test for and you don't
```
elylucas, line 83: If they are following this guide, they already have the stepper component to test so the wording here might be confusing. I think we should probably just show how to use the app to create a spec and get rid of the section directly above on creating the spec file manually. They might think they need to do both here.

content/guides/component-testing/mounting-vue.md
lmiller1990: To clarify:

- Spec: a file on your computer
- Suite: `describe` (can contain 0 or more suites and tests)
- Test: `it`

This is the common definitions across all test runners - we should definitely make sure we use the right definitions.

lmiller1990 (commented):
Content seems fine to me -- looks like there's a few minor things.

content/guides/component-testing/mounting-vue.md
astone123: Okay I updated it to say "spec" instead of "test", and I removed the 

> and you don't have a spec pattern configured for your project

condition, since this will be removed in the next release. I'll target `develop` so that this doc will go out with the next release

astone123:
> We can still review and merge if we change the base branch to `develop` to get it squared away before release, iirc.

@marktnoonan I'm kind of confused about the branches. It looks like `develop` is way behind `master`, which I wouldn't expect if we're using `develop` as a way to delay things until release 🤔

elylucas:
> > We can still review and merge if we change the base branch to `develop` to get it squared away before release, iirc.
> 
> @marktnoonan I'm kind of confused about the branches. It looks like `develop` is way behind `master`, which I wouldn't expect if we're using `develop` as a way to delay things until release 🤔

The docs don't follow that same branching methodology (though it may not be a bad idea to start). Sometimes a release branch (ie: 10.x.x) is setup to curate multiple PRs into a single one for releases.

marktnoonan:
Yeah my bad, I was thinking of the instructions from the gatsby site. In which case, we should just hold on merging/deploying till next week for this.

astone123:
> Yeah my bad, I was thinking of the instructions from the gatsby site. In which case, we should just hold on merging/deploying till next week for this.

Fine with me, lets hold off until then.

lmiller1990 (approved):

marktnoonan (approved):

PR #4696, opened 2022-08-29
Update Cypress Studio documentation

### LINKED ISSUE #4678: Update documentation surrounding Cypress Studio
Once we re-introduce Cypress Studio in v10, we need to update our documentation to match the new implementation. A few things come to mind:

https://docs.cypress.io/guides/references/cypress-studio

- Remove the warning banner that says Cypress Studio has been removed
- Take new screenshots for RWA that use the v10 app

astone123:
Closes  #4678 

This PR moves the Cypress Studio page back to Core Concepts and makes updates to reflect changes in version 10. 

We are re-introducing Cypress Studio in the 10.7.0 release #23230


content/_data/sidebar.json
```diff
@@ -204,6 +204,10 @@
               "title": "Cypress App",
               "slug": "cypress-app"
             },
+            {
```
jaffrepaul, line 207: I don't think I would consider this a Cypress Core Concept. It feels like this content being under References makes more sense.

pstakoun: We moved it back here since this is where it was initially, before it was removed in v10. I am ok with either though. Maybe References does make more sense while Studio is experimental..

content/guides/core-concepts/cypress-studio.md
```diff
@@ -2,17 +2,9 @@
 title: Cypress Studio
 ---
```
jaffrepaul, line 3: You can add `e2eSpecific: true` to the front matter at the top of the file, like [this](https://github.com/cypress-io/cypress-documentation/blob/master/content/guides/end-to-end-testing/auth0-authentication.md)

There's logic on the header to pick this up and add the pill next to the header, above the horizontal rule so it's a "page level" concept.

content/_data/sidebar.json
jaffrepaul: References includes things like experiments(as Studio still is), advanced topics, bundled libs, etc. I'd say it feels more at home there vs. a fundamental concept of the product.

content/guides/core-concepts/cypress-studio.md
```diff
@@ -2,17 +2,9 @@
...
 
-<Alert type="warning">
+<E2EOnlyBadge class="my-10"></E2EOnlyBadge>
 
-<strong class="alert-header"><Icon name="exclamation-triangle"></Icon>
```
jaffrepaul, line 7: It's probably worth keeping this banner up top to make it clear that this is still an experimental feature. We take the opportunity to do so on the [session API page](https://docs.cypress.io/api/commands/session). 

The verbiage used here could help simplify the first part of the where it's noted under the Using Cypress Studio section.

jaffrepaul (changes requested):
Cool. Sorry, I didn't think to mention this last time but I think it's worth adding a brief line or two about this being E2E only in the "experimental" l banner. Or the other way around...maybe why this isn't compatible with CT. I don't actually know the specifics here to make a suggestion. 

We have the E2E pill up top on the Studio page, but I think noting something more clearly is a nice touch. Especially considering the bones of v10 and the concern when Studio was temporarily removed. I could see this coming up if we don't state it more clearly.

astone123:
> Cool. Sorry, I didn't think to mention this last time but I think it's worth adding a brief line or two about this being E2E only in the "experimental" l banner. Or the other way around...maybe why this isn't compatible with CT. I don't actually know the specifics here to make a suggestion.
> 
> We have the E2E pill up top on the Studio page, but I think noting something more clearly is a nice touch. Especially considering the bones of v10 and the concern when Studio was temporarily removed. I could see this coming up if we don't state it more clearly.

I added a section titled "Limitations" because I also wanted to mention that we don't support cross origin tests in Studio, so I also mentioned that you can only use it with E2E there

content/guides/references/cypress-studio.md
```diff
@@ -1,14 +1,28 @@
 ---
 title: Cypress Studio
+e2eSpecific: true
```
lmiller1990, line 3: Above screenshot seems usually wide?

Also tip - if you hold SPACE during the screenshot, you can get the real clean "capture window only" style like the previous ones - good to keep in mind.

content/guides/references/cypress-studio.md
```diff
@@ -21,6 +35,12 @@ in a later release. This page is for reference only.
 
 </Alert>
 
+## Limitations
+
+- Studio is currently only available for writing E2E tests, and doesn't yet work
+  with component testing.
```
lmiller1990, line 41:
```suggestion
  with Component Testing.
```

content/guides/references/cypress-studio.md
```diff
@@ -21,6 +35,12 @@ in a later release. This page is for reference only.
 
 </Alert>
 
+## Limitations
+
+- Studio is currently only available for writing E2E tests, and doesn't yet work
```
lmiller1990, line 40:
```suggestion
- Studio is currently only available for writing End To End tests, and doesn't yet work
```

I think we normally capitalize testing types

content/guides/references/cypress-studio.md
```diff
@@ -21,6 +35,12 @@ in a later release. This page is for reference only.
...
+  with component testing.
+- Studio does not support writing tests that use domains of multiple origins.
```
lmiller1990, line 42:
```suggestion
- Cypress Studio does not support writing tests that use domains of multiple origins.
```

We tend to refer to features as "Cypress ___"

content/guides/references/cypress-studio.md
```diff
@@ -98,9 +98,11 @@ file.
 #### Step 1 - Run the spec
 
 We will use Cypress Studio to perform a "New Transaction" user journey. First,
-launch Cypress and run the spec created in the previous step.
+launch Cypress and select E2E testing, then choose a browser to run specs in.
```
lmiller1990, line 101:
```suggestion
launch Cypress and select End To End testing, then choose a browser to run specs in.
```

Just for consistency (docs are a bit of a mess, gotta start somewhere, heh)

lmiller1990 (commented):
Few nits, looks good

lmiller1990 (approved):

jaffrepaul (approved):
LGTM 🎉

PR #4699, opened 2022-08-31
docs: add dev server function config API

lmiller1990:
Docs for https://github.com/cypress-io/cypress/pull/23605


astone123 (approved):
Looks good

lmiller1990:
Just waiting on original feature to be review by Brian.

brian-mann:
From looking at the implementation code it seems like `baseConfig` is yielded to the async `webpackConfig` function, but no example demonstrates that

PR #4713, opened 2022-09-02
Fix anchor

edhgoose:
There is a small bug in https://docs.cypress.io/guides/guides/environment-variables.

The link to "However, [it is possible to set Cypress environment variables from OS-level environment variables](https://docs.cypress.io/guides/guides/environment-variables#Option-3-CYPRESS)." is incorrect, and is missing a trailing `_`


jaffrepaul:
Thanks for spotting this @edhgoose! This was updated in a different PR 🎉

PR #4718, opened 2022-09-07
GitHub Action Docs - Explicit version of Cypress GHA

robertguss:
- updated the GitHub Actions docs to let users know that they should specifically reference the Cypress GitHub Action version number in the GHA config.


admah (approved):
LGTM 🚀

PR #4733, opened 2022-09-14
Update Footer.vue

nbouvrette:
The current Discord link is expired.


jaffrepaul:
Thanks for spotting this @nbouvrette 🙏 The Docs site & architecture has been updated since this PR and this was updated.

PR #4743, opened 2022-09-19
Update media to reflect v10 content

jaffrepaul:
1. Bring back media section & content
2. Fix broken layouts in new docs site
3. Assorted content fixes


docs/examples/media/podcasts-media.mdx
```diff
@@ -0,0 +1,16 @@
...
+:::info
+For an archive of pre-v10 content, see the [Cypress v9 Podcast](http://v9.docs.cypress.io/examples/media/podcasts-media) page.
+:::
+
+<!--
+  If you're looking to add content to our Podcasts page,
+  Go to src/data/podcasts.json and add your podcast.
+-->
+
+<!-- <PodcastsList /> -->
```
admah, line 16: Should this be commented out?

docs/examples/media/talks-media.mdx
```diff
@@ -0,0 +1,16 @@
...
+:::info
+For an archive of pre-v10 content, see the [Cypress v9 Talks](http://v9.docs.cypress.io/examples/media/talks-media) page.
+:::
+
+<!--
+  If you're looking to add content to our Talks page,
+  Go to src/data/talks.json and add your talk.
+-->
+
+<!-- <TalksList /> -->
```
admah, line 16: Should this be commented out?

src/components/webinars-list/index.tsx
```diff
@@ -19,7 +19,7 @@ export default function WebinarsList(): WebinarProps {
             </ul>
           </div>
 
-          <div className="container">
+          {/* <div className="container">
```
admah, line 22: Does this need to be commented out?

admah (changes requested):
@jaffrepaul I had a couple of questions about things that were commented out. 

I noticed that the Blogs list is still placeholder content. Do you want to remove that?

docs/examples/media/podcasts-media.mdx
jaffrepaul: Ah, yes. These are in flight until we work out the details with the Ecosystem team. This is because there's no v10 specific content here and this interim state is to show the pages with only the v9 callout at the top. I'll shift this PR to WIP to avoid confusion.

PR #4744, opened 2022-09-20
Update link to cypress-tags repo

annaet:
The cypress-tags repo has changed owners from annaet to infosum so the link needs to be updated.


jaffrepaul:
Thanks @annaet for the commit! The Docs site was migrated to a new platform at the end of last year and the folder structure has changed. Please rebase this with the latest main branch and we can get it merged ⚡

PR #4745, opened 2022-09-20
add cypress 10, remove cypress 9

mattvCypress:
Request to add a new course here: https://cypressio.slack.com/archives/C7R5R7KBP/p1663681960598429

I removed the version 9 course, however, we can keep it if we see value in it. I am not familiar with the course.

### DIFF
--- a/content/_data/courses.json
+++ b/content/_data/courses.json
@@ -1,12 +1,13 @@
 {
   "courses": [
     {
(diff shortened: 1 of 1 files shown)


mattvCypress:
@elylucas is this something you could review/merge for us (or what's SOP)? Thank you.

elylucas (approved):

PR #4777, opened 2022-10-12
docs: fix component testing vue example

messenjer:
The example for replicating plugins with vue is not working, it use undefined variable like `VApp` used in others example

### DIFF
--- a/content/guides/component-testing/custom-mount-vue.md
+++ b/content/guides/component-testing/custom-mount-vue.md
@@ -108,10 +108,9 @@ plugins.
 <code-block label="JavaScript" active>
 
 ```js
-import { createPinia } from 'pinia' // or Vuex
+import { createPinia } from 'pinia'
 import { createI18n } from 'vue-i18n'
 import { mount } from 'cypress/vue'
-import { h } from 'vue'
 
 // We recommend that you pull this out
 // into a constants file that you share with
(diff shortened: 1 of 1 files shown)


lmiller1990:
Hi! I think VApp is added later, isn't it? https://docs.cypress.io/guides/component-testing/vue/examples#Replicating-the-expected-Component-Hierarchy

Did you go through the entire example and verify it doesn't work?

messenjer:
Sorry, I think the example was not working. I haven't worked on it for 2 months.

It was an after watching the presentation "Cypress Component Testing" of [Jessica Sachs](https://github.com/JessicaSachs) at [ViteConf](https://viteconf.org/).

I have the repo with the example here. https://github.com/messenjer/viteconf-2022-cypress-component-testing-vuei18n

 I can check next week.

messenjer:
Hi! 

I had time to re-read the doc and tried again the examples.

The first example with i18n and pinia doesn't work because it uses Vapp from Vuetify which is not included. And the next example with Vuetify does not use i18n and pinia, so for me the 2 examples are different and not related.

This PR fixes the 1st example by removing Vapp which is not useful.

Regards

Mathieu

jaffrepaul:
@lmiller1990 were you ever able to confirm this?

lmiller1990:
@jaffrepaul I don't think I ended up following through on this. I'm sorry. I did review the docs in general, and I agree they are confusing and could be a lot better. In fact, we are working on this in April 2023 (now). Here's a ticket: https://github.com/cypress-io/cypress-internal/issues/566

I will close this one off, but we are incorporating the feedback and improvements into updating the docs in general. I expect they land this quarter - improvements will be incremental, but the first thing we are looking to do is assess and improve the structure and examples.

PR #4784, opened 2022-10-14
Style error message inline TOC

jaffrepaul:
Style inline table of contents to match current site on all FAQ Error Message pages.


elylucas (approved):
lgtm

PR #4785, opened 2022-10-17
`.invoke()` no longer awaits promises in 11.x

BlueWinds:
Matching code PR: https://github.com/cypress-io/cypress/pull/24203

## User-facing changelog:
I'm not adding an 11.0.0 Changelog file yet so that merging into the `v11` branch later is a bit less painful; creating the changelog entries will be a TODO when merging `issue-7306` into `v11`.

- `.invoke()` no longer waits for promises to resolve. If you wish to await a promise, use `cy.then()` instead.
- If `cy.getCookie()` fails to find a cookie, you can no longer traverse the non-existent cookie's properties (eg, with `.its('value)`). If you expect a cookie not to exist, assert that directly, eg: `cy.getCookie('foo1').should('not.exist')`

### DIFF
--- a/content/api/commands/invoke.md
+++ b/content/api/commands/invoke.md
@@ -143,53 +143,6 @@ const double = (n) => n * n
 cy.wrap([reverse, double]).invoke(1, 4).should('eq', 16)
 ```
 
-### Invoking an async function
-
-In this example we have a little text input field and we invoke an async action
-which will disable this input field. `.invoke()` will then wait until the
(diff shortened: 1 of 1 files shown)


chrisbreiding (approved):

flotwig (approved):

BlueWinds:
Folding this in with my main PR to the v12 branch. https://github.com/cypress-io/cypress-documentation/pull/4835/files

PR #4789, opened 2022-10-18
Update cy.origin dependencies / remove Cypress.require()

chrisbreiding:
For https://github.com/cypress-io/cypress/pull/24294


content/api/commands/origin.md
```diff
@@ -507,27 +507,37 @@ into the callback.
...
-[dynamic ES module `import()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports)
-within the callback. However, [`Cypress.require()`](/api/cypress-api/require)
-can be utilized to include [npm](https://www.npmjs.com/) packages and other
-files. It is functionally the same as using
-[CommonJS `require()`](https://nodejs.org/en/knowledge/getting-started/what-is-require/)
-in browser-targeted code.
+can be used within the callback to include [npm](https://www.npmjs.com/)
+packages and other files.
+
+If you utilize a custom preprocessor, note that this feature requires the latest
```
mschile, line 516: Suggest calling out the actual version, e.g. requires version 5.14.0 or greater

mjhenkes, line 516: Is there a good place to call out what the error would look like if you didn't use the latest preprocessor? or Webpack at all?

chrisbreiding: I'm more inclined to do that here than in the code, since it can be changed at will if something goes wrong with the package release. It should get released once the code PR is in develop, so I'll update this with the version then.

chrisbreiding: Added it right after this line. Addressed in https://github.com/cypress-io/cypress-documentation/pull/4789/commits/590ce263e36bf9c2a5a1014fcad73cb699ec33b6

mjhenkes (approved):

mschile (approved):

content/api/commands/origin.md
chrisbreiding: Addressed in https://github.com/cypress-io/cypress-documentation/pull/4789/commits/42354f2865a0abe95011ba798769dfff2879a8df

PR #4830, opened 2022-11-03
docs(styling): code and tabs style updates

elylucas:


src/css/custom.css
```diff
@@ -419,14 +439,53 @@ div.markdown h3 a:hover {
...
+  margin: 0;
+  padding: 0.75rem 1rem;
+  font-size: 1.05rem;
+}
+
+.tabs-container .tabs__item:hover {
+  background-color: var(--ifm-tabs-hover-color);
+}
+
+/* api tables */
```
jaffrepaul, line 481: where do these classes get applied? didn't see em anywhere.

src/css/custom.css
```diff
@@ -419,14 +439,53 @@ div.markdown h3 a:hover {
...
+.tabs-container .margin-top--md .theme-code-block {
+  margin-top: -1rem;
+}
+
+/* code title styling */
+div.codeBlockTitle_node_modules-\@docusaurus-theme-classic-lib-theme-CodeBlock-Content-styles-module {
+  font-size: 1.05rem;
+  font-weight: 700;
+  color: var(--ifm-color-gray-300);
+  border-bottom: 3px solid var(--ifm-code-background-color-light);
```
jaffrepaul, line 455: This selector won't hold up, it's targeting the theme you have in your node modules.  Notice they didn't get applied in the preview URL. I did this a lot in the beginning 😄 

Some dynamic info gets added to these after it builds so you should be fine to target this block like this. Also tweaked the border-bottom to be a little more subtle. 

```suggestion
div[class^='codeBlockTitle'] {
  font-size: 1.05rem;
  font-weight: 700;
  color: var(--ifm-color-gray-300);
  border-bottom: 2px solid var(--ifm-code-background-color-light);
}
```

src/css/custom.css
```diff
@@ -419,14 +439,53 @@ div.markdown h3 a:hover {
...
   background: var(--docusaurus-highlighted-code-line-bg);
 }
+
+.tabs-container {
+  background-color: var(--ifm-code-background-color);
+  border-radius: var(--ifm-code-border-radius);
+}
+
+.tabs-container .tabs {
+  box-shadow: inset 0px -6px 0px -3px var(--ifm-code-background-color-light);
```
jaffrepaul, line 468: Same here, just a little lighter on the stroke feels better me thinks. 
```suggestion
  box-shadow: inset 0px -6px 0px -4px var(--ifm-code-background-color-light);
```

src/css/custom.css
```diff
@@ -419,14 +439,53 @@ div.markdown h3 a:hover {
...
+}
```
jaffrepaul, line 469: I think it's good to have a subtle contrast between the tabs & the codeblock like we have in prod. 

```suggestion
.tabs-container .tabs {
  background-color: #1f2937;
  border-radius: 0.4rem 0.4rem 0 0;
  box-shadow: inset 0px -6px 0px -4px var(--ifm-code-background-color-light);
}
```

src/css/custom.css
```diff
@@ -419,14 +439,53 @@ div.markdown h3 a:hover {
...
+.tabs-container .tabs__item {
+  margin: 0;
+  padding: 0.75rem 1rem;
+  font-size: 1.05rem;
+}
+
+.tabs-container .tabs__item:hover {
+  background-color: var(--ifm-tabs-hover-color);
+}
+
```
jaffrepaul, line 480: One other thing I think is a nice touch is hiding the title if there is one with tabs. Feels pretty clunky like this

Screen Shot 2022-11-03 at 9 41 49 AM

```suggestion
.tabs-container div[class^='codeBlockTitle'] {
  display: none;
}
```

jaffrepaul (changes requested):
This is a nice touch. Good stuff. I added a few subtle tweaks that I think really tie a bow around it. Let me know what you think.

src/css/custom.css
elylucas: they are in the new CT docs

src/css/custom.css
elylucas: I agree its clunky but I don't want the title to disappear, its important the user knows which file the code block belongs to.

src/css/custom.css
jaffrepaul: 👍

src/css/custom.css
jaffrepaul: Is there an option for the tabs to be the file name like we currently have? Or is it subjective based on the content that something like the tabs in the above image should have that text?

elylucas: Not sure what you mean. Nothings changed here but the style. It looks more prominent now because the background color of the tab container is the same as the code block.

jaffrepaul: Sorry, to clarify, I meant like we currently have with the file name as the tab text on production. Is it deliberate to name the tabs something different and if so does warrant having the tabs - filename - codeblock sandwich? 

Screen Shot 2022-11-03 at 5 13 51 PM

elylucas: The reason the filename isn't in the tab in production is because it had to go there as the old docs don't support title attributes on code blocks. I'm not sure what code example this is and not sure if it still exists in the new CT docs. We can probably reorganize how the tabs and titles are done, but I would consider that out of scope for this css update.

jaffrepaul (approved):
LGTM. 

Let's see where we land after pulling in the latest CT content and see if it's worth updating the tab/title situation.

PR #4835, opened 2022-11-03
First rework of retryability guide

BlueWinds:
For reviewing this, I suggest *reading the new version in its entirety first,* and commenting on that. Then go back and look at the diff from the old one, to see if there's anything that was left out. Comparing them section-by-section is probably not all that fruitful.


content/api/commands/clearcookies.md
```diff
@@ -40,8 +40,7 @@ Pass in an options object to change the default behavior of `cy.clearCookies()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`cy.clearCookies()` yields `null`.</li><li>`cy.clearCookies()` cannot
-be chained further.</li></List>
```
emilyrohrbough, line 44: I think it's valuable to callout it cannot be chained further.

content/api/commands/clear.md
```diff
@@ -52,8 +52,9 @@ Pass in an options object to change the default behavior of `.clear()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`.clear()` yields the same subject it was given from the previous
-command.</li></List>
+- `.clear()` yields the same subject it was given.
```
emilyrohrbough, line 55: why drop `from the previous command.` from all of the API docs?
```suggestion
- `.clear()` yields the same subject it was given from the previous command.
```

content/api/commands/children.md
```diff
@@ -52,7 +52,8 @@ Pass in an options object to change the default behavior of `.children()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`.children()` yields the new DOM element(s) it found.</li></List>
+- `.children()` yields the new DOM element(s) it found.
+- `.children()` is a query, and it is _safe_ to chain further methods.
```
emilyrohrbough, line 56:
```suggestion
- `.children()` is a query, and it is _safe_ to chain further commands.
```

emilyrohrbough: Should these all be consistent between the docs?

content/api/commands/document.md
```diff
@@ -32,7 +32,8 @@ Pass in an options object to change the default behavior of `cy.document()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`cy.document()` 'yields the `window.document` object' </li></List>
+- `cy.document()` 'yields the `window.document` object.
+- `cy.document()` is a query, and it is _safe_ to chain further methods.
```
emilyrohrbough, line 36: @BlueWinds Should all Query commands be moved to live under APIs/queries to help alleviate confusion when a user is looking at the docs api face up?

emilyrohrbough: It might also provide another (or better) face up opportunity to describe the difference between query & command and how to add custom and the retry-ability differences are (at this point only through the upper docs so you might have done this...)

content/api/commands/end.md
```diff
@@ -26,8 +26,7 @@ cy.end() // Does not make sense to chain off 'cy'
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`.end()` yields `null`.</li><li>`.end()` cannot be chained
-further.</li></List>
+- `.end()` yields `null`.
```
emilyrohrbough, line 29: Is end a query or a command?

content/api/commands/clearlocalstorage.md
```diff
@@ -46,8 +46,7 @@ Pass in an options object to change the default behavior of
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`cy.clearLocalStorage()` yields
-`null`.</li><li>`cy.clearLocalStorage()` cannot be chained further.</li></List>
+- `cy.clearLocalStorage()` yields `null`.
```
emilyrohrbough, line 49: Is end a query or a command?

content/api/commands/clearcookies.md
```diff
@@ -40,8 +40,7 @@ Pass in an options object to change the default behavior of `cy.clearCookies()`.
...
+- `cy.clearCookies()` yields `null`.
```
emilyrohrbough, line 43: Is clearCookies a query or a command?

content/api/commands/clearcookie.md
```diff
@@ -44,8 +44,7 @@ Pass in an options object to change the default behavior of `cy.clearCookie()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`cy.clearCookie()` yields `null`.</li><li>`cy.clearCookie()` cannot be
-chained further.</li></List>
+- `cy.clearCookie()` yields `null`.
```
emilyrohrbough, line 47: Is clearCookie a query or a command?

content/api/commands/and.md
```diff
@@ -68,8 +73,8 @@ Whatever was passed to the function is what is yielded.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>In most cases, `.and()` yields the same subject it was given from the
-previous command.</li></List>
+- In most cases, `.and()` yields the same subject it was given.
+- `.and()` is an assertion, and it is _safe_ to chain further methods.
```
emilyrohrbough, line 77: Are assertions commands or queries?

content/api/commands/getcookies.md
```diff
@@ -44,6 +44,9 @@ following properties:
 - `secure`
 - `value`
 
+`cy.getCookies()` is a command, not a query. It will not update the returned
+list if further cookies are added after it initially executes.
```
emilyrohrbough, line 48:
```suggestion
list, if further cookies are added, after it initially executes.
```

content/api/commands/go.md
```diff
@@ -41,9 +41,9 @@ Pass in an options object to change the default behavior of `cy.go()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`cy.go()` 'yields the `window` object after the page finishes loading'
-
-</li></List>
+- `cy.go()` 'yields the `window` object after the page finishes loading.
```
emilyrohrbough, line 44:
```suggestion
- `cy.go()` yields the `window` object after the page finishes loading.
```

content/api/commands/intercept.md
```diff
@@ -154,8 +154,7 @@ See ["Intercepted requests"][req] and
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-- `cy.intercept()` yields `null`.
-- `cy.intercept()` can be aliased, but otherwise cannot be chained further.
+- `cy.intercept()` yields `null`, but can be aliased.
 - Waiting on an aliased `cy.intercept()` route using
```
emilyrohrbough, line 158: query or command?

content/api/commands/invoke.md
```diff
@@ -26,15 +35,19 @@ subject, use [`.its()`](/api/commands/its).
 **<Icon name="check-circle" color="green"></Icon> Correct Usage**
 
 ```javascript
-cy.wrap({ animate: fn }).invoke('animate') // Invoke the 'animate' function
+cy.get('.input').invoke('val').should('eq', 'foo') // Invoke the 'val' function
```
emilyrohrbough, line 38: given the alert above, wouldn't `invoke('val)` run 1-many times until should is satisfied or it times out? Is this the expected usage of invoke?

content/api/commands/invoke.md
```diff
@@ -26,15 +35,19 @@ subject, use [`.its()`](/api/commands/its).
...
 ```
 
 **<Icon name="exclamation-triangle" color="red"></Icon> Incorrect Usage**
 
 ```javascript
 cy.invoke('convert') // Errors, cannot be chained off 'cy'
 cy.wrap({ name: 'Jane' }).invoke('name') // Errors, 'name' is not a function
+cy.wrap({ animate: fn })
+  .invoke('animate')
+  .then(() => {}) // 'animate' will be called multiple times
```
emilyrohrbough, line 50: .then() is one of the few commands that doesn't actually doesn't retry (unless that has changed) so we might want to update this example

content/api/commands/clearcookies.md
BlueWinds: We chain off commands that return `null` all the time in our own tests.

```
it.only('works', () => {
  cy.visit('/fixtures/dom.html')
    .end()
    .get('button')
    .end()
    .clearCookies('foo')
    .get('form')
})
```

It's always worked this way, our docs were just plain wrong. :woman_shrugging:

content/api/commands/clear.md
BlueWinds: "From the previous command, query or assertion" is pretty verbose, and I don't think it actually adds any clarity. Assertions usually don't change the subject, but they can! So even before queries, this wasn't quite accurate.

content/api/commands/children.md
BlueWinds: The "unsafe" verbage has further qualification because it's unsafe in *some cases*, and I wanted to call out what those cases are (queries or commands that rely on the subject).

Here though, it's just always safe. You can chain any method (query, command or assertion) off of a query, whether or not it relies on the subject. I don't think "it's always safe" needs further qualification.

content/api/commands/document.md
BlueWinds: Do you mean changing the URL, so that it's `/apis/query/document` rather than `/api/command/document`? I'm definitely open to the idea - we'd need to leave in place redirects so that we don't break links, but that's not hard.

I'm not sure I follow the second comment though. Can you explain what you mean again?

content/api/commands/end.md
BlueWinds: Good catch, it's a query. Moving it to the proper section of the sidebar/TOC.

content/api/commands/clearlocalstorage.md
BlueWinds: It's a command.

It yields null though, I don't think it needs to be called out explicitly that you can chain things off of it.

content/api/commands/then.md
```diff
@@ -51,20 +51,23 @@ Pass a function that takes the previously yielded subject as its first argument.
...
-`return`.
-
-When `undefined` is returned by the callback function, the subject will not be
-modified and will instead carry over to the next command.
-
-Just like Promises, you can return any compatible "thenable" (anything that has
-a `.then()` interface) and Cypress will wait for that to resolve before
-continuing forward through the chain of commands.
+flow into the next command (with the exception of `undefined` or `null`). If the
+return value is a Promise or other "thenable" (anything with a `.then()`
```
emilyrohrbough, line 56:
```suggestion
return value is a Promise or a Chained `.then()` command
```
doesn't it way for any returned cypress command?

content/api/commands/clearcookies.md
BlueWinds: Command. Also not sure it needs to be called out in the 'Yields' header though - it yields null, that's kind of the end of it. :woman_shrugging:

content/api/commands/and.md
BlueWinds: Neither - they're assertions. They've always been a special case, and continue to be so.

content/api/commands/then.md
```diff
@@ -51,20 +51,23 @@ Pass a function that takes the previously yielded subject as its first argument.
...
+interface), Cypress will wait for it to resolve before continuing forward
+through the chain of commands.
+
+If `undefined` or `null` are returned (or there is no `return`), the result of
```
emilyrohrbough, line 60:
```suggestion
If `undefined` or `null` are returned (or there is no return value), the result of
```

content/api/commands/getcookies.md
BlueWinds: I don't think this comma is correct.

content/api/commands/go.md
BlueWinds: :+1:

content/api/commands/then.md
```diff
@@ -51,20 +51,23 @@ Pass a function that takes the previously yielded subject as its first argument.
...
+the last Cypress command in the callback function will be yielded as the new
+subject instead, and flow into the next command.
+
+If `undefined` or `null` are returned (or there is no `return`) and the callback
+does not call any Cypress commands, the subject will not be modified and the
+previous subject will carry over to the next command.
+
+`.then()` is a command, and will not be retried. It is **unsafe** to return DOM
+elements from the callback and then use further assertions, queries of commands
```
emilyrohrbough, line 69:
```suggestion
elements from the callback and then use further assertions, queries or commands
```

emilyrohrbough (commented):
I'm partially through. Will need to continue later.

content/api/commands/wait.md
```diff
@@ -64,8 +64,9 @@ Pass in an options object to change the default behavior of `cy.wait()`.
 
 #### When given a `time` argument:
 
-<List><li>`cy.wait()` yields the same subject it was given from the previous
-command.</li></List>
+- `cy.wait()` yields the same subject it was given.
+- `.screenshot()` is a command, and it is **unsafe** to chain further methods
```
emilyrohrbough, line 68:
```suggestion
- `cy.wait()` is a command, and it is **unsafe** to chain further methods
```

content/api/commands/invoke.md
BlueWinds: Yes, this is the expected usage of invoke, and one of the most common ways people use it (as measured via querying metabase test bodies).

"Get an input, invoke the `val` method on the jquery object, assert that the value matches a string" is :+1:

content/api/commands/invoke.md
BlueWinds: I actually switched to `.then()` exactly for this reason - even if the following command doesn't retry and doesn't care about the subject, `.invoke()` will *still* get called twice - once when executing it, and then a second time when determining the subject for `.then()`.

content/api/commands/reload.md
```diff
@@ -38,8 +38,9 @@ reload without cache.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`cy.reload()` 'yields the `window` object after the page finishes
-loading' </li></List>
+- `cy.reload()` 'yields the `window` object after the page finishes loading.
```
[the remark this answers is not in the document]
BlueWinds: :+1: 

Also fixed the line right below, where I'd copy-pasted .go() rather than .reload().

content/api/commands/root.md
```diff
@@ -41,6 +41,8 @@ The root element yielded is `<html>` by default. However, when calling `.root()`
 from a [`.within()`](/api/commands/within) command, the root element will point
 to the element you are "within".
 
+`.root()` is a query, and it is _safe_ to chain further methods.
```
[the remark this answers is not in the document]
BlueWinds: :+1:

content/api/commands/scrollintoview.md
```diff
@@ -42,8 +42,9 @@ Pass in an options object to change the default behavior of `.scrollIntoView()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`.scrollIntoView()` yields the same subject it was given from the
-previous command.</li></List>
+- `.scrollIntoView()` yields the same subject it was given.
+- `.scrollIntoView()` is a command, and it is **unsafe** to chain further
```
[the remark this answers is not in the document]
BlueWinds: Yep. This was the source of some of our test flake - when list virtualization is involved, scrolling causes a great many DOM rerenders and updates, and the original element you wanted to scroll into vue (view :grin: ) may be replaced.

content/api/commands/spy.md
```diff
@@ -38,11 +38,12 @@ The name of the `method` on the `object` to be wrapped.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-Unlike most Cypress commands, `cy.spy()` is _synchronous_ and returns a value
-(the spy) instead of a Promise-like chain-able object.
-
-`cy.spy()` returns a [Sinon.js spy](https://sinonjs.org/releases/v6.1.5/spies/).
-All methods found on Sinon.JS spies are supported.
+- `cy.spy()` is a _utility function_, and is neither a command nor a query.
```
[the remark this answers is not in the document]
BlueWinds: `.spy()` and `.stub()` are super weird cases - they really belong under the `Utilities` header and slug, and should not be using the "command" template at all, pretending things like 'assertions', 'timeouts' and 'yields' apply to them.

They don't retry - they don't even enter the command queue. They cannot be chained at all. `cy.spy().as()` doesn't even use the `.as()` command - it uses a custom function, and doesn't return a $Chainer.

I'll try and update the docs to be clearer about this.

content/api/commands/then.md
BlueWinds: That's not actually what the docs are trying to say - it means literally "if the return value is a promise or something like `return { then: () => {}}`."

With that said, I do think it is extremely unclear, and am updating the wording to be more straightforward.

content/api/commands/then.md
BlueWinds: :+1:

content/api/commands/then.md
BlueWinds: :+1:

content/api/commands/wait.md
BlueWinds: :+1:

lmiller1990 (approved):
These look good. The most interesting file is hidden by GH by default: https://github.com/cypress-io/cypress-documentation/pull/4835/files#diff-4d501082dd1739f23f2ed58152f22a6747b15cdc3b77f830a86226c8f3b5ac55

I hope the distinction between queries/commands will help make things more clear to users. It kind of already existed implicitly, now it's more clear and defined, which is great.

content/api/commands/document.md
BlueWinds: Changing the slug wouldn't necessarily remove them from the same page - you'd still see them in the TOC.

Screenshot_20221109_073631

/api/commands/..., /api/utilities/... and /api/plugins/... are already there.

@elylucas - if I wanted to move `/api/commands/as` to `/api/queries/as`, and have `/api/commands/as` redirect to the new URL, how would I go about this?

lmiller1990: Right, gotcha.

elylucas (changes requested):

content/api/commands/clearcookies.md
BlueWinds: Showed an example in a comment here - https://github.com/cypress-io/cypress-documentation/pull/4835#discussion_r1017145845

The problem is that "parent" and "child" commands do *not* work the way you expect them to from the name.

- Parent commands are those that *don't care about their subject* - being a parent does not limit where in a chain a command can be!
- Dual commands accept both null and a valid subject. They behave differently based on what subject they're given.
- Child commands require a subject. But this is a rule about the subject they're passed, not about their position. `cy.end().click()` is invalid because the subject is null, not because `.end()` can't be chained off of.

This is all existing behavior from Cy1-11, covered by our own tests - the docs have just never explained it clearly.

BlueWinds:
```
  cy.visit('/fixtures/dom.html')
    .clearCookies('foo')
    .clearCookies('foo')
    .clearCookies('foo')
    .get('form')
    .get('form')
    .get('form')
```

Seven parent commands chained together. Our own tests do this all the time, as do our example repos. Forcing parents to actually be parents or disallowing chaining after a command that returns `null` would be a massively breaking change.

content/api/commands/as.md
```diff
@@ -46,8 +46,8 @@ using an `@` prefix.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`.as()` yields the same subject it was given from the previous
-command.</li></List>
+- `.as()` yields the same subject it was given.
+- `.as()` is a query, and it _safe_ to chain further methods.
```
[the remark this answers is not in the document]
BlueWinds: It's a query because it's implemented with `Commands.addQuery()`, not `Commands.add()`. It's implemented this way so that it's safe to chain off of.

```
cy.get('button').as('btn').click()
```

is something we want to support, so `.as()` is a query. Basically, if something *can* be a query, it *should* be, to give our users maximum safety and freedom to string together methods as they please.

I think that's the basic intuition I'm having trouble conveying: "Queryness" is about *how methods chain together*, not about the DOM. Very open to a better name, if you have any suggestions.

content/api/commands/spread.md
```diff
@@ -49,10 +49,10 @@ Pass in an options object to change the default behavior of `.spread()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`.spread()` 'yields the return value of your callback function'
-
-</li><li>`.spread()` wll not change the subject if `null` or `undefined` is
-returned.</li></List>
+- `.spread()` yields the return value of your callback function
```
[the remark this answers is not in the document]
BlueWinds: :+1:

content/api/commands/as.md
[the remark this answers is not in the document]
elylucas: there seems to be a number of "utility" functions that fall into scenario. I think calling them queries is confusing. Maybe we can refer to them as utilities (also open to better naming), and call out they are chainable?

content/api/commands/invoke.md
```diff
@@ -11,6 +11,15 @@ subject, use [`.its()`](/api/commands/its).
 
 </Alert>
 
+<Alert type="warn">
+
+If you chain further methods off of `.invoke()`, it will be called multiple
+times. If the method is not idempotent, treat `.invoke()` like a command, and
```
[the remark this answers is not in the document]
BlueWinds: This isn't to avoid detached DOM errors - it's to avoid calling the method more than once. Updating the line to be more clear:

```
If you chain further methods off of `.invoke()`, it will be called multiple
times. If you only want the method to be called once, treat `.invoke()` like a
command, and consider it unsafe to chain further methods.
```

content/api/commands/spy.md
```diff
@@ -38,11 +42,13 @@ The name of the `method` on the `object` to be wrapped.
...
+  Cypress concepts like retries and timeouts do not apply.
+- `cy.spy()` is _synchronous_ and returns a value (the spy) instead of a
+  Promise-like chain-able object. It can be aliased.
+- `cy.spy()` returns a
+  [Sinon.js spy](https://sinonjs.org/releases/v6.1.5/spies/). All methods found
```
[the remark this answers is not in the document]
BlueWinds: I think this would be less useful  than the current link - that section basically just points users straight back to this page, without any details about the `spy` API.

content/api/commands/spy.md
```diff
@@ -80,8 +86,8 @@ const stub = cy.stub(obj, 'foo').log(false)
 
 ### Aliases
 
-Adding an alias using [`.as()`](/api/commands/as) to spies makes them easier to
-identify in error messages and Cypress' command log.
+You can alias spies, similar to how [`.as()`](/api/commands/as) works. This can
+make your spies easier to identify in error messages and Cypress's command log.
```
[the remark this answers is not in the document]
BlueWinds: Updated this line, and also the example below to show that usage.

content/api/commands/document.md
BlueWinds: Wouldn't onlinks continue to work, if we add a redirect inside the docs? Eg, `on.cypress.io/get` -> `docs.cypress.io/api/commands/get` -> `docs.cypress.io/api/queries/get`?

content/api/commands/as.md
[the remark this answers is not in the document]
BlueWinds: Utility is heavily overloaded already - we have 'utility commands', and a 'Utilities' section in the API docs.

How about just removing the word "query" here?

```
- It is _safe_ to chain further methods after `.as()`.
```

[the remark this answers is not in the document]
BlueWinds: The three sticking points for the query name seem to be:
- .as()
- .invoke()
- .its()

I'd argue that .invoke() and .its() fit very well.

"I to know this button's title" -> cy.get('button').invoke('attr', 'title')
"I want to know the value of this form" -> cy.get('form').invoke('val')
"I want to know the request body" -> cy.get('@myRequest.1').its('req.body')
"I want to know the user's name" -> cy.wrap(userObj).its('name.first')

All of these are very naturally thought of as "queries". In SQL, you query the DOM. In CSS, you query the DOM. In Cypress, you query your application - Cypress allows subjects other than DOM elements, so you should be able to query things other than DOM elements.

content/api/commands/clearcookies.md
chrisbreiding: The reason we use the term "yields" and not "returns" is because it's stating which subject is passed to the next command, not what is returned from the command. All commands return a chainer object that allows chaining further commands. 

While we do indeed use a lot of unnecessary chaining in our tests, there's at least one legitimate use for chaining after a command that yields `null`:

```js
cy.clearCookies().then(() => {
  // do something after clearing the cookies has completed
})
```

content/api/cypress-api/custom-commands.md
```diff
@@ -411,6 +416,15 @@ You can also modify the behavior of existing Cypress commands. This is useful to
...
 
+<Alert type="warning">
+
+`Cypress.Commands.overwrite` can only overwrite commands, not queries. If you
+want to modify the behavior of a query, you'll need to use
+[`Cypress.Commands.overwriteQuery`](/api/cypress-api/custom-queries#overwriting-existing-queries)
+instead.
+
+</Alert>
+
```
chrisbreiding, line 427:
```suggestion
```

Not sure if this should be completely removed or if maybe there's some other advice we can offer if they're trying to overwrite a query?

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,378 @@
...
+since it is loaded before any test files are evaluated via an import statement
+in the
+[supportFile](/guides/core-concepts/writing-and-organizing-tests#Support-file).
+
+</Alert>
+
+Queries follow three main rules.
+
+1. Queries are _synchronous._ They do not return or await promises.
+2. Queries are _retrieable._ Once you return the inner function, Cypress takes
```
chrisbreiding, line 28:
```suggestion
2. Queries are _retriable._ Once you return the inner function, Cypress takes
```

chrisbreiding (changes requested):

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,378 @@
...
+### Using existing queries
+
+Many custom queries are wrappers around other, already implemented queries -
+most commonly `.get()`.
+
+This is quite simple:
+
+```javascript
+Cypress.Commands.addQuery('getFirstButton', function getFirstButton(options) {
+  const getFn = cy.now('get', 'button:first', options)
```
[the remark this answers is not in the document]
elylucas: I see its explained a bit below but I still don't quite understand it. It also causes errors in TS as theres no definition for `cy.now()` (unless that is new in your branch)

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,378 @@
...
+## History
+
+| Version                                       | Changes              |
+| --------------------------------------------- | -------------------- |
+| [12.0.0](/guides/references/changelog#12-0-0) | `addQuery` API added |
+
+## See also
+
+- See how to add
+  [TypeScript support for custom commands](/guides/tooling/typescript-support#Types-for-custom-commands)
```
elylucas, line 376: is there anything different that needs to be done to set up types for queries vs commands?

content/guides/component-testing/slots-vue.md
```diff
@@ -39,21 +39,23 @@ previous sections, we'll start simple.
 ```js
 it('renders the modal content', () => {
   cy.mount(Modal, { slots: { default: () => 'Content' } })
-    .get(modalSelector)
-    .should('have.text', 'Content')
+
+  cy.get(modalSelector).should('have.text', 'Content')
```
elylucas, line 43: All of these component testing pages have gone away, and most of these examples (if they still exist at all) are now either in the quickstart or example guides for each framework.

elylucas (changes requested):

content/api/commands/document.md
elylucas: > Wouldn't onlinks continue to work, if we add a redirect inside the docs? Eg, `on.cypress.io/get` -> `docs.cypress.io/api/commands/get` -> `docs.cypress.io/api/queries/get`?

They would, but good to update to keep the double redirect from happening and to be correct in the services repo.

content/guides/component-testing/slots-vue.md
BlueWinds: Thanks for the heads up. Merged in latest v12 and fixed the merge conflicts.

content/api/commands/as.md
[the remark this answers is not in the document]
BlueWinds: We discussed in a meeting an decided that we'd change things around a bit into a hierarchy:

Commands:
- Queries
- Actions
- Assertions
- ...other commands...

I'm updating the PR to reflect these changes.

content/api/commands/children.md
BlueWinds: I've rewritten these "safe/unsafe" sections, so going to resolve all these comments and we can start fresh threads if there are still more changes we want to see.

content/api/commands/document.md
BlueWinds: I've re-organized the table of contents following discussions with Brian and Emily. We now go:

Commands
- Assertions
- Actions
- Queries

and then all the other commands that aren't one of these three beneath "Commands". I've added links to relevant guides to each section.

With this hierarchy change, I'm not sure it makes sense to update the slugs any more - they're all still "Commands", some of them just also have additional modifiers ("query command", "action command").

content/api/commands/end.md
BlueWinds: https://github.com/cypress-io/cypress/pull/24681

content/api/cypress-api/custom-commands.md
```diff
@@ -8,6 +8,15 @@ defined below.
...
+If you want your method to have builtin
+[retry-ability](/guides/core-concepts/retrysability), and especially if you
+return a DOM element for further commands to act on, consider writing a
+[custom query](/api/cypress-api/custom-queries) instead.
+
+</Alert>
+
+<Alert type="info">
+
 A great place to define or overwrite commands is in your
```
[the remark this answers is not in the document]
BlueWinds: :+1:

content/api/cypress-api/custom-commands.md
BlueWinds: Updated to offer new advice about how to customize queries.

```
`Cypress.Commands.overwrite` can only overwrite commands, not queries. If you
want to modify the behavior of a query, you'll need to add your updated version
under a new name using
[`Cypress.Commands.addQuery`](/api/cypress-api/custom-queries) instead.
```

content/api/cypress-api/custom-commands.md
```diff
@@ -644,18 +650,17 @@ You can also mix optional commands **with** validations.
 ```javascript
 // this is how .contains() is implemented
```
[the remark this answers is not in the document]
BlueWinds: :+1:

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,378 @@
+---
+title: Custom Queries
+---
+
+Cypress comes with its own API for creating custom queries. The built in Cypress
```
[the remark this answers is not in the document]
BlueWinds: Good idea, added.

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,378 @@
...
+
+If your method needs to be asynchronous, or it shouldn't be called more than
+once, consider writing a [custom command](/api/cypress-api/custom-command)
+instead.
+
+</Alert>
+
+<Alert type="info">
+
+A great place to define queries is in your `cypress/support/commands.js` file,
```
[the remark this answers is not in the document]
BlueWinds: :+1:

content/api/cypress-api/custom-queries.md
BlueWinds: :+1:

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,378 @@
...
+
+</Alert>
+
+Queries follow three main rules.
+
+1. Queries are _synchronous._ They do not return or await promises.
+2. Queries are _retrieable._ Once you return the inner function, Cypress takes
+   control, handling retries on your behalf.
+3. Queries are _idempotent._ Once you return the inner function, Cypress will
+   invoke it repeatedly.
```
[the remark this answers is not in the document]
BlueWinds: Went with `Invoking the inner function multiple times must not change the state of your application.`, to make it clear that this isn't just about the DOM, it's about the whole application. Reseeding the database or resetting cookies are also not queries.

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,378 @@
...
+**<Icon name="angle-right"></Icon> name** **_(String)_**
+
+The name of the query you're adding.
+
+**<Icon name="angle-right"></Icon> callbackFn** **_(Function)_**
+
+Pass a function that receives the arguments passed to the query.
+
+This outer function is invoked once. It should return a function that takes a
+subject and returns a new subject; this inner function will be called multple
```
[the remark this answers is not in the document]
BlueWinds: :+1:

content/api/cypress-api/custom-queries.md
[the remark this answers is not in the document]
BlueWinds: Hm. We do this several times in our own code - I'm not sure why it's a TS error. `cy.now()` has always existed. Will look into the TS error in a moment here.

Added another paragraph trying to explain this further.

content/api/cypress-api/custom-queries.md
BlueWinds: Nope, they are the same from a TypeScript perspective.

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,385 @@
...
+
+  return () => {} // Inner function
+}
+```
+
+Piece by piece, then.
+
+```javascript
+function focused(options = {}) {}
+```
```
emilyrohrbough, line 129: Not sure where this is tied in

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,385 @@
...
+
+```javascript
+if (options === null || !_.isPlainObject(options)) {
+  const err = `cy.root() requires an \`options\` object. You passed in: \`{options}\``
+  throw new TypeError(err)
+}
+```
+
+This is a general pattern: when something goes wrong, queries just throw an
+error. Cypress will handle displaying this in the Command Log nicely.
```
emilyrohrbough, line 146:
```suggestion
error. Cypress will handle displaying the error in the Command Log.
```

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,385 @@
...
+
+```javascript
+const log = options.log !== false && Cypress.log({ timeout: options.timeout })
+```
+
+If the user has not set `{ log: false }`, we create a new `Cypress.log()`
+instance. See [`Cypress.log()`](/api/cypress-api/cypress-log) for more
+information.
+
+This is setup code - we only want it to run once, creating the log message when
```
emilyrohrbough, line 156: can we call this outer function or change the above outer func reference to setup code?

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,385 @@
...
+Cypress first begins executing this query. We hold onto a reference to `Log`
```
emilyrohrbough, line 157: Does this mean these log references always exist until a test has completed due to the command chain re-executing the inner function?

emilyrohrbough (changes requested):

content/api/commands/focus.md
```diff
@@ -4,6 +4,9 @@ title: focus
 
 Focus on a DOM element.
 
+It is [unsafe](/guides/retry-ability#Only-Queries-are-retried) to chain further
+commands that rely on the subject after `.focus()`.
```
emilyrohrbough, line 8: This contradicts our current recommendation for .blur() by saying this is unsafe to chain further...

> This element must currently be in focus. If you want to ensure an element is
focused before blurring, try using [`.focus()`](/api/commands/focus) before
`.blur()`.

content/api/commands/blur.md
```diff
@@ -4,6 +4,9 @@ title: blur
 
 Blur a focused element.
 
+It is [unsafe](/guides/retry-ability#Only-Queries-are-retried) to chain further
```
emilyrohrbough, line 7:
```suggestion
It is [unsafe](/guides/core-concepts/retry-ability#Only-Queries-are-retried) to chain further
```

content/api/commands/viewport.md
```diff
@@ -80,8 +80,7 @@ Pass in an options object to change the default behavior of `cy.viewport()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`cy.viewport()` yields `null`.</li><li>`cy.viewport()` cannot be
-chained further.</li></List>
+- `cy.viewport()` yields `null`.
```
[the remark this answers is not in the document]
BlueWinds: `cy.viewport()` is used to set the viewport size, not query it. It's closer to `cy.visit()` than it is `cy.get()`.

I don't think it makes sense to say "viewport is not an action command", and for much the same reason I don't think it makes sense to call out "viewport is not a query command."

content/api/commands/wrap.md
```diff
@@ -37,7 +37,7 @@ Pass in an options object to change the default behavior of `cy.wrap()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`cy.wrap()` 'yields the object it was called with' </li></List>
+- `cy.wrap()` yields the object it was called with.
```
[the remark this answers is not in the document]
BlueWinds: The unsafe warning is meant to be "here's how to avoid a common pitfall, detached DOM errors", and I added it to commands that yield DOM elements.

It could potentially be unsafe to do something like `cy.wrap(cy.$$('button'))`, but if people are doing that, then they need to go re-read "Introduction to Cypress", not a specific warning on the `cy.wrap()` "yields" section.

content/api/commands/blur.md
```diff
@@ -49,8 +52,9 @@ Pass in an options object to change the default behavior of `.blur`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`.blur()` yields the same subject it was given from the previous
-command.</li></List>
+- `.blur()` yields the same subject it was given.
```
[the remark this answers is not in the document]
BlueWinds: https://docs.cypress.io/api/commands/blur#Actionability

> #### Blur is not an action command
> .blur() is not implemented like other action commands, and does not follow the same rules of [waiting for actionability](https://docs.cypress.io/guides/core-concepts/interacting-with-elements).

> .blur() is a helpful command used as a shortcut. Normally there's no way for a user to "blur" an element. Typically the user would have to perform another action like clicking or tabbing to a different element. Needing to perform a separate action like this is very indirect.
> Therefore it's often much more efficient to test the blur behavior directly with .blur().

content/api/commands/scrollintoview.md
```diff
@@ -42,8 +45,9 @@ Pass in an options object to change the default behavior of `.scrollIntoView()`.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`.scrollIntoView()` yields the same subject it was given from the
-previous command.</li></List>
+- `.scrollIntoView()` yields the same subject it was given.
```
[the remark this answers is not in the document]
BlueWinds: Nope. We don't verify that elements are actionable before we scroll to them. You can scroll to elements that have 0 height/width or to disabled form elements, for example.

https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Actionability is the complete list of action commands.

content/api/commands/log.md
```diff
@@ -33,8 +33,7 @@ to the number of arguments.
 
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
-<List><li>`cy.log()` yields `null`.</li><li>`cy.log()` cannot be chained
-further.</li></List>
+- `cy.log()` yields `null`.
```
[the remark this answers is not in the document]
BlueWinds: Hm, this is a really good question. `cy.log` is "cy.wrap(), but it yields null and has a different name." This keeps confusing me - I always expect it to log the subject, rather than log its arguments, but that's not what it's used for.

Thinking about it a bit more, .log() should probably be a command in the code rather than a query. I'll open a PR to change it back to a command.

content/guides/references/trade-offs.md
```diff
@@ -37,7 +37,6 @@ Many of these issues are currently being worked on or are on our
 - [Workarounds for the lack of a `cy.hover()` command.](/api/commands/hover)
 - [`cy.tab()` command.](https://github.com/cypress-io/cypress/issues/299)
 - [There is not any native or mobile events support.](https://github.com/cypress-io/cypress/issues/311#issuecomment-339824191)
-- [Testing file uploads is application specific.](https://github.com/cypress-io/cypress/issues/170#issuecomment-340012621)
```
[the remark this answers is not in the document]
BlueWinds: This should have been removed back when I released `selectFile`, back late last year, but I didn't notice it at that point. 
We resolved the issue a year ago, just forgot to clean up a link in the docs. :woman_shrugging: 

It can be its own PR, but I hate the overhead of creating a branch, opening a PR with description and getting multiple reviewers for a one-line cleanup.

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,385 @@
+---
+title: Custom Queries
+---
+
+Starting in Cypress 12, Cypress comes with its own API for creating custom
+queries. The built in Cypress queries use the very same API that's explained
```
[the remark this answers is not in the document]
BlueWinds: Updated this top section to make it clearer. Also added more hints pointing people towards custom commands instead of queries, since that's what 95% of people will want.

> Queries are a type of command, used for *querying* the state of your
application. They are different from other commands in that they follow three
important rules:
> 1. Queries are _synchronous._ They do not return or await promises.
> 2. Queries are _retriable._ Once you return the inner function, Cypress takes
   control, handling retries on your behalf.
> 3. Queries are _idempotent._ Once you return the inner function, Cypress will
   invoke it repeatedly. Invoking the inner function multiple times must not
   change the state of your application.

> ### \<Info box>
> If you want to chain together existing Cypress commands as a shortcut, you
probably want to write a [custom command](/api/cypress-api/custom-command)
instead.

> You'll also want to write a command instead of a query if your method needs to
be asynchronous, or it shouldn't be called more than once.

content/api/cypress-api/custom-queries.md
BlueWinds: I've added "..." to indicate that I'm omitting the function body. 

But this is just "starting at the top." It's the first line of the outer function. I wanted to talk about it, so I'm repeating it, same as I do for each other line.

content/api/cypress-api/custom-queries.md
BlueWinds: Updated.

> This line is setup code, so it lives in the outer function  - we only want it
to run once, creating the log message when Cypress first begins executing this
query. We hold onto a reference to `Log` instance. We'll update it later with
additional details when the inner function executes.

content/api/cypress-api/custom-queries.md
BlueWinds: I'm not entirely sure what you're getting at? Yes, the closure holds onto a reference to the $Log object, so as long as the closure is referenced, $Log can't be GCed.

Existing commands do the exact same thing, though they mostly store their reference as $Command.attributes._log, rather than in a closure. Changing the place we hold onto a reference (as a closure local variable rather than as an attribute on an object) shouldn't have any memory-usage implications.

content/api/cypress-api/custom-queries.md
```diff
@@ -0,0 +1,385 @@
...
+- `cy.ensureWindow(subject)`: Ensure that the passed in `subject` is a
+  `document`.
+- `cy.ensureDocument(subject)`: Ensure that the passed in `subject` is a
+  `window`.
+
+- `cy.ensureAttached(subject, queryName)`: Ensure that DOM element(s) are
+  attached to the page.
+- `cy.ensureNotDisabled(subject)`: Ensure that form elements aren't disabled.
+- `cy.ensureVisibility(subject)`: Ensure that a DOM element is visible on the
+  page.
```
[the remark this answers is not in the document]
BlueWinds: Yeah. I don't really think these belong on `cy` itself, but they've been there for years and I'm not sure I want to move them as part of this release. :/

I'll include adding TS types for these methods as part of my next code PR.

content/api/commands/focus.md
BlueWinds: Updated the recommendation in `.blur()` to make it clearer.

> This element must currently be in focus. If you want to ensure an element is
focused before blurring, try using [`.focus()`](/api/commands/focus) before
`.blur()`.

> ```javascript
> cy.get('button').as('btn').focus()
> cy.get('@btn').blur()
> ```

content/api/commands/blur.md
BlueWinds: Whoops. Fixed in all files.

chrisbreiding (approved):

content/api/cypress-api/custom-queries.md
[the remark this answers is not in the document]
BlueWinds: https://github.com/cypress-io/cypress/pull/24697 - will move this out of draft once tests pass.

PR #4842, opened 2022-11-07
Update validation error scenarios

emilyrohrbough:
Docs PR for https://github.com/cypress-io/cypress/pull/24511


content/api/commands/session.md
```diff
@@ -140,10 +140,10 @@ before `setup` runs, regardless of the testIsolation configuration.
...
+| ------------------ | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `validate`         | `undefined` | Validates the newly-created or restored session.<br><br>Function to run immediately after the session is created and `setup` function runs or after a session is restored and the page is cleared. If throws an exception, contains any failing Cypress command, returns a Promise which rejects or resolves to `false`, or the last Cypress command yielded `false`, the session is considered invalid.<br><br>- If validation fails immediately after `setup`, the test will fail.<br>- If validation fails after restoring a session, `setup` will re-run. |
```
chrisbreiding, line 145:
```suggestion
| `validate`         | `undefined` | Validates the newly-created or restored session.<br><br>Function to run immediately after the session is created and `setup` function runs or after a session is restored and the page is cleared. If it throws an exception, contains any failing Cypress command, returns a Promise which rejects or resolves to `false`, or the last Cypress command yielded `false`, the session is considered invalid.<br><br>- If validation fails immediately after `setup`, the test will fail.<br>- If validation fails after restoring a session, `setup` will re-run. |
```

This might be more of a question for the code PR, but why allow `return Promise.resolve(false)` (or even `cy.wrap(false)`) but not `return false`? I feel like we should simplify it to only have errors fail the validation.

chrisbreiding (changes requested):

content/api/commands/session.md
emilyrohrbough: It's what @mjhenkes and I had discussed with Brian. Not sure the background of why we are support so many failure options.

chrisbreiding: Gotcha. There's still a minor typo here that needs to be addressed.

content/api/commands/session.md
```diff
@@ -396,13 +404,6 @@ function validate() {
   cy.visit('/account', { failOnStatusCode: false })
   cy.url().should('match', /^/account/)
 }
-
-// Or just return false if the session is invalid
```
BlueWinds, line 400: Do we not want an example for "the `validate` function returns a Promise that resolves to `false` or rejects" anymore?

BlueWinds (approved):

content/api/commands/session.md
emilyrohrbough: I'm going to do more comprehensive evaluation for the docs for the v12 release.

AtofStryker (approved):

PR #4846, opened 2022-11-08
Remove CT adapter readme links

astone123:
[We're removing a lot of info from the adapter readme documents](https://github.com/cypress-io/cypress/pull/24590), so we shouldn't link to them from our docs page anymore.

### DIFF
--- a/content/api/commands/mount.md
+++ b/content/api/commands/mount.md
@@ -35,6 +35,10 @@ to start with for your commands:
 <template #react>
 
 ```js
+// React 18
+import { mount } from 'cypress/react18'
+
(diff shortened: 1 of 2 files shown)


content/api/commands/mount.md
```diff
@@ -35,6 +35,10 @@ to start with for your commands:
 <template #react>
 
 ```js
+// React 18
```
astone123, line 38: Not sure if we actually want to do this... it seems consistent with the way that we do the Vue/Vue2 examples

elylucas (approved):

PR #4851, opened 2022-11-10
Update cy.session API docs for v12

### LINKED ISSUE #4507: Update cy.session api page
- Remove experimental info
- Command Log
    - Currently documented in `cy.session()` API docs [[here](https://docs.cypress.io/api/commands/session#Command-Log)](https://docs.cypress.io/api/commands/session#Command-Log)
    - Does this need to be moved?
    - Update content & screenshots
 - Update test isolation info

debrisapron:
Closes #4507

~~This is a draft PR as I still haven't used the updated API, once I can confirm that everything I've written is correct I'll mark it ready for review.~~ 👍 

The main changes here are addressing the update to the behavior of the `validate` option, and around when state and/or page content gets wiped.

~~There are other issues with this page, in particular a few sections I don't really understand the point of, but I will address those in a later PR.~~ NVM I think I managed to clarify it a bit


content/api/commands/session.md
```diff
@@ -110,10 +111,10 @@ before `setup` runs, regardless of the testIsolation configuration.
...
+| `validate`         | `undefined` | Validates the newly-created or restored session.<br><br>Function that runs immediately after the `setup` function is run **and** after a cached session is restored. If this returns `false`, throws an exception, contains any failing Cypress command, or returns a Promise which rejects or resolves to `false`, the session is considered invalid.<br><br>- If validation fails immediately after setup, the test will fail.<br>- If validation succeeds immediatedly after setup, the session will not be saved until `validate` is finished.<br>- If validation fails after restoring a session, `setup` will re-run. |
```
emilyrohrbough, line 116:
```suggestion
| `validate`         | `undefined` | Validates the newly-created or restored session.<br><br>Function that runs immediately after the `setup` function is run **and** after a cached session is restored. If this returns `false`, throws an exception, contains any failing Cypress command, or returns a Promise which rejects or resolves to `false`, the session is considered invalid.<br><br>- If validation fails immediately after setup, the test will fail.<br>- If validation succeeds immediately after setup, the session will not be saved until `validate` is finished.<br>- If validation fails after restoring a session, `setup` will re-run. |
```

content/api/commands/session.md
```diff
@@ -146,13 +145,11 @@ Cypress.Commands.add('login', (username, password) => {
 ```javascript
 Cypress.Commands.add('login', (username, password) => {
   cy.session([username, password], () => {
```
emilyrohrbough, line 147: any ID with username & password should be removed from the examples IMO. We dont' have a log:false option and we don't want to direct users to use sensitive information which they may not want video recorded/face-up.

content/api/commands/session.md
```diff
@@ -340,13 +265,27 @@ it('should transfer money between users', () => {
...
+
+After `setup`, the session is not saved until `validate` has returned or
+resolved. This allows you to use `validate` to ensure that the login flow is
+fully complete by, for example, repeatedly checking that the user's session
+token has been set. This must be done with a
+[retry-able command](/guides/core-concepts/retry-ability), or your assertion may
+fail the test simply because the page is running a little slow. Once your
+assertion passes, the session is cached.
+
+When restoring a saved session, if the `validate` function return `false`,
```
emilyrohrbough, line 279:
```suggestion
When restoring a saved session, if the `validate` function
```

content/api/commands/session.md
```diff
@@ -340,13 +265,27 @@ it('should transfer money between users', () => {
...
+and `setup` will be re-run. So your `validate` function serves a dual purpose,
+it ensures that the initial setup has worked correctly **and** it checks that
+the session remains valid before re-use.
 
-The page is not cleared after the `validate` function is executed. If you use
-`cy.visit()` in your validation, your test will continue on the visited page
-once `cy.session()` succeeds.
+Note that the page is not cleared after the `validate` function is executed. If
+you use `cy.visit()` in your validation, your test will continue on the visited
+page once `cy.session()` succeeds.
```
emilyrohrbough, line 288: This is no longer true. When`testIsolation='on'`, we always clear before the command finishes. If `testIsolation='off'` the page will remain in tact.
```suggestion
```

content/api/commands/session.md
```diff
@@ -368,6 +307,7 @@ function validate() {
 }
 
 // Or just return false if the session is invalid
+// TODO: How do they check this repeatedly? Maybe we need a `cy.retry` method?
 function validate() {
   if (!MyApp.isSessionValid()) {
     return false
```
emilyrohrbough, line 313: Return false is no longer a value "fail" option.

content/api/commands/session.md
```diff
@@ -340,13 +265,27 @@ it('should transfer money between users', () => {
 
 ### Validating the session
 
-If the `validate` function return `false`, throws an exception, returns a
-Promise that resolves to `false` or rejects, or contains any failing Cypress
-command, the session will be considered invalid, and `setup` will be re-run.
+The `validate` function has a slightly different function depending on whether
```
emilyrohrbough, line 268: > has a slightly different function

I'd re-word to indicate this function should be able to run after setup or a restored context and should not rely on the previous page state. 

It would also be worth expanding on this is intended to ensure authentication with the back-end to ensure the application is authenticated correctly as the user interacts with the page and/or when new requests are made

content/api/commands/session.md
debrisapron: This is a legitimate concern but we currently do this absolutely everywhere - guides, API docs, recipes, RWA, Kitchen Sink etc etc. If we're going to ban this practice there needs to be a larger conversation about that.

debrisapron:
Updated page preview here: https://cypress-documentation-f1rw1rmci-cypress-io.vercel.app/api/commands/session#Syntax

content/api/commands/session.md
emilyrohrbough: We don't need to ban it, but we should remove it from our examples to provide good guidance. It is something we need to be updating in our documentation and examples.

mjhenkes (approved):

emilyrohrbough (approved):

PR #4865, opened 2022-11-17
Add initial definitions and layout for release stage page

rockindahizzy:
This page sets out to better define how we approach releases to features that are in different stages of development.


content/guides/references/release-stages.md
```diff
@@ -0,0 +1,105 @@
...
+title: Release Stages
+---
+
+Cypress follows Semantic Versioning so that it is clear how you may be affected
+when upgrading to a newer version of Cypress. If there is a minor version
+increment, 11.0.0 → 11.1.0 for example, you should not experience disruption in
+your usage of Cypress. If there is a major version increment, 11.1.0 → 12.0.0
+for example, you may need to make changes to continue using Cypress. For most
+scenarios this is a valid way to consider complexity when upgrading Cypress. In
+some cases, however, a feature of cypress may be in a stage of development in
```
lmiller1990, line 11:
```suggestion
some cases, however, a feature of Cypress may be in a stage of development in
```

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,105 @@
...
+stages.
+
+## Experimental
+
+Experimental features consist of features that we are looking to gather user
+feedback on before we commit fully to the initial implementation or the feature
+entirely. These type of features require you to opt-in to via a configuration
+value before the feature is activated. These types of features may be unstable
+and contain issues. You should expect to see frequent changes and improvements
+during the feature’s experimental lifetime.
```
lmiller1990, line 24: Consider linking to our current ongoing experiments page

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,105 @@
...
+  - Affected users will be pointed to documentation if there blockers are
+    perceived to be possible.
+- Breaking Changes
+  - Breaking changes can be introduced during any release cycle and do not bump
+    semver by a major version
+- Removals
+  - Removal of any new functionality related to the feature can happen during
+    any release cycle and does not bump semver by a major version.
+
+## Beta
```
lmiller1990, line 59: We should clarify how a user can know if a feature is alpha/beta/GA etc. For CT it was easy, there was a label, but for thing like cy.origin, this might not always be the case. Any thoughts on this? I wonder if the experiments page could be expanded to have a "Stage" for each experiment - but I suppose it wouldn't really be an experiments page then, but a "pre-release features" page.

lmiller1990 (commented):
I know this isn't done but I wanted to give some early feedback

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,105 @@
+---
+title: Release Stages
+---
+
+Cypress follows Semantic Versioning so that it is clear how you may be affected
+when upgrading to a newer version of Cypress. If there is a minor version
+increment, 11.0.0 → 11.1.0 for example, you should not experience disruption in
```
emilyrohrbough, line 7:
```suggestion
increment,` 11.0.0` → `11.1.0` for example, you should not experience disruption in
```

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,105 @@
+---
+title: Release Stages
+---
+
+Cypress follows Semantic Versioning so that it is clear how you may be affected
```
emilyrohrbough, line 5: can we link to the sem ver doc?

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,105 @@
...
+when upgrading to a newer version of Cypress. If there is a minor version
+increment, 11.0.0 → 11.1.0 for example, you should not experience disruption in
+your usage of Cypress. If there is a major version increment, 11.1.0 → 12.0.0
+for example, you may need to make changes to continue using Cypress. For most
+scenarios this is a valid way to consider complexity when upgrading Cypress. In
```
emilyrohrbough, line 10: complexity or disruptive changes. personally dont love the term complexity because not all major version bumps are complex, just something we've changed will def require action from the user and/or break their current usage.

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,105 @@
...
+transparency of when this might be the case, we have consolidated definitions of
+different release stages and how we approach changes and communication in those
+stages.
+
+## Experimental
+
+Experimental features consist of features that we are looking to gather user
+feedback on before we commit fully to the initial implementation or the feature
+entirely. These type of features require you to opt-in to via a configuration
+value before the feature is activated. These types of features may be unstable
```
emilyrohrbough, line 22: >  These type of features require you to opt-in to via a configuration
value before the feature is activated. 

I do not think we should narrow ourselves into a box with a configuration flag. Reading the terms for `Experimental` vs `Alpha`, they sounds almost 1:1 in terms of what users can expect in breaking changes or removal of functionality. 

Also reading the definitions of `Alpha` and `Beta`, they are identical. The key differences are how changes / breaking changes are released.

Is there any way we can condense `Alpha` & `Experimental`? What are the reasons not to condense them?

There will be "beta" E2E features which will alway sneed to be "opted" in to via configuration, which isn't to different than users "opting" into using Beta component testing setups which is also configuration based.

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,105 @@
...
+when upgrading to a newer version of Cypress. If there is a minor version
+increment, 11.0.0 → 11.1.0 for example, you should not experience disruption in
+your usage of Cypress. If there is a major version increment, 11.1.0 → 12.0.0
+for example, you may need to make changes to continue using Cypress. For most
+scenarios this is a valid way to consider complexity when upgrading Cypress. In
+some cases, however, a feature of cypress may be in a stage of development in
+which it does not adhere to these rules. In order to better provided
+transparency of when this might be the case, we have consolidated definitions of
+different release stages and how we approach changes and communication in those
+stages.
```
emilyrohrbough, line 15: Can we list these stages? I think it might be easier for reader to read from in the this order GA -> Beta -> Alpha -> Experimental, since GA is what they will be expecting with semantic versioning.

content/guides/references/release-stages.md
emilyrohrbough: Actually looking closer at `Alpha` vs `Beta`, is the only different for users in confidence on if we provide migration message if we break APIs or remove functionality?

lmiller1990: I had this thought too... I wonder if we really need 4 stages at all? On the other hand, In other OSS projects I work on, we've often had 4 stages:

- alpha: Basically what we call experimental - you can try it, the main purpose is feedback and power users. Expect things to not work perfectly. Things are missing.
- beta: All features are present, but likely there are some bugs/unexpected behavior.
- release candidate: Similar the final release in terms of API and features - the main point of this stage is the API is set and there won't be any major breaking changes. Users can start adopting the feature with confidence.
- GA - it's ready.

I wonder if these could help mold our definitions?

content/guides/references/release-stages.md
rockindahizzy: How about this?
```suggestion
scenarios this is a valid way to consider how disruptive upgrading to the next version of Cypress may be. In
```

content/guides/references/release-stages.md
rockindahizzy:
```suggestion
Cypress follows Semantic Versioning ([see summary here](https://semver.org/#summary)) so that it is clear how you may be affected
```

content/guides/references/release-stages.md
rockindahizzy:
```suggestion
during the feature’s experimental lifetime. The current experimental features can be found [here](guides/references/experiments).
```

content/guides/references/release-stages.md
rockindahizzy: I've made adjustments to the paragraph highlighted and the overall order of the stages in [this commit](https://github.com/cypress-io/cypress-documentation/pull/4865/commits/d72d840821ae2ceec050ecae8afdfdedd9263422)

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,110 @@
...
+Beta features consist of features that we are looking to gather user feedback on
+by providing them as default functionality before an official release. Beta
+features may not be feature complete, but they should be close enough where
+changes are few and far between.
+
+**Strategies for releasing and communicating changes to Beta features**
+
+- Modifications
+  - Modifications can be introduced during any release cycle and may or may not
+    increment semver.
```
mjhenkes, line 54: Wouldn't any release increment semver?

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,110 @@
...
+  - In the majority of cases, removals will be proceeded by a period in which
+    the feature is ‘deprecated’
+  - Migration steps will be provided to minimize pain points in these situations
+
+## Beta
+
+Beta features consist of features that we are looking to gather user feedback on
+by providing them as default functionality before an official release. Beta
+features may not be feature complete, but they should be close enough where
+changes are few and far between.
```
mjhenkes, line 48: Maybe we should specify the caliber of changes that are few and far between.

```suggestion
by providing them as default functionality before an official release. Beta
features may not be feature complete, but they should be close enough where
large, possibly breaking, changes to, either the API or functionality are few and far between.
```

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,110 @@
...
+
+- Modifications
+  - Modifications can be introduced during any release cycle and may or may not
+    increment semver.
+  - Affected users will be pointed to documentation if there blockers are
+    perceived to be possible.
+- Breaking Changes
+  - Breaking changes can be introduced during any release cycle and do not bump
+    semver by a major version
+  - Breaking changes require in-app messaging to help users unblock themselves
```
mjhenkes, line 60:
```suggestion
  - Breaking changes require in-app messaging to help users unblock themselves
  - The introduction and impact of breaking changes should be carefully considered at this stage as the cost to users is higher than with an alpha feature.
```

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,110 @@
...
+    as needed
```
mjhenkes, line 61:
```suggestion
  - Breaking changes require in-app messaging to help users unblock themselves
    as needed
  - The introduction and impact of breaking changes should be carefully considered at this stage as the cost to 
    users is higher than with an alpha feature.
```

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,110 @@
...
+    semver by a major version
+- Removals
+  - Removal of any new functionality related to the feature can happen during
+    any release cycle and does not bump semver by a major version.
+
+## Experimental
+
+Experimental features consist of features that we are looking to gather user
+feedback on before we commit fully to the initial implementation or the feature
+entirely. These type of features require you to opt-in to via a configuration
```
mjhenkes, line 93:
```suggestion
Experimental features consist of features that we are looking to gather user
feedback on before we commit fully to further development of the feature.
These type of features require you to opt-in to via a configuration
```

content/guides/references/release-stages.md
mjhenkes: In concept and definition I'm good with the stages listed here. The main point of the stage is convey to our users the level of risk they're taking on by relying on the feature:
* experimental - High risk, this feature may disappear
* alpha - Medium risk, this feature may change drastically before becoming GA, but it'll likely get done eventually.
* beta - low risk, this is pretty much done, but might have some breaking changes here and there for the better.
* GA - lowest risk, breaking changes are rare and will be clearly conveyed. Cypress is committed to this feature for the long haul.

The other thing to keep in mind is that we don't have to progress through the stages directly, we can skip from experimental to GA if we want (cy.origin) or start with a beta if we're confident enough.

mjhenkes (approved):
This all looks good to me. I added a few minor wording changes but nothing that would prevent me from approving.

content/guides/references/release-stages.md
rockindahizzy:
```suggestion
  - Modifications can be introduced during any release cycle.
```

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,111 @@
...
+- Removals
+  - Removals will only be introduced during major releases in which semver is
+    bumped by a major version
+  - In the majority of cases, removals will be proceeded by a period in which
+    the feature is ‘deprecated’
+  - Migration steps will be provided to minimize pain points in these situations
+
+## Beta
+
+Beta features consist of features that we are looking to gather user feedback on
```
marktnoonan, line 45:
```suggestion
Beta features are features that we are looking to gather user feedback on
```

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,111 @@
...
+by providing them as default functionality before an official release. Beta
+features may not be feature complete, but they should be close enough where
+large, possibly breaking, changes to, either the API or functionality are few and far between.
```
marktnoonan, line 48:
```suggestion
large, possibly breaking, changes to, either the API or functionality are few and far between, 
and would be largely motivated by real-world user feedback provided during the Beta stage.
```

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,111 @@
...
+
+**Strategies for releasing and communicating changes to Beta features**
+
+- Modifications
+  - Modifications can be introduced during any release cycle.
+  - Affected users will be pointed to documentation if there blockers are
+    perceived to be possible.
```
marktnoonan, line 55:
```suggestion
  - Modifications will be documented, with recommended code changes
  highlighted if appropriate.
```

If there are blockers, presumably we'd consider it breaking, so suggesting new wording here.

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,111 @@
...
+  - Removal of any new functionality related to the feature can happen during
+    any release cycle and will increment semver. These do not need to coincide
+    with a major release that is bumping semver by a major version.
+  - Removals at this stage will be rare and a migration strategy along with
+    reasoning for the removal will be documented.
+
+## Alpha
+
+Alpha features consist of features that we are looking to gather user feedback
+on by providing them as default functionality before an official release. Alpha
```
marktnoonan, line 73: > Alpha features consist of features that we are looking to gather user feedback
on by providing them as default functionality before an official release.

This is the same description as Beta, is that intended? I guess the second sentence represents the difference between the two. Maybe we should reorder the sentences to the differences come first. Or tweak some other way.

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,111 @@
...
+features may not be feature complete and changes are possible during this stage.
+
+**Strategies for releasing and communicating changes to Alpha features**
+
+- Modifications
+  - Modifications can be introduced during any release cycle and may or may not
+    increment semver.
+  - Affected users will be pointed to documentation if there blockers are
+    perceived to be possible.
```
marktnoonan, line 82:
```suggestion
  - Modifications will be documented, with recommended code changes
  highlighted if appropriate.
```

Same change as suggested above.

marktnoonan (commented):
I like it! I'd suggest adding a section to each Stage called something like "Stability" that would clarify whether breaking changes are expected or not. This would help define the differences between alpha and beta. I think it's good if users can understand the "risk" of adopting a feature from this perspective, even though it might be a bit repetitive.

Suggestions based on the doc:

GA - Feature is stable. Breaking changes not allowed outside of a new major version of Cypress
Beta - Feature is close to stable. Breaking changes not expected, but possible if indicated by learnings from the Beta stage. Beta is the evaluation stage of a proposed final state.
Alpha - The feature is on the path to release but not yet considered stable. Breaking changes are possible. 
Experimental - The experiment is necessary to gather information about the best future direction of the feature. Breaking changes are probable. 

This content could also be added just as summary under each "breaking changes" section, curious what you think.

content/guides/references/release-stages.md
marktnoonan: I made some comments [here](https://github.com/cypress-io/cypress-documentation/pull/4865#pullrequestreview-1192021452) before reading this thread, but broadly I agree with these definitions.

> The other thing to keep in mind is that we don't have to progress through the stages directly, we can skip from experimental to GA if we want (cy.origin) or start with a beta if we're confident enough.

I think it's worth adding some mention of this, that it's not required for features to hit each stage.

The reason I think we should keep both alpha and experimental is that they communicate different ideas about whether something is "on the path to release". Sometimes, the outcome of an experiment will be "we learned some things, and we decided not to build the feature". 

For an alpha stage feature, my supposition is that we've decided we definitely want it to be a part of Cypress, _that_ decision is behind us, and it's just a matter of how complete/defined the implementation is whether we call it alpha or beta. Users should expect that it's safe to use/depend on in the sense that we don't expect to remove it, even though breaking changes may occur that they have to keep up with.

I expect alpha will be the least-used stage, but it makes sense for say, a new framework or bundler in CT, where we've decided to support it, but implementation will come in stages or depends on feedback. If we aren't sure we want to support it, we'd use the experimental flag to enable, so users don't discover something in the UI that's a speculative feature on our part.

marktnoonan: When I dug into this, I was _sure_ I could make it so alpha and experimental had exactly the same definition, but I talked myself out of it for the above reasons.

content/guides/references/release-stages.md
```diff
@@ -0,0 +1,111 @@
...
+- Removals
+  - Removal of any new functionality related to the feature can happen during
+    any release cycle and does not bump semver by a major version.
+
+## Experimental
+
+Experimental features consist of features that we are looking to gather user
+feedback on before we commit fully to further development of the feature.
+These type of features require you to opt-in to via a configuration
+value before the feature is activated. These types of features may be unstable
```
emilyrohrbough, line 95: I'm still not sold on calling out an experiment is `opt-in to via a configuration` because some "alpha/beta" ready concepts will need an opt'd-in to  on the end-to-end side and won't necessarily be able to provided by default.

marktnoonan: Would it be better this way:

> These type of features require you to opt-in to via an `experimental`-prefixed configuration value

I think that came up in one of our chat about this. 

Agree that not all features (even fully released features) can be on by default at the same time. I think what's more important is that users can't accidentally turn on an experiment without knowing that it's an experiment.

lmiller1990:
Quite a few comments, can we get the final changes made and merge this up by EOB tomorrow @rockindahizzy? We can always iterate on the fine details and language, as long as the general definitions are reasonable, I think that's the most important thing.

content/guides/references/release-stages.md
rockindahizzy:
```suggestion
These type of features require you to opt-in to via a configuration
value prefixed with `experimental`. These types of features may be unstable
```

lmiller1990 (approved):
Good

Just need to fix linting

PR #4872, opened 2022-11-23
update CT mount docs

marktnoonan:
Small updates to fix some examples to match new mount API, standardized wording around `MountReturn`.

### DIFF
--- a/content/guides/component-testing/react/api.md
+++ b/content/guides/component-testing/react/api.md
@@ -123,7 +123,7 @@ it('mounts', () => {
 
 ### MountReturn
 
-Type that the `mount` function returns
+Type that the `mount` function yields
 
 <table class="api-table">
   <caption>members</caption>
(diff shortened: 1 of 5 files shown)


elylucas (approved):

lmiller1990 (approved):

content/guides/component-testing/react/api.md
```diff
@@ -123,7 +123,10 @@ it('mounts', () => {
 
 ### MountReturn
 
+Type that the `mount` function returns
```
emilyrohrbough, line 126: Shouldn't this be yielded, not returned?

lmiller1990: Ah yeah, good catch - wrong verbiage. 

Just to clarify for anyone reading/wondering about this

```js
cy.mount(<Foo />).then(val => ...)
```

The `val` is *yielded*. `cy.mount` _does_ return something, a `Chainable` (so you can do `cy.mount(...).get(...)`). The yielded value is what you get in the `then`.

marktnoonan: Yeah, I copied existing verbiage from the Angular, command, will update both.

marktnoonan: cool, updated everywhere

PR #4898, opened 2022-12-05
.within() now throws an error if given more than one subject

BlueWinds:
Related code PR: https://github.com/cypress-io/cypress/pull/24975

This brings the docs in line with the new behavior.


marktnoonan (approved):

marktnoonan:
Already approved but actually how about updating the examples a little bit:

Screen Shot 2022-12-05 at 3 32 05 PM

`cy.get('.className')` is only safe when it happens to yield just one element. Maybe we should use something like `contains()` that always yields one thing, or add a `.first()`. And under "Incorrect Usage" we could add an example where multiple things are yielded by the previous command.

content/_changelogs/12.0.0.md
```diff
@@ -11,6 +11,11 @@ _Released MM/DD/YYYY_
 - Cypress now throws an error if commands are invoked from inside a `.should()`
   callback. This previously resulted in unusual and undefined behavior; it is
   now explicitly an error.
+- The `.within()` command now throws an error if given more than one DOM element
```
emilyrohrbough, line 14: @BlueWinds Can you add a migration guide entry on how to resolve this error?

emilyrohrbough, line 14:
```suggestion
- The [`.within()`](/api/commands/within) command now throws an error if given more than one DOM element
```

BlueWinds:
> `cy.get('.className')` is only safe when it happens to yield just one element. Maybe we should use something like `contains()` that always yields one thing, or add a `.first()`. And under "Incorrect Usage" we could add an example where multiple things are yielded by the previous command.

Updated the usage section.

**<Icon name="check-circle" color="green"></Icon> Correct Usage**

```javascript
cy.get('.list').first().within(($list) => {}) // Yield the first `.list` and scope all commands within it
```

**<Icon name="exclamation-triangle" color="red"></Icon> Incorrect Usage**

```javascript
cy.within(() => {}) // Errors, cannot be chained off 'cy'
cy.getCookies().within(() => {}) // Errors, 'getCookies' does not yield DOM element
cy.get('div').within(($divs) => {}) // Probably errors, because get('div') yields multiple elements
```

content/_changelogs/12.0.0.md
BlueWinds: Done.

BlueWinds: Added migration guide.

content/guides/references/migration-guide.md
```diff
@@ -310,6 +310,69 @@ cy.get('button')
    .then(res => { ...handle response... })
 ```
 
+#### `.within()`
+
+[`.within()`](/api/commands/within) now throws an error if it is passed multiple
```
emilyrohrbough, line 315:
```suggestion
The [`.within()`](/api/commands/within) command now throws an error if it is passed multiple
```

content/guides/references/migration-guide.md
```diff
@@ -310,6 +310,69 @@ cy.get('button')
...
+elements as the subject. This previously resulted in inconsistent behavior,
+where some commands would use all passed in elements, some would use only the
+first and ignore the rest, and `.screenshot()` would throw an error.
```
emilyrohrbough, line 318: add link to screenshot cmd

content/guides/references/migration-guide.md
```diff
@@ -310,6 +310,69 @@ cy.get('button')
...
+
+```diff
+cy.get('button')
+-  .should(($button) => {
+
+    })
++  .then(api => api.makeARequest('http://example.com'))
+   .then(res => { ...handle response... })
+```
+
```
emilyrohrbough, line 375: copy paste?

content/guides/references/migration-guide.md
```diff
@@ -310,6 +310,69 @@ cy.get('button')
...
+```diff
+cy.get('tr')
+-  .within(() => {
+-    cy.contains('foo')
+-    cy.contains('bar')
+-  })
++  .as('rows') // Store multiple elements as an alias
+
++cy.get('@rows').contains('foo')
```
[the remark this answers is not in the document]
BlueWinds: I'll change the example; this works for .contains, but they might be running assertions, performing actions, or doing anything else that can't be used in this way, which was my intent in this example.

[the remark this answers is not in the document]
BlueWinds: (any set of commands using `.within()` can be written without .within(). It doesn't provide any *unique* functionality)

content/guides/references/migration-guide.md
BlueWinds: Applied this change to all the bits in this section (.invoke() and .should() were phrased the same way).

content/guides/references/migration-guide.md
BlueWinds: Done.

mjhenkes (approved):

emilyrohrbough (approved):

PR #4913, opened 2022-12-07
Small doc fixes in Retry-ability Guide

wlsf82:
Small doc fixes that I missed on my previous PR: https://github.com/cypress-io/cypress-documentation/pull/4910.

## Fix 1

This piece of code `.children('li').should('have.length', 1)` refers to:

```javascript
cy.get('.todoapp') // query
  .find('.todo-list li') // query
  .should('have.length', 1) // assertion

```

And so, it should actually be `.find('.todo-list li').should('have.length', 1)`

## Fix 2

This text _**Cypress will retry for up to 10 seconds to find a visible element of class `mobile-nav` with text containing "Home".**_ refers to this piece of code:

```javascript
// we've modified the timeout which affects default + added assertions
cy.get('[data-testid="mobile-nav"]', { timeout: 10000 })
  .should('be.visible')
  .and('contain', 'Home')

```

So, I suggest it to be _**Cypress will retry for up to 10 seconds to find a visible element with `data-testid` attribute `mobile-nav` with text containing "Home".**_

## Fix 3

This text _**Any command that isn't a query, such as `cy.click()`, follows different rules than queries do.**_ refers to the `.click()` command as if it could be called directly from the `cy` object, but since it cannot, I'm suggesting it to be _**Any command that isn't a query, such as `.click()`, follows different rules than queries do.**_

Cc. @marktnoonan.

### DIFF
--- a/content/guides/core-concepts/retry-ability.md
+++ b/content/guides/core-concepts/retry-ability.md
@@ -111,8 +111,8 @@ app.TodoModel.prototype.addTodo = function (title) {
 ```
 
 The test still passes! `cy.get('.todo-list')` passes immediately - the
-`todo-list` exists - but `.children('li').should('have.length', 1)` show the
-spinning indicators, meaning Cypress is requerying for them.
+`todo-list` exists - but `.find('.todo-list li').should('have.length', 1)` show
+the spinning indicators, meaning Cypress is requerying for them.
 
 <DocsImage src="/img/guides/retry-ability/retry-assertion.gif" alt="Retrying assertion"></DocsImage>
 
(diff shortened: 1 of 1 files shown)


wlsf82:
👋 Hi @marktnoonan, could you review this PR, please?
I hope you like the contributions.

emilyrohrbough (approved):

marktnoonan:
Hi @wlsf82, I don't have the power to trigger the build so I'm tagging @jaffrepaul

jaffrepaul:
👍 currently building

BlueWinds (approved):

PR #4923, opened 2022-12-09
Rebrand two Test Orchestration features

debrisapron:
Addresses #4929 
Run failed specs first -> Spec Prioritization
Cancel test run when a test fails -> Auto Cancellation


content/guides/cloud/introduction.md
```diff
@@ -67,11 +67,11 @@ With our [Smart Orchestration](/guides/cloud/smart-orchestration) features, you
...
-to surface problems earlier, and
-[cancel whole test runs on failure](/guides/cloud/smart-orchestration#Cancel-test-run-when-a-test-fails)
-to save on resource usage. You can also
+setup required! You can prioritize recently failed specs with
+[Spec Prioritization](/guides/cloud/smart-orchestration#Spec-Prioritization) to
+surface problems earlier, and cancel whole test runs on failure with
+[Auto Cancellation](/guides/cloud/smart-orchestration#Auto-Cancellation) to save
+on resource usage. You can also
 [cancel in-progress runs](/guides/cloud/runs#Cancel-run) manually from Cypress
 Cloud if you need to.
```
ryanpei, line 77: I'm wondering if we can refactor the "Analyze and diagnose" section above, because that to me is actually two separate and equally very powerful themes for Cloud. I'd maybe call one "Debugging and root cause analysis", and have that go into the benefits of the individual test debug features, and the other is "Diagnosing and improving overall test health and performance", which can focus on Analytics, Latest Tests, and now, ROAR. ROAR's added benefit is seeing not only the run duration dimension of overall test health, but also some "Recommendations" to improve this measure.

The only feature that I see still wouldn't fit very neatly into either theme is Flake Detection, but I suppose it can be slotted in the latter one, for now.

content/guides/cloud/smart-orchestration.md
```diff
@@ -42,41 +41,40 @@ time.**
...
 
 <Alert type="success">
 
 <strong class="alert-header"><Icon name="star"></Icon> Premium Cypress Cloud
 Feature</strong>
 
-**Running failed specs first** is a _Smart Orchestration_ feature available to
-users with a [Business Cypress Cloud plan](https://cypress.io/pricing).
+**Spec Prioritization** is a _Smart Orchestration_ feature available to users
+with a [Business Cypress Cloud plan](https://cypress.io/pricing).
```
ryanpei, line 52: I think we can expand this callout to help the reader understand the impact ROAR can have on their decision about potentially upgrading their subscription, assuming they are a Team tier org. Something along the lines of, "Wondering if upgrading your Team tier subscription is worth it? You can always check out the Run Overview page to see how much time we predict these features would save you"

content/guides/cloud/smart-orchestration.md
```diff
@@ -99,48 +97,47 @@ failure will:
...
 <Alert type="success">
 
 <strong class="alert-header"><Icon name="star"></Icon> Premium Cypress Cloud
 Feature</strong>
 
-**Canceling test runs when a test fails** is a _Smart Orchestration_ feature
-available to users with a
-[Business Cypress Cloud plan](https://cypress.io/pricing).
+**Auto Cancellation** is a _Smart Orchestration_ feature available to users with
+a [Business Cypress Cloud plan](https://cypress.io/pricing).
```
ryanpei, line 108: and same comment as above, about using ROAR to help an "undecided" reader make an informed decision here

ryanpei (commented):
@debrisapron I realize your original task here was perhaps only concerned with renaming the features, but if it's helpful I also suggested changes to address a few other issues I saw related to Auto Cancellation or Spec Prioritization, as well as some ideas for how we can enhance the persuasiveness of the content. Let me know if you think a different PR is more appropriate for these.

content/guides/cloud/smart-orchestration.md
```diff
@@ -99,48 +97,47 @@ failure will:
```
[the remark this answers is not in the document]
debrisapron: Good question!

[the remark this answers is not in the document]
debrisapron: I think the Spec Prioritization section has the "Consider the Scenario" alert instead and we don't want to drown the user in alerts. But I agree it would be better to align the two sections. Will circle back.

content/guides/cloud/introduction.md
debrisapron: I think this is a good idea but belongs in another PR

content/guides/cloud/smart-orchestration.md
debrisapron: I like this a lot, using the recommendations as another funnel.

debrisapron:
Updated content preview: https://cypress-documentation-h6yzwwhb6-cypress-io.vercel.app/guides/cloud/smart-orchestration

debrisapron:
Updated again: https://cypress-documentation-ppzefcief-cypress-io.vercel.app/guides/cloud/smart-orchestration

content/guides/cloud/smart-orchestration.md
```diff
@@ -91,56 +96,61 @@ changes.
...
 To enable or disable this feature:
 
 1. Select the desired project within Projects view to visit Cypress Cloud.
 2. Click "Project Settings" with the right-hand sidebar.
-3. Scroll to the Smart Orchestration section within Project Settings page
-4. Toggle "Cancel run when a test fails."
+3. Scroll to the Smart Orchestration section within Project Settings page.
+4. Toggle "Auto Cancellation".
+5. Set the "failures before auto canceling" threshold (or just leave it at the
+   recommended value).
```
ryanpei, line 144:
```suggestion
   default value of 1).
```
This isn't necessarily a "recommended" value; we recognize this is case-by-case.

ryanpei (approved):

PR #4944, opened 2022-12-21
docs: updating styles with style guidelines

elylucas:


jaffrepaul (commented):
Cool! I had wanted to break these css files up for awhile. 

Notice any build lag with new import rules? The article styles didn't make it into the `custom.scss` file so the main content well is wider and more pushed against the TOC. In some cases it leaves a lot of dead space in the admonitions and code blocks. Was that intentional?

elylucas:
@jaffrepaul ya it was intentional. I felt it was limiting the size of the content area unnecessarily on larger screens. I added a gap to the content area when the size of the main area gets too close to the sidebar for smaller screens. Let me know if that takes care of that concern. For the other concern, do you have any instances where the admonitions look weird?

elylucas:
Also haven't noticed any build lag. Is that a known issue?

jaffrepaul:
Gotcha. It's subtle but also felt like a big change. Not on a big screen atm, would have to check later. It's not a framework known issue. css `@import` loads styles separately (more http calls) vs in parallel. It's more site load time than build time that it could impact. We probably don't have enough to make a noticeable difference. I was just curious about this when I was planning on trying this maneuver.

PR #4953, opened 2023-01-03
[ DOCS ] Vue component testing/events

jakarlse88:
Update Vue component testing docs. These changes were motivated by [this discussion](https://github.com/cypress-io/cypress/discussions/25198), and specifically:

- ~moves assertions on event spies to `then()` in order to make sure that the event has a chance both to fire and "register"~
- notes that an example event defined with `defineEmits('change')` will be compiled by Vue to `props.onChange()`


marktnoonan:
Thanks for this contribution! I agree with adding this part for clarity.

> while the event is defined as `change`, the spy is passed for the prop
`onChange`; this is due to the fact that Vue compiles `emit('change')` to
`props.onChange()`).

But I'm not sure this is correct:

> We chain `then()` off of `cy.get()` to make sure the event has
fired and registered as such

We have many tests that assert on an event spy without wrapping in `.then()` callback, here's [one example](https://github.com/cypress-io/cypress/blob/develop/packages/app/src/specs/CreateSpecCard.cy.tsx#L50-L57).

Do you have a situation where the `cy.get('@yourAlias')` doesn't work as expected without being called from within the `cy.then` callback? I do see this in your conversation with Lachlan that you linked to, but I'm not following the reasoning.

jakarlse88:
> Do you have a situation where the `cy.get('@yourAlias')` doesn't work as expected without being called from within the `cy.then` callback? 

My situation was that I initially wrote some tests that--successfully--used `cy.get('@yourAlias')` _without_ a `then` block, and then ran into the issue described in the conversation when proceeding to test another component. The `then` approach suggested by @lmiller1990 [here](https://github.com/cypress-io/cypress/discussions/25198#discussioncomment-4453793) is the only way I can consistently get all event-related component tests to work. 

I don't really have any reasoning to explain--these changes aren't so much my thinking as just echoing @lmiller1990. Have I misunderstood him?

jakarlse88:
@marktnoonan I made a mistake here, sorry. The original error arose because I passed the spy with the wrong prop name (IE. sans 'on'), but I didn't think to go back and test without `then()` before now. I've done so now and can verify that all my tests work without requiring `then()`. 

I've pushed a commit to revert the change re: `then()`.

content/guides/component-testing/vue/examples.md
```diff
@@ -58,8 +58,11 @@ prop and validate it was called:
 it('clicking + fires a change event with the incremented value', () => {
   const onChangeSpy = cy.spy().as('onChangeSpy')
   cy.mount(Stepper, { props: { onChange: onChangeSpy } })
-  cy.get('[data-cy=increment]').click()
-  cy.get('@onChangeSpy').should('have.been.calledWith', 1)
+  cy.get('[data-cy=increment]')
+    .click()
+    .then(() => {
```
marktnoonan, line 63: Could you revert this as well, and the other two parts of the diff that add `.then()`? Let's keep just the text update to call out that `eventName` emitted from the component becomes `onEventName` in the props when using the component.

marktnoonan (changes requested):
Ah cool, thanks for double checking. There's still 3 places in the diff that have changed examples to add `.then()`, let's remove those and then I'll approve 👍

content/guides/component-testing/vue/examples.md
jakarlse88: Yes, of course. My bad!

marktnoonan (approved):

lmiller1990 (approved):

lmiller1990:
Nice, can you rebase and fix the conflicts @jakarlse88? Then we ship! :boat:

PR #4957, opened 2023-01-03
feat: turning on GA and FS

elylucas:


jaffrepaul (commented):
LGTM. Any way to test this before launching the new site?

elylucas:
> LGTM. Any way to test this before launching the new site?

don't think so, it looks like both sites block requests due to the origin. I'll monitor it closely when it launches to make sure its tracking.

jaffrepaul (approved):

PR #4966, opened 2023-01-09
chore: moving to npm

elylucas:


emilyrohrbough:
just curious - why move from yarn?

elylucas:
> just curious - why move from yarn?

More accessible for those to contribute to the docs that aren't familiar with Yarn.

jaffrepaul (approved):
I added one non-blocking bit that can be removed either here or in a separate PR.

PR #4981, opened 2023-01-12
Post launch tweaks

jaffrepaul:
Handful of minor styling tweaks 🥳

### DIFF
--- a/src/css/custom.scss
+++ b/src/css/custom.scss
@@ -318,10 +318,3 @@ div[class^='codeBlockTitle'] {
 [class^='docItemContainer'] {
   margin: 0 8px;
 }
-
-/* mobile navigation is always in dark mode,
-   so keep active state consistent, regardles
-   of site color mode state */
-.navbar-sidebar .menu__link--active {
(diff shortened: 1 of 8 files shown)


src/css/custom.scss
```diff
@@ -318,3 +318,10 @@ div[class^='codeBlockTitle'] {
 [class^='docItemContainer'] {
   margin: 0 8px;
 }
+
+/* mobile navigation is always in dark mode,
+   so keep active state consistent, regardles
+   of site color mode state */
+.navbar-sidebar .menu__link--active {
```
elylucas, line 325: could we move this to navbar.scss?

elylucas (changes requested):

src/css/custom.scss
jaffrepaul: Sure. I almost did, but paused as it's the `navbar-sidebar` class and we have both navbar & sidebar `.scss` files 😆

elylucas: image

elylucas (approved):

PR #4998, opened 2023-01-24
Fix option links

jphelps92:


emilyrohrbough (approved):
wow! good catch! thank you!

PR #5009, opened 2023-01-27
feat: styling announcement bar and adding one for new auth guides

elylucas:


jaffrepaul (approved):
LGTM. There may be some dialog around the link underline since there isn't one in the figma file. Also, looks like you can add one of the indigo shades to the close button and we'd need that icon exported to match 1:1.

PR #5014, opened 2023-01-28
docs: Add cypress-duration-metrics to plugins list.

archfz:
I've created a new plugin for cypress: [cypress-duration-metrics](https://github.com/archfz/cypress-duration-metrics).

While it has tests, they can only be used manually, so no automated tests, thus cannot add a CI pipeline yet. Given these timings are always random I am yet to figure out a way to effectively test the integration with cypress. Hope this is not an issue for listing it (given contribution guidelines).


archfz:
@emilyrohrbough :smile:

elylucas:
Thanks @archfz, looks like a great plugin.

PR #5015, opened 2023-01-28
docs: fixed command line to run single test file using npm script

leandroandrade:
Hi, this PR fix the command line documentation that shows how to test a single spec file. The `--` is needed to pass parameters to `npm run cy:run`.

Thanks!!!


emilyrohrbough (approved):
nice catch! we use `yarn` alot which will pipe through without the extra `--`.

PR #5017, opened 2023-01-30
.readFile() is now a query

BlueWinds:
Matching code PR: https://github.com/cypress-io/cypress/pull/25595

`.readFile()` will be a query in Cypress 13. This should have minimal impact on most users, but is a breaking change for anyone who was overwritting the command.

### DIFF
--- a/docs/guides/references/migration-guide.mdx
+++ b/docs/guides/references/migration-guide.mdx
@@ -12,6 +12,18 @@ version 13.0.
 
 In Cypress 13, the [`.readFile()`](/api/commands/as) command is now a query. Tests written using it should continue to operate exactly as before; no changes are necessary.
 
+This means that `readFile()` will re-read the file from disk if any upcoming command in the same chain fails. Assertions no longer have to be directly attached.
+
+```js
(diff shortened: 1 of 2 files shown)


docs/guides/references/migration-guide.mdx
```diff
@@ -2,6 +2,68 @@
...
+```
+
+Beginning with Cypress 13, the above test will re-read the file until the file
+exists, it has the requested property, and it passes the assertion.
+
+In previous versions of Cypress, the above command would retry until the file
+existed, but would _not_ re-read it from disk if the file didn't have the
+requested property or the contents didn't match.
+
+#### `.readFile()` can no longer be overwritten with `Cypress.Commands.overwrite()`
```
BlueWinds, line 32: My hope is to have https://github.com/cypress-io/cypress/issues/25078 done before we release Cy13, which would necessitate rewriting this part of the migration guide (to say "actually you can totally still do it, here's how").

chrisbreiding (approved):

mjhenkes (approved):

PR #5021, opened 2023-01-31
Update best-practices.mdx

Eyuel-Nigussie:
A and an are different forms of the indefinite article. Words where the “u” sound is pronounced like “you”, such as user, usual, or utilized use “a” instead of “an.” Since “unique” begins with this “you” sound, you use “a unique” instead of “an unique.”

Source: https://www.scribbr.com/frequently-asked-questions/an-or-a-unique/


docs/guides/references/best-practices.mdx
```diff
@@ -147,7 +147,7 @@ by test code.
 The [Selector Playground](/guides/core-concepts/cypress-app#Selector-Playground)
 automatically follows these best practices.
 
-When determining an unique selector it will automatically prefer elements with:
+When determining a unique selector it will automatically prefer elements with:
```
emilyrohrbough, line 150:
```suggestion
When determining a unique selector, it will automatically prefer elements with:
```

emilyrohrbough (approved):

Issue #4881, opened 2022-11-30
Document `Cypress.ensure`


BlueWinds:
### Subject

API - Cypress.ensure

### Description

In https://github.com/cypress-io/cypress/pull/24697, we created a new public API, `ensure`, on the global Cypress object.  We mentioned this new API in https://github.com/cypress-io/cypress-documentation/pull/4835, but did not fully document it. We should add a page for `Cypress.ensure` same as for all our other exposed APIs.

emilgoldsmith:
Also, the current Custom Queries#Validation section is purely referencing deleted methods that don't work and should reference `Cypress.ensure`, I only figured out the problem by diving into Cypress source code and finding what was used internally

DamienCassou:
This documentation is important.

BlueWinds:
Sorry for leaving this so long; the docs are written up in the linked PR.

PR #5048, opened 2023-02-06
Docs for Cypress.Commands.overwriteQuery

BlueWinds:
- Closes https://github.com/cypress-io/cypress-documentation/issues/4881

Matching code PR: https://github.com/cypress-io/cypress/pull/25674

See that PR for more details.


docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+```
+
+:::caution
+
+Many of these functions accept an optional `onFail` argument. This is a legacy
+feature used to customize the thrown error, and may be removed in a future
+release; we recommend against relying on it. If you need more control over the
+error thrown, write your own `ensure` function instead.
+
+:::
```
emilyrohrbough, line 44: Lets just not even document this if we ever want to remove it.

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
+---
```
emilyrohrbough, line 1: Can we break these docs into their own PR?

docs/api/cypress-api/ensure.mdx
BlueWinds: The reason I documented it here is because people's intellisense / TS will show them that there's an additional argument, and they'll see it if they examine the code.

Undocumented arguments are the worst, so I figure a little warning "this exists but don't use it" would be preferable.

docs/api/cypress-api/ensure.mdx
BlueWinds: https://github.com/cypress-io/cypress-documentation/pull/5054

docs/api/cypress-api/custom-queries.mdx
```diff
@@ -55,6 +56,10 @@ Cypress.Commands.addQuery(name, callbackFn)
 Cypress.Commands.addQuery('getById', function (id) {
   return (subject) => newSubject
 })
+
+Cypress.Commands.addQuery('get', function (originalFn, ...args) {
```
flotwig, line 60:
```suggestion
Cypress.Commands.overwriteQuery('get', function (originalFn, ...args) {
```

docs/api/cypress-api/custom-queries.mdx
```diff
@@ -306,6 +311,70 @@ be executed later (which is not idempotent!). `cy.now()` lets us directly call
...
+`Cypress.Commands.addQuery` - it is a **function that returns a function.** This
+gives you access to both the outer arguments (before you call `originalFn`) and
+the inner function (the return value of `originalFn`), giving you a great deal
+of control over how the query executes.
+
+Remember that query functions rely on `this` - when you invoke `originalFn`, be
+sure to use `.call` or `.apply`.
+
+#### Adding alias support to `.contains()`
+
```
flotwig, line 351: I think it'd be nice to explain this code segment briefly:

```suggestion

In this example, `cy.contains()` is extended to support querying for aliased subjects, like `cy.contains('@foo')`.

```

flotwig (approved):

docs/api/cypress-api/custom-queries.mdx
BlueWinds: Good call.

mjhenkes (approved):

PR #5050, opened 2023-02-06
docs: refresh github integration content w/ latest

elylucas:


pstakoun (approved):

pstakoun:
Note: we'll probably re-add the screenshot comment setting in the near future, but we'll loop back after that happens.

PR #5053, opened 2023-02-07
Restructure API namespace for Cy13

BlueWinds:
Matching code PR: https://github.com/cypress-io/cypress/pull/25738

This PR reorganizes the API docs.
- It moves all commands - query, action and otherwise - under a single namespace and into the same folder, making them easy to locate alphabetically in the sidebar.
- I've also added a new table-of-contents page, which groups commands by category - assertions, queries, non-queries, in logical groups with a description of each command, to help users find what they're looking for when they don't know a command name.
- On each page, I've moved the contents of the "Rules" heading up, to be part of the "Syntax" heading - this puts all the rules of execution (arguments, requirements, assertions, and yielded value) in the same place. Everything you need to know to successfully use a command is up top, rather than split half at the top and half at the bottom of each page.
  - I've updated the "requires being chained off of..." section of each command to make it clearer - many commands *ignore the subject they're given*, but very few *must be chained off `cy`*.
  - For the "Assertions" section of each command, I've clarified what is asserted *before* a command executes and what is asserted *after* it executes, for commands which do both.
  - Some commands were missing ways in which they could time out; I've added all that I'm aware of.
  - I've standardized the wording of the "Yields" section, improving how it explains the query/non-query divide. For commands which "yield the dom elements they've found", I *de*-standardized the language, so the "Yields" sections clearly says what elements the command is locating.


docs/guides/references/migration-guide.mdx
```diff
@@ -29,6 +93,11 @@ In previous versions of Cypress, the above command would retry until the file
 existed, but would _not_ re-read it from disk if the file didn't have the
 requested property or the contents didn't match.
 
+This change also lays the groundwork for (freely chaining Cypress
+commands)[https://github.com/cypress-io/cypress/issues/25655] and eventually
```
flotwig, line 97:
```suggestion
This change also lays the groundwork for [freely chaining Cypress
commands](https://github.com/cypress-io/cypress/issues/25655) and eventually
```

BlueWinds: Doh

docs/api/commands/getallcookies.mdx
```diff
@@ -80,8 +78,7 @@ cy.getAllCookies()
 
 ### Assertions [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Assertions)
 
-- `cy.getAllCookies()` will only run assertions you have chained once, and will
-  not [retry](/guides/core-concepts/retry-ability).
+- `cy.getAllCookies()` has no implicit assertions.
```
emilyrohrbough, line 81:
```suggestion
- `cy.getAllCookies()` has no implicit assertions.
- `cy.getAllLocalStorage()`
  [only reads the data once](/guides/core-concepts/retry-ability#Only-queries-are-retried), so any assertions chaining the command will not retry because the yielded value is static.
```

docs/api/commands/exec.mdx
```diff
@@ -200,8 +201,8 @@ describe('has data available from database', { execTimeout: 90000 }, () => {
 
 ### Assertions [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Assertions)
 
-- `cy.exec()` will only run assertions you have chained once, and will not
-  [retry](/guides/core-concepts/retry-ability).
+- `cy.exec()` asserts that the command exits successfully. If the command exits
+  with a non-zero exit code, `cy.exec()` will fail the test.
```
emilyrohrbough, line 205: if this is the case, why does it yield stderr? Users can assert of use that value?

docs/api/commands/each.mdx
```diff
@@ -116,8 +116,7 @@ function.
 
 ### Assertions [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Assertions)
 
-- `.each()` will only run assertions you have chained once, and will not
-  [retry](/guides/core-concepts/retry-ability).
+- `.each()` has no implicit assertions.
```
emilyrohrbough, line 119: I only know what this means because I have review our code so often. I do not believe our users will understand what this truly means. We need to to provide more context here.

I think only calling out when a Cypress command performs an internal assertion makes the most sense.

Also when I read these docs, I am more interested in understanding how assertions or if assertion retry for a command chain. In the case of `.each()`, assertions are not retried because .each() is not a query command. It's yielded value is static so re-trying the assertions will not change the result.

Something like this might be better:
```
- When the `.each()` command runs, Cypress performs chained assertions once. because `cy.exec()` is not a query and the
  [only executes the command once](/guides/core-concepts/retry-ability#Only-queries-are-retried), the yielded value is static so retrying the assertion does not change the assertion result.
```

docs/api/commands/fixture.mdx
```diff
@@ -73,7 +73,8 @@ Pass in an options object to change the default behavior of `cy.fixture()`.
 
 - `cy.fixture()` yields the contents of the file. Formatting is determined by
   its file extension.
-- The yielded subject is **not** updated if the contents change on disk.
+- `cy.fixture()` is not a query. It
```
emilyrohrbough, line 76: Based on our definition of a query:
> Most commands are not retried because they could potentially change the state of the application under test
theres really no reason .fixture() isn't query command. It's "querying" the file for data

docs/api/assertions/_category_.json
```diff
@@ -1,4 +0,0 @@
-{
-  "label": "Assertions",
```
emilyrohrbough, line 2: I still believe we need to have a seperate Assertions Section in the API docs to guide for easy reference by users. Writing tests are really composes of:

1. query something
2. do something (action - maybe)
3. assert state

Assertions are fundamental to the validity of the test and should be easy to find - regardless of the implementation details under the hood.

emilyrohrbough (changes requested):
Left some comments.

docs/guides/references/migration-guide.mdx
```diff
@@ -8,15 +8,79 @@ This guide details the changes and how to change your code to migrate to Cypress
...
+
+Assertions being queries allows for extremely flexible and fluent chaining of
+Cypress commands. See
+[#25134](https://github.com/cypress-io/cypress/issues/25134) for an example.
+This change also lays the groundwork for (freely chaining Cypress
+commands)[https://github.com/cypress-io/cypress/issues/25655] and eventually
+native [`async / await`](https://github.com/cypress-io/cypress/issues/1417)
+support.
+
+#### Assertions now retry after _all_ commands.
```
emilyrohrbough, line 27: This isn't true. There are several Cypress commands that only execute once.

emilyrohrbough: okay - mis-read this, however,
> However, assertions now execute
independently of the previous command, meaning they always retry.

This needs more context to how it behaved previously and how this is influenced by query commands.

docs/api/commands/each.mdx
emilyrohrbough: Right - this is changing so assertions always retry - however this isn't clear how it ties to the subject being "updated" with queries vs non-query commands.

docs/api/commands/exec.mdx
BlueWinds: Many commands write to stderr even when successful. `curl` is one example.

https://askubuntu.com/questions/944788/how-does-curl-print-to-terminal-while-piping

> When you are using curl to open an URL, you'll get two output:
>
>     The status of the curl itself.
>     The contents of that URL.
>
> Curl should use a way to show these two separately otherwise processing of the real output (URL's content) would be hard and I'll end up with unnecessary contents (curl's status).
>
> So it uses stderr for its status and stdout for the content.

You wouldn't usually `curl` with `cy.exec`, but it's a common pattern in linux commands.

docs/api/assertions/_category_.json
BlueWinds: @flotwig had a good point in slack earlier:

> So since we've split the "commands" in the docs into "Assertions", "Actions", "Queries", and "Other Commands", I've found it kinda difficult to find the command I'm looking for. I'm sure I'll get the hang of it eventually but it's more difficult than searching through a single a-z list.
> Was there any discussion about keeping them in a single a-z list and using badges to denote the types? like

```
and [assert]
click [action]
debug [query]
eq [query]
type [action]
reload
```

I agree with your point that there's probably some need to have assertions called out right at the top - they're very important. Wondering how we want to structure this section as a whole though, I think breaking up the commands by type may not be quite the right approach.

Maybe we should make a new index page which breaks up commands by type, and then leave the sidebar entirely alphabetical / undifferentiated?

docs/api/commands/each.mdx
BlueWinds: Hm. I didn't like this at first, but the more I read it, the more I think you're right, and reitterating how queries / non-queries interact in the `### Assertions` section of each page is the right approach.

docs/api/commands/fixture.mdx
BlueWinds: I'd like to leave the language somewhat standardized. How about 

> `cy.fixture()` is not a query. It [only reads the data once](/guides/core-concepts/retry-ability#Only-queries-are-retried), If you want a command that re-reads the file until chained assertions pass, use [`cy.readFile()`](/api/commands/readfile) instead.

docs/api/assertions/_category_.json
BlueWinds: That can get us the best of both worlds - users click on API and are presented with a useful index, showing things they need to learn, but then also there's easy access to 'where are the docs on command X' scrolling down an alphabetical list.

BlueWinds:
Pushed up a reorganization of the `/api` namespace, including adding a new table of contents.

I still have more feedback to address, but wanted to push up this much for people to look at.

docs/guides/references/migration-guide.mdx
BlueWinds: Added more text explaining how it used to behave and how it behaves now, including an additional example.

docs/api/assertions/_category_.json
BlueWinds: Based on slack discussions, I've redone the layout of the API section of the docs. Give it a look and let me know what you think.

docs/api/commands/writefile.mdx
```diff
@@ -213,8 +213,7 @@ cy.readFile(filename, null).then((obj) => {
 
 ### Assertions [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Assertions)
 
-- `cy.writeFile()` will only run assertions you have chained once, and will not
-  [retry](/guides/core-concepts/retry-ability).
+- `cy.writeFile()` has no implicit assertions.
```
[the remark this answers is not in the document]
BlueWinds: writeFile yields `null`, and makes no implicit assertions about the subject it yields.

In my mind, and implicit assertion isn't "any way a command could fail" - it's "things Cypress already checks about the subject". You never need to write ".should('exist')" about subjects, because Cypress implicitly asserts that they exist. In this case the subject is `null`, and there are no implicit assertions that you could make explicit about it.

BlueWinds:
I've moved some sections of this to their own PRs, to reduce the size of this one.

- https://github.com/cypress-io/cypress-documentation/pull/5080 contains some improvements in the assertions reference that aren't specific to Cy 13 and can be merged in separately.
- https://github.com/cypress-io/cypress-documentation/pull/5081 contains changes to the "guides" folder for Cy 13.

docs/api/assertions/_category_.json
flotwig: Could you summarize the changes you made to the layout in the PR comment? It's kinda hard to tell what's going on by looking at the files changed

BlueWinds: I should have done this before sending back to review. Updated PR description with more detail now that the scope has changed.

docs/api/commands/each.mdx
BlueWinds: Have done this in subsequent revisions of the PR; going to mark this conversation resolved, and we can discuss more on a fresh version of the copy if needed.

docs/api/table-of-contents.mdx
```diff
@@ -0,0 +1,198 @@
...
+details.
+
+### Additional APIs
+
+In addition to the above APIs, Cypress exposes a number of properties and
+methods on the global `Cypress` object.
+
+| Property                                                                  | Usage                                                                                                                          |
+| ------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| [`.Cypress.arch`](/api/cypress-api/arch)                                  | CPU architecture name of the underlying OS, as returned from Node's `os.arch()`.                                               |
```
chrisbreiding, line 164: Why are all these `Cypress.*` APIs prepended with a period?

docs/api/table-of-contents.mdx
```diff
@@ -0,0 +1,198 @@
...
+### Events
+
+Cypress emits a series of events as it runs in your browser. These events are
+useful not only to control your application's behavior, but also for debugging
+purposes. Learn more about them in the
+[Catalog of Events](/api/cypress-api/catalog-of-events).
+
+### Custom Commands
+
+In addition to the builtin commands, the Cypress community has created a large
```
chrisbreiding, line 150:
```suggestion
In addition to the built-in commands, the Cypress community has created a large
```

docs/api/table-of-contents.mdx
```diff
@@ -0,0 +1,198 @@
...
+| [`.Cypress.$`](/api/utilities/$)                 | Cypress automatically includes [jQuery](http://jquery.com/) and exposes it as `Cypress.$`.                                               |
+| [`.Cypress.Blob`](/api/utilities/blob)           | Cypress automatically includes a [Blob](https://github.com/nolanlawson/blob-util) library and exposes it as `Cypress.Blob`.              |
+| [`.Cypress.Buffer`](/api/utilities/buffer)       | Cypress automatically includes a [Buffer polyfill](https://github.com/feross/buffer) for the browser and exposes it as `Cypress.Buffer`. |
+| [`.Cypress._`](/api/utilities/_)                 | Cypress automatically includes [lodash](https://lodash.com/) and exposes it as `Cypress._`.                                              |
+| [`.Cypress.minimatch`](/api/utilities/minimatch) | Cypress automatically includes [minimatch](https://github.com/isaacs/minimatch) and exposes it as `Cypress.minimatch`.                   |
+| [`.Cypress.promise`](/api/utilities/promise)     | Cypress automatically includes `Bluebird` and exposes it as `Cypress.Promise`.                                                           |
```
chrisbreiding, line 197:
```suggestion
| [`.Cypress.Promise`](/api/utilities/promise)     | Cypress automatically includes [Bluebird](http://bluebirdjs.com) and exposes it as `Cypress.Promise`.                                                           |
```

BlueWinds: Done.

docs/api/table-of-contents.mdx
```diff
@@ -0,0 +1,198 @@
...
+| [`.clearCookie()`](/api/commands/clearcookie)                       | Clear a specific browser cookie.                                                                                                                                                |
+| [`.clearCookies()`](/api/commands/clearcookies)                     | Clear browser cookies for a domain.                                                                                                                                             |
+| [`.clearLocalStorage()`](/api/commands/clearlocalstorage)           | Clear data in `localStorage` for current domain and subdomain.                                                                                                                  |
+| [`.clock()`](/api/commands/clock)                                   | Override the browser's builtin time objects (`Date`, `setTimeout` and similar), allowing them to be controlled via `cy.tick()`.                                                 |
```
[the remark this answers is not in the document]
BlueWinds: Done.

docs/api/table-of-contents.mdx
BlueWinds: Done.

chrisbreiding (approved):

docs/api/cypress-api/catalog-of-events.mdx
```diff
@@ -1,5 +1,6 @@
 ---
 title: Catalog of Events
+sidebar_position: 10
```
emilyrohrbough, line 3: I would expect events to be its own document / api navigation section. I would not expect to find these under the API section.

docs/api/commands/writefile.mdx
```diff
@@ -72,6 +72,21 @@ parameter and include encoding there. This is the same behavior as
...
 
+### Requirements [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Chains-of-Commands)
+
+- `cy.writeFile()` ignores the subject it is given.
+- `cy.writeFile()` requires the file be successfully written to disk. Anything
+  preventing this such as OS permission issues will cause it to fail.
+
+### Assertions [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Assertions)
+
+- `cy.writeFile()` has no implicit assertions.
```
emilyrohrbough, line 83: lets link to the implicit assertion documentation for users unware of what this means

docs/api/commands/writefile.mdx
```diff
@@ -72,6 +72,21 @@ parameter and include encoding there. This is the same behavior as
...
+
+### Timeouts [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Timeouts)
+
+- `cy.writeFile()` can time out when encoding or writing the file to disk takes
+  a significant amount of time.
```
emilyrohrbough, line 88:
```suggestion
- `cy.writeFile()` can time out when encoding or writing the file to disk.
```

docs/api/commands/uncheck.mdx
```diff
@@ -57,11 +57,28 @@ Pass in an options object to change the default behavior of `.uncheck()`.
 | `timeout`                    | [`defaultCommandTimeout`](/guides/references/configuration#Timeouts)           | Time to wait for `.uncheck()` to resolve before [timing out](#Timeouts)                                                                            |
 | `waitForAnimations`          | [`waitForAnimations`](/guides/references/configuration#Actionability)          | Whether to wait for elements to [finish animating](/guides/core-concepts/interacting-with-elements#Animations) before executing the command.       |
 
+### Requirements [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Chains-of-Commands)
```
emilyrohrbough, line 60: I don't mind how this reads, but I do not believe the section `Syntax` is the correct heading for these concepts since this is more than the syntax at this time.

docs/api/commands/window.mdx
```diff
@@ -31,10 +31,26 @@ Pass in an options object to change the default behavior of `cy.window()`.
...
+- `cy.window()` has no implicit assertions.
+
+### Timeouts [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Timeouts)
+
+- `cy.window()` cannot time out. The `timeout` option is only useful to set the
+  timeout of following assertions.
+
 ### Yields [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Subject-Management)
 
+- `cy.window()` is a query command, and is
```
emilyrohrbough, line 49: I'd expect this to be the second bullet after we say what it yields

docs/api/commands/wait.mdx
```diff
@@ -60,18 +60,31 @@ Pass in an options object to change the default behavior of `cy.wait()`.
...
 
-#### When given an `alias` argument:
+- `cy.wait()` has no implicit assertions.
 
-- `cy.wait()` 'yields an object containing the HTTP request and response
-  properties of the request.
+### Timeouts [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Timeouts)
+
+- When given a `time` argument, `cy.wait()` cannot time out.
+- When given an alias algument, `cy.wait()` can time out waiting for the request
```
emilyrohrbough, line 75:
```suggestion
- When given an alias argument, `cy.wait()` can time out waiting for the request
```

docs/api/commands/wait.mdx
```diff
@@ -60,18 +60,31 @@ Pass in an options object to change the default behavior of `cy.wait()`.
...
+  to go out.
+- When given an alias algument, `cy.wait()` can time out waiting for the
+  response to return.
```
emilyrohrbough, line 78:
```suggestion
  to go out or the
  response to return.
```

docs/api/commands/uncheck.mdx
```diff
@@ -57,11 +57,28 @@ Pass in an options object to change the default behavior of `.uncheck()`.
...
+
+- `.uncheck()` requires being chained off a command that yields DOM element(s).
+- `.uncheck()` requires the element to have type `checkbox` or `radio`.
+
+### Assertions [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Assertions)
+
+- Before executing, `.uncheck()` will wait for the element to reach an
```
emilyrohrbough, line 67:
```suggestion
- Before executing, `.uncheck()` will wait for the subject to reach an
```

emilyrohrbough (changes requested):
A few things:
1- can we update this PR title to reflect these changes.
2- I initially only looked at the doc preview since this PR has 111 file changes. with the restructuring of commands layout, I couldn't/still can't determine which commands are assertions or query commands when clicking through the list of commands. I didn't really realize the table of contents was trying to tell this story. Can we add a badge on the command pages to quickly highlight this?

BlueWinds:
> A few things:
> 1- can we update this PR title to reflect these changes.

Done.

> 2- I initially only looked at the doc preview since this PR has 111 file changes. with the restructuring of commands layout, I couldn't/still can't determine which commands are assertions or query commands when clicking through the list of commands. I didn't really realize the table of contents was trying to tell this story. Can we add a badge on the command pages to quickly highlight this?

A badge is possible, I'm just not sure the benefit. When visiting a command page, do you need to tell at a glance if a command is a query or not? I put it in the `### Yields` section of each command because that's where I felt it was relevant, when chaining commands together.

docs/api/table-of-contents.mdx
```diff
@@ -0,0 +1,198 @@
+---
+title: Table of Contents
+sidebar_position: 10
+---
+
```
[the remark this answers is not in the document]
BlueWinds: Added them.

docs/guides/guides/debugging.mdx
```diff
@@ -246,7 +246,7 @@ by [Cypress Ambassador](https://www.cypress.io/ambassadors/) Josh Justice.
 ## Log Cypress events
 
 Cypress emits multiple events you can listen to as shown below.
-[Read more about logging events in the browser here](/api/events/catalog-of-events#Logging-All-Events).
+[Read more about logging events in the browser here](/api/cypress-api/catalog-of-events#Logging-All-Events).
```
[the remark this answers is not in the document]
BlueWinds: That on-link is pointing to a different sort of event, where developers meet up to discuss things. https://www.cypress.io/events/

The slug here looks to be correct to me?

docs/api/cypress-api/catalog-of-events.mdx
BlueWinds: I spent some time playing around with the layout of this section, and did a little bit of reorganization.

image

This based on the fact that I really like the way this looks in the table of contents:

image

I think it flows very naturally like this.

docs/faq/questions/using-cypress-faq.mdx
```diff
@@ -411,7 +411,7 @@ to either:
 - Prevent Cypress from failing the test
 
 This is documented in detail on the
-[Catalog Of Events](/api/events/catalog-of-events) page and the recipe
+[Catalog Of Events](/api/cypress-api/catalog-of-events) page and the recipe
```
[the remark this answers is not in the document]
BlueWinds: I think that it makes sense to roll it out all at once with Cy 13, so that we don't have to create multiple versions of the same changes (eg, create an "assertions" section of the new table of contents, then remove it in another PR on another branch). Reorganizing things piecemeal can result in a disjointed user experience; I'd rather have all the changes living in one branch so we can look at the organization holistically.

docs/api/commands/writefile.mdx
BlueWinds: Added links to each command page.

I added the link to `/guides/core-concepts/introduction-to-cypress#Implicit-Assertions` - in this branch the URL is `/guides/core-concepts/introduction-to-cypress#Default-Assertions`, but I'm renaming the section `Implicit Assertions` in https://github.com/cypress-io/cypress-documentation/pull/5081/files, so using the new name here to avoid extra work later.

docs/api/commands/writefile.mdx
BlueWinds: Applied a similar change to reduce the wordiness.

docs/api/commands/uncheck.mdx
BlueWinds: I haven't committed these changes / applied them to every command yet because it'd involve manually moving around each file for a direction I'm not 100% sure on yet.

docs/api/commands/window.mdx
BlueWinds: So the reason I put the "is a query command" first is that some commands have long `yields` sections. Eg:

> image

------

I was aiming to always have it in the same place (eg, first or second), and in the cases where we need to spend time explaining what's yielded in more detail, having it second ended up making it feel buried under the other text.

> image

-------

I'd be fine breaking the consistency here, having it sometimes below what's actually yielded (you're right it is better in the `short case`) and sometimes above (I do think it needs to come first in the 'long case'). Thoughts?

docs/api/commands/wait.mdx
BlueWinds: Applied, and to the next line as well. :+1:

docs/api/commands/wait.mdx
BlueWinds: I broke this out into two sentences because you can specify different request and response timeouts - it seemed like two different timeout options = two different ways it can time out.

docs/api/commands/uncheck.mdx
BlueWinds: Why would we use a more generic term here? We know it's an element(s), rather than any generic "subject" (this is also detailed above as one of the 'requirements' of the command).

docs/api/commands/uncheck.mdx
BlueWinds: Applied it to the first three commands: .and(), .as(), and .blur(), so people can see it 'live'. If we like the direction I'll go through and apply it to all commands.

AtofStryker:
closing as this work is unplanned for the v13 release

PR #5054, opened 2023-02-07
Documentation for Cypress.ensure

BlueWinds:
- Closes https://github.com/cypress-io/cypress-documentation/issues/4881

This part of our API was added in Cypress 12, but never documented properly. Pulled the changes out from https://github.com/cypress-io/cypress-documentation/pull/5048 into its own PR.


docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
+---
+title: Cypress.ensure
+---
+
+`Cypress.ensure.method()` is a collection of helper methods for verifying
+conditions while executing Cypress commands. They are mostly useful when writing
```
flotwig, line 6:
```suggestion
`Cypress.ensure` is a collection of helper methods for making assertions. They are mostly useful when writing
```

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+[custom queries](/api/cypress-api/custom-queries) or
+[commands](/api/cypress-api/custom-commands).
+
+Most functions on `Cypress.ensure` accept `subject` argument, and assert
+something about it, throwing an error if it doesn't match. They have no return
+value; their only purpose is to ensure the state of the application and throw an
+error if it doesn't match.
```
flotwig, line 13: Feels like this could be less verbose, here's a stab at it:

```suggestion
Most functions on `Cypress.ensure` accept a `subject` argument, check an assertion, and throw an error if the assertion fails. These functions have no return value.
```

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+:::caution
+
+Many of these functions accept an optional `onFail` argument. This is a legacy
+feature used to customize the thrown error, and may be removed in a future
+release; we recommend against relying on it. If you need more control over the
+error thrown, write your own `ensure` function instead.
+
+:::
+
```
flotwig, line 45: IMO we should just not document this, since the user can just `try...catch` to get the same functionality as `onFail`, right? Then we can just rip it out when we're done using it internally.

```suggestion
```

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+    return subject
+  }
+})
+```
+
+## See also
+
+- Writing [custom queries](/api/cypress-api/custom-queries) contains more
+  information about writing custom queries, which is the main use-case for the
+  `ensure` functions.
```
flotwig, line 81:
```suggestion
- ["Custom Queries"](/api/cypress-api/custom-queries) contains more
  information about writing custom queries, which is the main use-case for the
  `ensure` functions.
```

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
+---
+title: Cypress.ensure
+---
+
+`Cypress.ensure.method()` is a collection of helper methods for verifying
+conditions while executing Cypress commands. They are mostly useful when writing
+[custom queries](/api/cypress-api/custom-queries) or
+[commands](/api/cypress-api/custom-commands).
```
emilyrohrbough, line 8:
```suggestion
[custom commands](/api/cypress-api/custom-commands).
```

emilyrohrbough: or remove custom from the query link if you'd prefer `custom [queries](link) or [commands](link)`

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+Most functions on `Cypress.ensure` accept `subject` argument, and assert
```
emilyrohrbough, line 10:
```suggestion
Most methods on `Cypress.ensure` accept `subject` argument to assert
```

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+something about it, throwing an error if it doesn't match. They have no return
```
emilyrohrbough, line 11:
```suggestion
something about it; if the assertion fails, an error is thrown. They have no return
```

docs/api/cypress-api/ensure.mdx
emilyrohrbough: or this!

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+// State of DOM element
+Cypress.ensure.isAttached(subject, name, cy)​
+Cypress.ensure.isNotDisabled(subject, name)​
+Cypress.ensure.isNotHiddenByAncestors(subject, name)​
+Cypress.ensure.isNotReadonly(subject, name)​
+Cypress.ensure.isScrollable($el, name)​
+Cypress.ensure.isStrictlyVisible(subject, name)​
+Cypress.ensure.isVisible(subject, name)​
+```
```
emilyrohrbough, line 35: Why are most of the referenced as "subject" vs "$el"?

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+error if it doesn't match.
+
+## Syntax
+
+```javascript
+// Command chain
+Cypress.ensure.isChildCommand(command, args, cy)​
+
+// Type of argument
+Cypress.ensure.isType(subject, type, name, cy)​
```
emilyrohrbough, line 22: I can't remember, but do these expect subject to be anything? We don't reference subject must in the docs and I'm thinking it might be helpful to expand above on what a subject might be for users who aren't familiar

docs/api/cypress-api/ensure.mdx
emilyrohrbough: I gave a similar suggestion to not document this https://github.com/cypress-io/cypress-documentation/pull/5048#discussion_r1099138603. The onFail ties in to how this would be removed up if this callback matched the command onFail behavior, so it's require quite a bit of user-knowledge to handle this correctly. I like @flotwig's suggestion for handling with try-catch.

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+    Cypress.ensure.isType(subject, ['element', 'document', 'window'], 'getChildById', cy)
+
+    return $$(`#${id}`, subject)
+  }
+})
+
+Cypress.Commands.addQuery('verifyElementActionable', function (id) {
+  // Throws an early error if this query is chained directly off `cy`, eg.
+  // `cy.verifyElementActionable('foo')` would throw an error
+  Cypress.ensure.isChildCommand(command, args, cy)​
```
emilyrohrbough, line 63: [Here](https://github.com/cypress-io/cypress/pull/25674#discussion_r1093771328) you said query's don't have a concept of parent/child commands. Is this a realistic example? Also where is ` command`, `args`, and `cy` coming form?

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+Cypress.ensure.isChildCommand(command, args, cy)​
+
+// Type of argument
+Cypress.ensure.isType(subject, type, name, cy)​
+Cypress.ensure.isElement(subject, name, cy)​
+Cypress.ensure.isWindow(subject, name, cy)
+Cypress.ensure.isDocument(subject, name, cy)​
+
+// State of DOM element
+Cypress.ensure.isAttached(subject, name, cy)​
```
emilyrohrbough, line 28: Can we expand that the name arg should be the custom query/command name to provide meaningful error messages?
```suggestion
Cypress.ensure.isAttached(subject, customCommandName, cy)​
```

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+```javascript
+Cypress.Commands.addQuery('getChildById', function (id) {
+  return (subject) => {
+    // Verify that the subject is an element, document, or window object
+    Cypress.ensure.isType(subject, ['element', 'document', 'window'], 'getChildById', cy)
+
+    return $$(`#${id}`, subject)
+  }
+})
+
```
emilyrohrbough, line 59:
```suggestion
const queryName = 'verifyElementActionable'
```

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
...
+Cypress.Commands.addQuery('verifyElementActionable', function (id) {
+  // Throws an early error if this query is chained directly off `cy`, eg.
+  // `cy.verifyElementActionable('foo')` would throw an error
+  Cypress.ensure.isChildCommand(command, args, cy)​
+
+  return (subject) => {
+    // Verify that the subject fulfills a variety of conditions
+    Cypress.ensure.isElement(subject, 'verifyElementActionable', cy)
```
emilyrohrbough, line 67: I initially thought this was a copy & paste error when they all used the same string until I looked at the Cypress code
```suggestion
    Cypress.ensure.isElement(subject, queryName, cy)
```

docs/api/cypress-api/ensure.mdx
BlueWinds: > The reason I documented it here is because people's intellisense / TS will show them that there's an additional argument, and they'll see it if they examine the code.

> Undocumented arguments are the worst, so I figure a little warning "this exists but don't use it" would be preferable.

Was the response I gave to Emily, and I still feel the same way - if it exists, we should not leave it silently undocumented. I've found myself in this boat several times dealing with other projects, where I see an argument is there, but the documentation is completely silent. What does it do? Why is it there? Your own code uses it, to what purpose?

I think a warning "it exists, but don't use it" is appropriate, as well as not including it in the documented function signatures.

docs/api/cypress-api/ensure.mdx
BlueWinds: We reference 'subject' all over the docs.

https://docs.cypress.io/guides/core-concepts/introduction-to-cypress
> How Cypress manages subjects and chains of commands

The very first paragraph of the 'core concepts' guide discusses them.

> It's very important to understand the mechanism Cypress uses to chain commands together. It manages a Promise chain on your behalf, with each command yielding a 'subject' to the next command

https://docs.cypress.io/api/cypress-api/custom-commands#Arguments
> prevSubject | Boolean, String or Array | false | how to handle the previously yielded subject.
> In addition to controlling the command's implicit behavior you can also add declarative subject validations such as:
>    element: requires the previous subject be a DOM element
>    document: requires the previous subject be the document
>    window: requires the previous subject be the window

and all over that page.

docs/api/cypress-api/ensure.mdx
BlueWinds: `subject` is what they'll see in intellesense, it's how these functions are used in the Cypress repo itself (), and the function signature of a query is `(subject: Any) => Any`.

The primary use of these is to call them on a `subject`, to ensure the subject matches the conditions, so I think they should be documented that way. `$el` is extremely non-descriptive in comparison, and doesn't indicate anything unless you're familiar with our convention that '$' means "A jquery object".

BlueWinds: Changed the one remaining over to `subject` so they're all consistent.

docs/api/cypress-api/ensure.mdx
BlueWinds: Applied to all entries.

docs/api/cypress-api/ensure.mdx
BlueWinds: Done.

docs/api/cypress-api/ensure.mdx
BlueWinds: Fixed the example to show where they come from.

Yes it's a realistic example; it's exactly how we use this function in our own code. I'll update that other file to be clear that queries don't *by default* have a concept of being child / parent - you can make them have one by using `ensure.isChildCommand()`.

docs/api/cypress-api/ensure.mdx
BlueWinds: Applied to all lines.

docs/api/cypress-api/ensure.mdx
BlueWinds: Applied Zach's suggestion to redo these lines.

docs/api/cypress-api/ensure.mdx
BlueWinds: Applied Zach's suggestion to redo these lines.

flotwig (approved):

docs/api/cypress-api/ensure.mdx
```diff
@@ -0,0 +1,81 @@
+---
+title: Cypress.ensure
+---
+
+`Cypress.ensure` is a collection of helper methods for making assertions. They
+are mostly useful when writing [custom queries](/api/cypress-api/custom-queries)
+or [custom commands](/api/cypress-api/custom-commands).
+
+Most functions on `Cypress.ensure` accept a `subject` argument, check an
```
emilyrohrbough, line 9:
```suggestion
Most functions on `Cypress.ensure` accept a [`subject`](/guides/core-concepts/introduction-to-cypress#Subject-Management) argument, check an
```

docs/api/cypress-api/ensure.mdx
emilyrohrbough: >  It's just not baked into the query API the way it's baked into the command API.

Gottchya - when you said it didn't have the concept of it, I had understood it as "parent" / "child" truely was no longer a concept vs a difference in how the concept was implemented.

emilyrohrbough (approved):

docs/api/cypress-api/ensure.mdx
BlueWinds: tl;dr - it *shouldn't* still be a concept, but it leaked into queries through a new implementation. :P

docs/api/cypress-api/ensure.mdx
BlueWinds: Applied.

PR #5056, opened 2023-02-08
content: Mention Cy Migrator in Angular Migration docs

marktnoonan:
Just noticed that we link this tool in the footer but do not mention it on this page where it's relevant.


docs/guides/end-to-end-testing/migration/protractor-to-cypress.mdx
```diff
@@ -65,6 +65,8 @@ describe('Authorization Tests', () => {
 })
 ```
 
+To see how this conversion would work with some of your own test code, you can paste it into the interactive [Cypress Migrator tool](https://migrator.cypress.io/), which will generate the equivalent Cypress code.
```
emilyrohrbough, line 68: IMO this should be at the top of the page in an alert banner! And possible referenced here as well. This is gold.

marktnoonan: Alert banner is a good idea. I'll tweak this some more when we have a chance. Also might be a good opportunity to mention Angular Component Testing.

jordanpowell88 (commented):
I feel like a screenshot or video could be really valuable here to so people can see  a preview of the migrator

PR #5069, opened 2023-02-16
CT Public API docs

lmiller1990:
closes https://github.com/cypress-io/cypress/issues/25800

Docs for the new Framework Definitions feature. Epic: https://github.com/cypress-io/cypress/issues/25637 (linked to all relevant issues).


docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+}
+
+/**
+ * @param {() => JSX.Element} - component to render
+ */
+export function mount(component, options = {}) {
+  const root = getContainerEl();
+  dispose = render(() => component, root);
+
+  return cy.wait(0, { log: false }).then(() => {
```
mike-plummer, line 183: Might be good to explain what's going on with the `cy.wait` - the async render workaround probably isn't obvious to most people

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+
+function cleanup() {
+  dispose?.();
+}
+
+/**
+ * @param {() => JSX.Element} - component to render
+ */
+export function mount(component, options = {}) {
+  const root = getContainerEl();
```
mike-plummer, line 180: // Retrieve root DOM element that Cypress has prepared for this test

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+  });
+}
+
+// Cleanup after each test
+setupHooks(cleanup);
+```
+
+This is different for each library, but the concept is the same - identify how to mount or render a component in your library, and implement it in a function named `mount` in `index.mjs`.
+
+When a user configures Component Testing with your Framework Definition, their `supportFile` will contain the correct `mount` function:
```
mike-plummer, line 199:
```suggestion
When a user configures Component Testing with your Framework Definition we automatically configure a `cy.mount` command using your `mount` function in the Component Testing `supportFile`:
```

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+import { mount } from '@lmiller1990/cypress-ct-solid-js'
+
+Cypress.Commands.add('mount', mount)
+```
+
+## package.json
+
+The final thing you need is a correctly configured `package.json`. This example was created using [our official template](https://github.com/cypress-io/cypress-ct-definition-template).
+
+The main things to note is the `name` property, which follows the `@org/cypress-ct-*` convention, and is the `exports` property. The `node` entry points to the Framework Definition, and the `default` entry points to the Mount Adapter:
```
mike-plummer, line 211:
```suggestion
The final thing you need is a correctly configured `package.json`. This example was created using [our official template](https://github.com/cypress-io/cypress-ct-definition-template).

There are two fields of note:
* `name`: Your package name which must follow the `cypress-ct-*`/`@org/cypress-ct-*` convention
* `exports`: Object referencing the two files we created to comprise the Framework Definition. The `node` entry points to the Definition file, and the `default` entry points to the Mount Adapter:
```

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+  },
+  "devDependencies": {
+    "solid-js": "^1.6.0"
+  },
+  "peerDependencies": {
+    "solid-js": "^1.6.0"
+  }
+}
+```
+
```
mike-plummer, line 238: Worth adding a blurb around writing tests? Even if we don't have examples/content might be nice to at least suggest it as a good idea

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+    "index.mjs"
+  ],
+  "dependencies": {
+    "@cypress/mount-utils": "^4.0.0"
+  },
+  "devDependencies": {
+    "solid-js": "^1.6.0"
+  },
+  "peerDependencies": {
+    "solid-js": "^1.6.0"
```
mike-plummer, line 234: I wonder if we should suggest adding a `cypress` `peerDependency` to restrict what versions of Cypress a plugin has been validated against? Probably an extra layer of complexity that's unnecessary here

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+}
+```
+
+## Publishing on npm
+
+That's it! Publish your Framework Definition on npm and start using it.
+
+## List of Framework Definitions
+
+If you want to share your Framework Definition with us, please send a Pull Request - we would be happy to feature a link to your package!
```
mike-plummer, line 245:
```suggestion
Once your Framework Definition is published and ready for use we would be delighted to mention it in our documentation so other Cypress users on the same framework can find it. [Submit a Pull Request](https://github.com/cypress-io/cypress-documentation/blob/main/CONTRIBUTING.md)!
```

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
+# Integrating with Cypress Component Testing
+
+:::info
+This feature is only available from Cypress 12.7.
+:::
+
+
+Cypress ships official support for many popular libraries and frameworks, such as [React](/guides/component-testing/react/overview) and [Angular](/guides/component-testing/angular/overview). We constantly monitor what libraries developers are interested in and are always open to adding new ones to Cypress.
+
+All officially supported libraries feature a first class onboarding experience, where we detect and scaffold the correct files, and a framework-specific `mount` adapter, to render your components. We call this collection of features a *Framework Definition*, since it defines the requirements for a library or framework to work in Cypress.
```
mike-plummer, line 10:
```suggestion
Cypress Component Testing includes [official support](/guides/component-testing/overview#Supported-Frameworks) for many popular libraries and frameworks such as [React](/guides/component-testing/react/overview), [Angular](/guides/component-testing/angular/overview), and [Vue](/guides/component-testing/vue/overview). All officially supported libraries feature a first class onboarding experience, where we detect and scaffold the correct files, and a framework-specific `mount` adapter to render your components. We call this collection of features a *Framework Definition* since it defines the requirements for a library or framework to work in Cypress.
```

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+
+If your favorite library isn't featured, don't worry - we expose the same API we use internally for third parties to define their own Framework Definitions.
```
mike-plummer, line 12:
```suggestion
If your favorite library isn't featured, don't worry - we expose the same API we use internally for the Cypress community to define their own Framework Definitions.
```

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+  `,
+});
+```
+
+Our Framework Definition shows up in Cypress!
```
mike-plummer, line 145: Worth mentioning that their entry will be marked as "Community" in the UI?

mike-plummer (changes requested):
Couple wording suggestions and questions, but overall this is awesome content!

One total nitpick - the screenshots have 12.6.0 but this feature is marked as 12.7+

elylucas:
We should probably update the [supported frameworks](http://localhost:3000/guides/component-testing/overview#supported-frameworks) to also mention community libraries with a link to this guide and link to a list of community libs.

Where would be a good place to list the community libraries in the docs? My initial thoughts is we can remove the older CT libs from the component testing section in plugins and list them there.

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+- `cypress-ct-react-js`
+- `cypress-ct-svelte-testing`
+- `@cypress/cypress-ct-react`
+- `@angular/cypress-ct-angular`
+
+Cypress will look at `node_modules` when it starts up, and load all the Framework Definitions found based on this naming convention.
+
+A simple example of a Framework Defintion for the [Solid.js](https://www.solidjs.com/) library might look like this:
+
+```js
```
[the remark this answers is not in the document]
lmiller1990: How about: https://github.com/cypress-io/cypress-documentation/pull/5069/commits/5b47f9521e5b1dfe7f2b8628e1001b14bba8cd01

docs/guides/component-testing/third-party-definitions.mdx
lmiller1990: [Sure thin](https://github.com/cypress-io/cypress-documentation/pull/5069/commits/62f251d8b58a614ad91aac551fde4279bb95c650)

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+<DocsImage
+  src="/img/guides/component-testing/framework-definition-2.png"
+  alt="required dependencies"
+/>
+
+## Mount Adapter
+
+The second part of defining a Framework Definition is the Mount Adapter. This is the function that renders the component in your tests using `cy.mount()`. 
+
+By default, Cypress will look for this as a `mount` function that is a **named export** from the package. This should be written in a `index.mjs` file. This example is for a [Solid.js](https://www.solidjs.com/) mount adapter:
```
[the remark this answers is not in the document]
lmiller1990: I thought about this but it adds quite a bit of complexity:

- Users won't directly import these modules anyway - even if type definitions exist, they don't actually do anything for the end user.
- The build step is messy. Cypress (the server component) will *always* require the CJS entry - in this case, `definition.cjs`. Both our bundlers (webpack and vite) want an ES module. We would need two separate build steps - one for the Node.js entry and one for the ESM entry.

Having types for the Mount Adapter *is* very useful, though. We could have a build step for the Mount Adapter, but recommend writing the Definition in CJS? Or, we can have two `tsconfig.json` and run two build steps.

I can put together a template that does this and see what it looks like.

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,245 @@
...
+    if (options.log) {
+      Cypress.log({
+        name: "mount",
+        message: "Mounted component",
+      });
+    }
+  });
+}
+
+// Cleanup after each test
```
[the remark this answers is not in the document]
lmiller1990: Let's go with between, it's more broad without getting into the implementation details. Thanks!

docs/guides/component-testing/third-party-definitions.mdx
lmiller1990: I'll throw it in for good measure.

docs/guides/component-testing/third-party-definitions.mdx
lmiller1990: I'll add some tests to the example repo so people have an idea of where to start.

lmiller1990: Done https://github.com/cypress-io/cypress-documentation/pull/5069/commits/284dd59b64672a922a18a85fc6ab1a2b18aae138

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,251 @@
...
+  "peerDependencies": {
+    "solid-js": "^1.6.0"
+    "cypress": "^12.7.0"
+  }
+}
+```
+
+## Testing
+
+If you develop the Framework Definition file using TypeScript, for example by using [our official template](https://github.com/cypress-io/cypress-ct-definition-template), as long as you don't have any compile time errors, everything should work as expected. If you do run into unexpected behavior, please [file an issue](https://github.com/cypress-io/cypress/issues).
```
lmiller1990, line 241: Worth noting we've got better warnings for definitions coming VERY soon (next sprint) https://github.com/cypress-io/cypress/issues/25838 but they are not supported **right now**. The issue requires the docs to be updated when they are added.

lmiller1990 (commented):
@elylucas @mike-plummer thanks for the reviews, I have done my best to address all comments. I also updated the official template to include TypeScript and my own example to use it.

elylucas (approved):

mike-plummer (approved):

lmiller1990:
Will merge once 12.7 goes out with https://github.com/cypress-io/cypress/pull/25780/files in it.

docs/guides/component-testing/third-party-definitions.mdx
```diff
@@ -0,0 +1,302 @@
...
+ */
+export function mount(component, options = {}) {
+  // Retrieve root DOM element that Cypress has prepared for this test
+  const root = getContainerEl()
+
+  dispose = render(() => component, root)
+
+  // Wait until next microtick to ensure any async render logic has executed
+  return cy.wait(0, { log: false }).then(() => {
+    if (options.log) {
```
astone123, line 207: This won't work unless we check the value of `options.log`. We're assuming that it's true here but that's not the case unless it is explicitly set. Should we change this to

```suggestion
    if (options.log !== false) {
```

astone123 (commented):
Not sure if I'm looking at the wrong deployment, but I'm seeing the entire page inside of the "info" banner 🤔 

Screenshot 2023-02-21 at 7 46 07 AM

astone123:
@lmiller1990 do you think we should mention dev server support at all? Maybe suggest a way to give users a custom dev server function to drop into their config until we build support for that? Is that something that we want to encourage? Not sure 🤔

lmiller1990:
@astone123 

- Banner CSS - what on earth, will look
- `options.log`- will fix
- custom dev server - we could add a link to our custom dev server docs, but those are kind of bad. If we could get a proper example (eg Preact) that'd be much more useful. I thought about a "caveats" section but it seems like too much information for the tiny group of developers who'd actually want to build a custom dev server. I'd love to see some better docs in general around this... We've got [this](https://docs.cypress.io/guides/component-testing/component-framework-configuration/#Custom-Dev-Server) but it's not exactly the most useful... :|

lmiller1990:
Banner looks good to me, tested locally - maybe preview was an old commit? 

Screenshot 2023-02-22 at 12 45 54 pm

elylucas:
> @lmiller1990 do you think we should mention dev server support at all? Maybe suggest a way to give users a custom dev server function to drop into their config until we build support for that? Is that something that we want to encourage? Not sure 🤔

I think this is a good idea. I've been trying to throw together one and the point I got stuck on was getting the dev server up and running. One project was vite and the other is a custom dev server and couldn't really get either working. Would be good to know (and update the custom dev server section under framework configuration) how to set one up from scratch.

elylucas:
> Banner looks good to me, tested locally - maybe preview was an old commit?
> 
> Screenshot 2023-02-22 at 12 45 54 pm

It got messed up in a commit suggestion, I pushed a quick fix for it in #81e6b7e

PR #5076, opened 2023-02-20
Update suggested Docker images in sample code from Node 12 to 14

rebeccaalpert:
Node 12 is not compatible with wait-on, which is mentioned elsewhere in the CI docs. This caused me some lost time and I suspect updating to at least 14 (which is compatible with it) or the latest might help someone else.


emilyrohrbough (approved):

emilyrohrbough (approved):

conversayShawn:
Resolved in #5184

PR #5080, opened 2023-02-21
Update assertions reference

BlueWinds:
As part of https://github.com/cypress-io/cypress-documentation/pull/5053, I was reviewing our guides, and think that the assertions reference could be improved to be more useful when writing tests.

I've done three things in this PR.
1. Added `.should()` examples of each assertion. Users should write their assertions like this in order to maintain retry-ability. Providing this as the first example in each case will help them make fewer mistakes. Also, I've always wished it were like this when I reference the page in my own test writing. I'm "users".
2. Removed the TDD style assertions list. We do not use it anywhere in our own codebase, and we have no tests around it. It also doesn't work with `.should() / .and()`, which is how we recommend users write tests.
3. Removed a contradictory section on negative assertions that is at odds with the text above it.

### DIFF
--- a/docs/guides/references/assertions.mdx
+++ b/docs/guides/references/assertions.mdx
@@ -25,57 +25,55 @@ guide.
 
 <Icon name="github" url="https://github.com/chaijs/chai" />
 
-### BDD Assertions
-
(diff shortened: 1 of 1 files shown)


lmiller1990 (approved):
🔥

PR #5085, opened 2023-02-23
docs: delete videos without failing/retried tests

evanlindsey:
The current example in the docs requires `del` and `lodash` packages to be installed.
This action can be achieved using only `Array.prototype.some()` and the Node `fs` library.


adamdehaven (approved):
LGTM

elylucas (approved):

elylucas:
thanks for improving the sample @evanlindsey!

PR #5099, opened 2023-03-02
docs: update slack guide with update images

elylucas:


pstakoun:
@elylucas I have an ask that was not in the original reqs – we have tag-based filtering for the Slack integration, but I don't think it's documented anywhere. Is that something we can tack on here?

emilyrohrbough (approved):

pstakoun (approved):

PR #5104, opened 2023-03-08
docs: updating supported browser versions

mschile:
Cypress uses the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#browser_compatibility) which isn't available until Chromium 80.

### DIFF
--- a/docs/guides/guides/launching-browsers.mdx
+++ b/docs/guides/guides/launching-browsers.mdx
@@ -49,8 +49,8 @@ browser by using the drop down near the top right corner:
 
 Cypress supports the browser versions below:
 
-- Chrome 64 and above.
-- Edge 79 and above.
+- Chrome 80 and above.
+- Edge 80 and above.
 - Firefox 86 and above.
 
 ### Download specific Chrome version


chrisbreiding (approved):

BlueWinds (approved):

ZengTianShengZ:
May I ask if this is true for all cypress versions or the latest cypress version. Which cypress version can I use if I want to test Chrome65

mike-plummer:
@ZengTianShengZ I believe Cypress was still compatible with Chrome 65 in version 10.11.0 (potentially early versions of v11.x as well). I cannot confirm since that is such an old browser version (over five years old!!!) - I would **_strongly_** recommend upgrading your browser as there are major security vulnerabilities that have since been fixed

MikeMcC399:
@mike-plummer 

I noticed that the checks in 
https://github.com/cypress-io/cypress/blob/f05f6236b9ae86f78cc0c40e492ec04e52438d56/packages/launcher/lib/known-browsers.ts#L3-L10

```js
// Chrome started exposing CDP 1.3 in 64
export const MIN_CHROME_VERSION = 64


// Firefox started exposing CDP in 86
export const MIN_FIREFOX_VERSION = 86


// Edge switched to Blink in 79
export const MIN_EDGE_VERSION = 79

```
are different from https://docs.cypress.io/guides/guides/launching-browsers#Browser-versions-supported

## Browser versions supported

Cypress supports the browser versions below:

Chrome 80 and above.
Edge 80 and above.
Firefox 86 and above.

---

Does that need fixing for Chrome and Edge?

mike-plummer:
Good callout @MikeMcC399 - just for context it's not always a 1-to-1 mapping since we to manage
1. Browser versions the Cypress driver can interface with properly (`run` mode)
2. Browser versions the Cypress app can run in properly (`open` mode)

In this case I think the Chrome 80 restriction is related to features used in the Cypress App (not being able to `npx cypress open`), the actual runner limitation determined by the version of CDP that's supported which is [verified here](https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/browsers/chrome.ts#L722). If I'm correct (can you sanity-check me @mschile ) you can still to use Chrome 65 to test your application headlessly you just can't utilize the interactive test runner.

At the moment we publish the version guidance you referenced which specifies what version we *fully* support

PR #5110, opened 2023-03-11
Typo in changelog regarding version 9.2.1

BambiMC:
Typo which does not recognize the version 9.2.1 as a new section.


BambiMC:
I am sorry, but i cant accept the cla, either it does not properly load when opening it or it goes through but the corresponding check is still not checked. Tried multiple browsers and devices

PR #5115, opened 2023-03-13
Update Cucumber FAQ

conversayShawn:
Add some verbiage around cucumber

### DIFF
--- a/docs/faq/questions/using-cypress-faq.mdx
+++ b/docs/faq/questions/using-cypress-faq.mdx
@@ -479,6 +479,8 @@ arguments, JSON files and other means.
 Yes.
 [You can override this with `userAgent` in your Cypress configuration](/guides/references/configuration#Browser).
 
+_Note_
+
 ## <Icon name="angle-right" /> Can I block traffic going to specific domains? I want to block Google Analytics or other providers.
 
 Yes.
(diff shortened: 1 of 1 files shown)


docs/faq/questions/using-cypress-faq.mdx
```diff
@@ -1109,10 +1111,22 @@ JavaScript Cypress tests.
 - try using the
   [Cucumber preprocessor](https://github.com/badeball/cypress-cucumber-preprocessor)
   and search our [Plugins](/plugins) page for additional helper plugins
+
+## <Icon name="angle-right" /> Should I use Cucumber to write tests?
```
emilyrohrbough, line 1115: Adding a new section under `Can I use Cucumber to write tests?` feels weird. I'd just combine the answer with the above response:
```suggestion
```

docs/faq/questions/using-cypress-faq.mdx
```diff
@@ -1109,10 +1111,22 @@ JavaScript Cypress tests.
...
+
+Cypress does not currently have an official stance on Cucumber as each use case
+is unique and based on the needs of the user.
```
emilyrohrbough, line 1118:
```suggestion
Deciding to use Cucumber is based on the needs of the users.
```

docs/faq/questions/using-cypress-faq.mdx
```diff
@@ -479,6 +479,8 @@ arguments, JSON files and other means.
 Yes.
 [You can override this with `userAgent` in your Cypress configuration](/guides/references/configuration#Browser).
 
+_Note_
```
emilyrohrbough, line 482:
```suggestion
```

elylucas (changes requested):
I agree with Emily's comment about not having a second question. Also was looking at the answer for the first question and I think we could summarize it all with something like:

While Cypress does not provide official support for Cucumber, you can use Cucumber through a [community plugin](link to the plugin). Using the plugin adds additional complexity to your testing workflow, so ensure your team understands the benefits and drawbacks before adopting Cucumber. 

If your team is looking to do behavior driven development, you might be interested in these articles that discuss using BDD's given/when/then syntax directly in Cypress instead of using Cucumber's scenarios:

(links you have to the blogs)

docs/faq/questions/using-cypress-faq.mdx
conversayShawn: 👍🏽

emilyrohrbough (approved):

elylucas (approved):

PR #5138, opened 2023-03-22
Add Table of Contents to API Docs

emilyrohrbough:
Extracted some of the API structure @BlueWinds did in #5053 and pull out the table of contents. They are solid changes that would be great to see now instead of waiting for v13.

- Added table to contents
- Updated routes so docs.cypress.io/api routes to docs.cypress.io/api/table-of-contents
- Moved Events under Cypress API section
- re-ordered sections to mirror expected usage

Before | After
--|--
Screen Shot 2023-03-22 at 12 34 46 PM|Screen Shot 2023-03-22 at 12 38 31 PM


jaffrepaul (approved):
This should help, thanks! One thin I would do is cross reference any "on" links that could be affected by some of the new URL paths.

emilyrohrbough:
@jaffrepaul I re-verified the on links and everything checks out.

PR #5141, opened 2023-03-24
add pnpm and yarn to github-action capabilities

MikeMcC399:
In addition to handling npm-based repositories https://github.com/cypress-io/github-action offers pnpm, Yarn Classic and Yarn Modern support. See https://github.com/cypress-io/github-action#examples

image

`pnpm` and `yarn` are now explicitly mentioned in 

https://docs.cypress.io/guides/continuous-integration/github-actions


jaffrepaul (approved):
👍

PR #5150, opened 2023-03-29
Cloud Projects page revisions

jaffrepaul:
Restructure and update content for clarity, swap screenshots, and general cleanup.

PR #5152, opened 2023-03-29
add caution badge and correct numbering sequence

conversayShawn:


jaffrepaul:
I have a bunch of changes to this page on deck, I'll tuck this in here https://github.com/cypress-io/cypress-documentation/pull/5150

That's a weird thing about numbered markdown lists. They can all be "1." and render still correctly. But agree it's better to not look wrong here.

Issue #5046, opened 2023-02-06
Contradictory GHA Explicit Version Number recommendations


MikeMcC399:
### Subject

GitHub-Action

### Description

- This issue is carried over from https://github.com/cypress-io/github-action/issues/671.

## Problem description

The section [Explicit Version Number](https://docs.cypress.io/guides/continuous-integration/github-actions#Explicit-Version-Number) of the [GitHub Actions](https://docs.cypress.io/guides/continuous-integration/github-actions) documentation is not aligned with the [README: Explicit version](https://github.com/cypress-io/github-action/blob/master/README.md#explicit-version) recommendations in the [GHA repository](https://github.com/cypress-io/github-action).

Releases in the [GHA repository](https://github.com/cypress-io/github-action) conform to [semantic versioning](https://semver.org/) conventions. This implies that known breaking changes cause a new MAJOR version to be released. Users should therefore be able to rely on this and **not** normally need to specify an explicit version.

[GitHub Action Versioning](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md) also discusses this topic:

> Recommendation: bind to major versions to get functionality and fixes but reserve binding to a specific release or SHA as a mitigation strategy for unforeseen breaks.

## Suggested changes

In  [GitHub Actions](https://docs.cypress.io/guides/continuous-integration/github-actions):

1. Recommend to use the latest major version `v5`
2. Explain that a specific version can be used as mitigation for a compatibility problem e.g. use `v5.0.8`
3. Update each example and convert from specific version `@v5.2.0` (which actually doesn't even exist at this time) or `@v5.x.x` to `@v5`.

MikeMcC399:
- At least the reference to a non-existent version has been fixed in PR https://github.com/cypress-io/cypress-documentation/pull/5091 even if the rest of this issue has not been addressed as yet.

MikeMcC399:
- I have submitted https://github.com/cypress-io/github-action/pull/852 and will come back to this issue after merge.

PR #5153, opened 2023-03-30
Recommend use of github-action v5 branch instead of v5.x.x tag

MikeMcC399:
PR changes https://docs.cypress.io/guides/continuous-integration/github-actions.

This PR resolves https://github.com/cypress-io/cypress-documentation/issues/5046 and aligns with the [README section: Action version](https://github.com/cypress-io/github-action/blob/master/README.md#action-version) of [cypress-io/github-action](https://github.com/cypress-io/github-action) to recommend binding to the major version of the action.

The GitHub action follows semantic versioning and strives not to introduce breaking changes without also changing the major version number. Users who specify the latest branch (currently `v5`) take advantage of incremental feature and fix improvements as they are released, which is not automatically the case when binding to an exact release version tag, such as the current [v5.6.1](https://github.com/cypress-io/github-action/releases/tag/v5.6.1) tag.

GitHub's documentation for [Action Versioning](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md) also discusses this topic:

> Recommendation: bind to major versions to get functionality and fixes but reserve binding to a specific release or SHA as a mitigation strategy for unforeseen breaks.

## Changes

- Each example<br>`uses: cypress-io/github-action@v5.x.x # use the explicit version number` is changed to simply<br>`uses: cypress-io/github-action@v5`.<br>This has the incidental advantage that each example can be copied and pasted without the need to edit it with an explicit tag version before it will work.
- The brackets for single event triggers such as<br>`[push]` are removed in alignment with PR https://github.com/cypress-io/github-action/pull/782 so the trigger becomes<br>`push`.
- Some grammatical changes are made


MikeMcC399:
@jaffrepaul 

Would somebody be able to review / merge the PRs related to `github-actions` here in the main documentation repo? I'm trying to get the information in line with what's described in the `github-actions` repository README. Readers shouldn't be confronted with conflicting information / advice.

jaffrepaul (approved):
👍
Thanks @MikeMcC399!

PR #5154, opened 2023-03-30
Set github-action examples to ubuntu-22.04

MikeMcC399:
This PR aligns the [Guides > Continuous Integration > Github Actions](https://docs.cypress.io/guides/continuous-integration/github-actions) page with the repository hosting [cypress-io/github-action](https://github.com/cypress-io/github-action). It consistently uses `ubuntu-22.04` as runner.

The GitHub runner version `ubuntu-22.04` is used as default in the [README](https://github.com/cypress-io/github-action/blob/master/README.md) file and [examples](https://github.com/cypress-io/github-action/tree/master/examples) of the action repository. This default is now carried over to the corresponding documentation page here in this repository.

Previously the GitHub runner `ubuntu-20.04` for the corresponding LTS version, released in the year 2020, was being referred to, or `ubuntu-latest`.

- `ubuntu-20.04` is older, so it makes more sense to use `ubuntu-22.04` which has a longer lifetime. According to [The Ubuntu lifecycle and release cadence](https://ubuntu.com/about/release-cycle), `ubuntu-22.04` will be fully supported for 5 years until April 2027.
- `ubuntu-latest` is theoretically problematic because it puts any version switch-over into the hands of GitHub Operations. In general it seems better to specify concrete major versions in examples, rather than `latest`. (The use of `latest` to select the `pnpm` version to install broke workflows unexpectedly when `pnpm 8.0.0` was released this week, for instance!)


jaffrepaul (approved):

MikeMcC399:
@jaffrepaul 
Does anybody else need to approve this before it is merged?

PR #5167, opened 2023-04-06
update SSO verbiage

conversayShawn:


docs/guides/cloud/organizations.mdx
```diff
@@ -338,9 +338,12 @@ organization to sign in through your SSO provider.
 
 #### **Notes**
```
jaffrepaul, line 339: We can kill the extra "**" syntax here as well since the `<hX>` tag applies these properties 🙏

jaffrepaul (approved):

PR #5168, opened 2023-04-06
update useragent verbiage

conversayShawn:
See https://github.com/conversayShawn/useragent for details

### DIFF
--- a/docs/api/commands/visit.mdx
+++ b/docs/api/commands/visit.mdx
@@ -356,7 +356,8 @@ cy.visit('index.html').then((contentWindow) => {
 
 Trying to change the `User-Agent`? You can set the `userAgent` as a
 [configuration value](/guides/references/configuration#Browser) in your
-configuration file.
+configuration file or for a **single** `cy.visit()` by setting the User-Agent
+header in the options
 
 ### Routing
 
--- a/docs/faq/questions/using-cypress-faq.mdx
+++ b/docs/faq/questions/using-cypress-faq.mdx
@@ -469,8 +469,12 @@ it('fails on unhandled rejection', () => {
 
 ## <Icon name="angle-right" /> Can I override environment variables or create configuration for different environments?
 
-Yes, you can pass configuration to Cypress via environment variables, CLI
-arguments, JSON files and other means.
+Yes.
+
+- You can pass configuration to Cypress via environment variables, CLI
+  arguments, JSON files and other means.
+- You can override the config value for a **single** `cy.visit()` by setting the
+  User-Agent header in the options to `cy.visit()`
 
 [Read the Environment Variables guide.](/guides/guides/environment-variables)
 
--- a/docs/api/cypress-api/config.mdx
+++ b/docs/api/cypress-api/config.mdx
@@ -157,9 +157,9 @@ Cypress.config() // => {defaultCommandTimeout: 10000, viewportHeight: 900, ...}
 ### Not all config values can be changed at all times
 
 Some configuration values are readonly and cannot be changed while running a
-test. Anything that's not directly under Cypress's control - like timeouts,
-`userAgent`, or environment variables - will be ignored at run-time. Be sure to
-review the list of
+test. Anything that's not directly under Cypress's control - like timeouts or
+environment variables - will be ignored at run-time. Be sure to review the list
+of
 [test configuration options](/guides/references/configuration#Test-Configuration).
 
 ### Test Configuration vs `Cypress.config()`


docs/api/commands/visit.mdx
```diff
@@ -356,7 +356,8 @@ cy.visit('index.html').then((contentWindow) => {
 
 Trying to change the `User-Agent`? You can set the `userAgent` as a
 [configuration value](/guides/references/configuration#Browser) in your
-configuration file.
+configuration file or for a **single** `cy.visit()` by setting the User-Agent
+header in the options
```
emilyrohrbough, line 360:
```suggestion
header in the options.
```

emilyrohrbough:
The `cy.visit()` documentation would need updated to call this out & would need to call out the cavieats you mentioned on the linked example. of this working

I am just curious, WHY would someone want to change the UA for a single request vs the entire run?

docs/faq/questions/using-cypress-faq.mdx
```diff
@@ -469,8 +469,12 @@ it('fails on unhandled rejection', () => {
 
 ## <Icon name="angle-right" /> Can I override environment variables or create configuration for different environments?
 
-Yes, you can pass configuration to Cypress via environment variables, CLI
-arguments, JSON files and other means.
+Yes.
+
+- You can pass configuration to Cypress via environment variables, CLI
+  arguments, JSON files and other means.
+- You can override the config value for a **single** `cy.visit()` by setting the
```
elylucas, line 476: This update to the FAQ doesn't feel needed. There's lots of options that can be overridden besides setting `userAgent`, so I don't think we need to call it out explicitly here. Unless im missing something?

docs/api/commands/visit.mdx
elylucas: lets add an example like: 

```ts
cy.visit('/heroes', {
  headers: {
    // override user-agent string
    'user-agent': 'MyBrowser/1.0 (CoolOS; v2)',
  },
});
```

elylucas (changes requested):
couple of small changes, but besides that looks good!

conversayShawn:
@emilyrohrbough 
In this case, it's not that the user does not want to update for an entire run, it's that the docs state it is not possible. 
This update is more of a `Can I` and less of a `Should I`?
Do you think we should call that out, specifically?

docs/faq/questions/using-cypress-faq.mdx
conversayShawn: I put this in the wrong section. Instead of the [Can I override environment variables or create configuration for different environments](https://docs.cypress.io/faq/questions/using-cypress-faq#Can-I-override-environment-variables-or-create-configuration-for-different-environments) section, it should have been under the [Can I override or change the default user agent the browser uses?](https://docs.cypress.io/faq/questions/using-cypress-faq#Can-I-override-or-change-the-default-user-agent-the-browser-uses) section

docs/api/commands/visit.mdx
```diff
@@ -354,9 +354,24 @@ cy.visit('index.html').then((contentWindow) => {
...
+```javascript
+// You can check your user-agent using a third-party tool
+cy.visit('https://www.whatsmyua.info/', {
+  headers: {
+    // override user-agent string
+    'user-agent': 'MyBrowser/1.0 (CoolOS; v2)',
+  },
+})
+```
+
```
emilyrohrbough, line 373: This is missing the caveats to this approach.

conversayShawn: removed example

docs/api/commands/request.mdx
```diff
@@ -364,6 +364,13 @@ functions as if it came from the browser.
 The intention of `cy.request()` is to be used for checking endpoints on an
 actual, running server without having to start the front end application.
 
+### User agent
+
+Trying to change the `User-Agent`?
+
+See
+[how to override the default user agent the browser uses](../../faq/questions/using-cypress-faq#Can-I-override-or-change-the-default-user-agent-the-browser-uses).
```
jaffrepaul, line 372: the "../.." step to create the relative path isn't necessary in the Docs site. You can use: 

```suggestion
[how to override the default user agent the browser uses](/faq/questions/using-cypress-faq#Can-I-override-or-change-the-default-user-agent-the-browser-uses).
```

docs/api/commands/visit.mdx
```diff
@@ -354,9 +354,10 @@ cy.visit('index.html').then((contentWindow) => {
 
 ### User agent
 
-Trying to change the `User-Agent`? You can set the `userAgent` as a
-[configuration value](/guides/references/configuration#Browser) in your
-configuration file.
+Trying to change the `User-Agent`?
+
+See
+[how to override the default user agent the browser uses](../../faq/questions/using-cypress-faq#Can-I-override-or-change-the-default-user-agent-the-browser-uses).
```
jaffrepaul, line 360: the "../.." step to create the relative path isn't necessary in the Docs site. You can use:

```suggestion
[how to override the default user agent the browser uses](/faq/questions/using-cypress-faq#Can-I-override-or-change-the-default-user-agent-the-browser-uses).
```

docs/faq/questions/using-cypress-faq.mdx
```diff
@@ -478,7 +478,15 @@ arguments, JSON files and other means.
...
 Yes.
-[You can override this with `userAgent` in your Cypress configuration](/guides/references/configuration#Browser).
+
+- Cypress recommends setting the `userAgent` as a
+  [configuration value](/guides/references/configuration#Browser) in your
+  configuration file.
+- It is possible to fake the `userAgent` for a **single**
+  [`cy.visit()`](/api/commands/visit) or
+  [`cy.request()`](/api/commands/intercept) by setting the `user-agent` header
+  in the options, but this will not propagate the `userAgent` in the browser and
```
emilyrohrbough, line 488:
```suggestion
  in the options, but this will not propagate to the `userAgent` in the browser and
```

emilyrohrbough (approved):

emilyrohrbough (approved):

docs/api/commands/request.mdx
```diff
@@ -364,6 +364,13 @@ functions as if it came from the browser.
 The intention of `cy.request()` is to be used for checking endpoints on an
 actual, running server without having to start the front end application.
 
+### User agent
+
+Trying to change the `User-Agent`?
+
+See
+[how to override the default user agent the browser uses](/faq/questions/using-cypress-faq#Can-I-override-or-change-the-default-user-agent-the-browser-uses).
```
jaffrepaul, line 372: Maybe a simpler way to say it is...

```suggestion
[how to override the browsers's default user agent](/faq/questions/using-cypress-faq#Can-I-override-or-change-the-default-user-agent-the-browser-uses).
```

docs/api/commands/visit.mdx
```diff
@@ -354,9 +354,10 @@ cy.visit('index.html').then((contentWindow) => {
 
 ### User agent
 
-Trying to change the `User-Agent`? You can set the `userAgent` as a
-[configuration value](/guides/references/configuration#Browser) in your
-configuration file.
+Trying to change the `User-Agent`?
+
+See
+[how to override the default user agent the browser uses](/faq/questions/using-cypress-faq#Can-I-override-or-change-the-default-user-agent-the-browser-uses).
```
jaffrepaul, line 360:
```suggestion
[how to override the browser's default user agent](/faq/questions/using-cypress-faq#Can-I-override-or-change-the-default-user-agent-the-browser-uses).
```

elylucas (approved):
sorry, lost track of this conversation and the subsequent changes. The latest changes lgtm though

jaffrepaul (approved):

PR #5184, opened 2023-04-11
node12 EOD, update example in GHA & GL docs

conversayShawn:
update node version container example to be compatible with the latest cypress version


jaffrepaul:
Nice! Is there a place we're using as the north star for docker image tags that are in sync with the _latest cypress version_?

jaffrepaul:
> I used the example here: https://github.com/cypress-io/cypress-realworld-app/blob/develop/.github/workflows/main.yml
> 
> This is referenced at the bottom of the page, however, I should also update the reference to Cypress Kitchen Sink in the `Try it out` section since this .yml does not exist in the Kitchen Sink.
> 
> Or should we reference the GHA examples repo instead? [This example here](https://github.com/cypress-io/github-action#docker-image) uses a node18 image.

Good call. I think we just point to the GHA example.

MikeMcC399:
@jaffrepaul 

Please consider also my two PRs on this page:

- https://github.com/cypress-io/cypress-documentation/pull/5186
- https://github.com/cypress-io/cypress-documentation/pull/5185

It's good to see @conversayShawn has updated the Docker versions in this PR! Node.js 12 transitioned into [end-of-life](https://github.com/nodejs/release#release-schedule) this time last year.

docs/guides/continuous-integration/github-actions.mdx
```diff
@@ -111,8 +111,8 @@ jobs:
 
 To try out the example above yourself, fork the
 [Cypress Kitchen Sink](https://github.com/cypress-io/cypress-example-kitchensink)
-example project and place the above GitHub Action configuration in
-`.github/workflows/main.yml`.
+example project and place the above GitHub Action configuration in a newly
```
jaffrepaul, line 114: I think the previous verbiage implied this, but I could see a newer engineer might not catch the nuance.

docs/guides/continuous-integration/github-actions.mdx
```diff
@@ -111,8 +111,8 @@ jobs:
 
 To try out the example above yourself, fork the
```
jaffrepaul, line 112: My only hesitation to do this is that there are already workflows in the Kitchen Sink and `single.yml` that already has `name: Cypress tests`. So adding this new workflow will likely cause a lot of noise in the GHA tab on github. It has almost the same name and all the other ones will be triggered with a new push as we discuss in the section below this. 

Let's sync up and see if we have a better repo to recommend. Otherwise, we may want to add a note about having this as the only workflow file so it's less confusing.

docs/guides/continuous-integration/github-actions.mdx
conversayShawn: Yeah, that was the goal here. GH Actions seems to be the go-to for many engineers since the ramp-up to get started is already part of their workflow.

docs/guides/continuous-integration/github-actions.mdx
conversayShawn: Sounds good. Additionally, based on [this GitHub Issue comment](https://github.com/cypress-io/github-action/issues/634#issuecomment-1437410549) from @MikeMcC399, I think it will also be better if we reverted the node versions to reflect the workaround for [`wait-on` with node18+](https://github.com/cypress-io/github-action#wait-on-with-nodejs-18) and will reflect the version used in the [Cypress RWA `main.yml`](https://github.com/cypress-io/cypress-realworld-app/blob/develop/.github/workflows/main.yml).

MikeMcC399: @conversayShawn 

I don't think that Node.js `16` should be used as an example. It is no longer the LTS version and it transitions to [end-of-life](https://github.com/nodejs/release#release-schedule) on Sep 11, 2023.

The [cypress-io/cypress-example-kitchensink](https://github.com/cypress-io/cypress-example-kitchensink) CI workflows have all been updated to use Node.js `18`.

The `wait-on` issue should not stop anybody using Node.js `18`. Unfortunately [Cypress RWA](https://github.com/cypress-io/cypress-realworld-app) is quite out of date. It is using `react-scripts` from [create-react-app](https://github.com/facebook/create-react-app) which is not being maintained (last update was one year ago) and installing RWA with `yarn` gives pages full of deprecation warnings. It should probably be migrated to use `Next.js` instead of `react-scripts`. This would also solve the `wait-on` issue. It should also be the goal to have [Cypress RWA](https://github.com/cypress-io/cypress-realworld-app) running on the LTS version of Node.js (currently `18`, moving to `20` later this year).

MikeMcC399:
@conversayShawn 

I took a second look at this, as the PR has conflicts and would overwrite some of the changes coming from me which have already been merged.

My suggestion is for you to only touch the container versions in [Guides > Continuous Integration > Github Actions](https://docs.cypress.io/guides/continuous-integration/github-actions) with the section heading [Parallelization](https://docs.cypress.io/guides/continuous-integration/github-actions#Parallelization).

For consistency with the RWA workflow [.github/workflows/main.yml](https://github.com/cypress-io/cypress-realworld-app/blob/develop/.github/workflows/main.yml) update

- `container: cypress/browsers:node12.18.3-chrome87-ff82` to
- `container: cypress/browsers:node16.16.0-chrome105-ff104-edge`

The containers in the other sections, which are already using Node.js `18` should stay that way.

MikeMcC399:
@conversayShawn 

- To move this along I have submitted a new PR https://github.com/cypress-io/cypress-documentation/pull/5202.

- It gets around the issue of this current PR which has merge conflicts.

docs/guides/continuous-integration/github-actions.mdx
conversayShawn: Makes sense. I have updated to node 18 and resolved the conflicts listed below.

conversayShawn:
> * To move this along I have submitted a new PR https://github.com/cypress-io/cypress-documentation/pull/5202.

@MikeMcC399 Great points. I updated to node 18. I will close the node 16 PR mentioned above.

MikeMcC399 (approved):
@conversayShawn 

Go for it!

I was concerned that updating to Docker with Node.js 18 would suggest that the RWA could also be updated to this version and still work, and that users would try to do it and fail because it is not compatible with Node.js 18.

When I tried to do this myself I realized that the RWA `main.yml` example is in any case very complex, and needs to be set up to run with Cypress Cloud in order to work.

So let's go with Docker for Node.js 18 in the documentation text even though it is not what the RWA example currently uses.

MikeMcC399:
I checked what happens if you actually update the Docker version in the example. It fails the Ubuntu tests, however the Windows tests run. This is because they don't run in Docker on Windows.

image

docs/guides/continuous-integration/github-actions.mdx
```diff
@@ -302,7 +302,7 @@ on: push
 jobs:
   install:
     runs-on: ubuntu-22.04
-    container: cypress/browsers:node12.18.3-chrome87-ff82
+    container: cypress/browsers:node18.12.0-chrome107
```
emilyrohrbough, line 305: We should align these updates across our examples: https://github.com/cypress-io/cypress-documentation/pull/5076

We recently re-added the `latest` tag to our docker-images. For the longevity of these examples, thoughts on using the `latest` tag? cc// @elylucas 

```suggestion
    container: cypress/browsers:latest
```

conversayShawn:
@MikeMcC399 Any objection to using `latest` as suggested by @emilyrohrbough? This seems like this would be the easiest way to keep from having to consistently update both pages of examples.

docs/guides/continuous-integration/github-actions.mdx
```diff
@@ -353,7 +353,7 @@ jobs:
 
   ui-chrome-tests:
     runs-on: ubuntu-22.04
-    container: cypress/browsers:node12.18.3-chrome87-ff82
+    container: cypress/browsers:latest
```
MikeMcC399, line 356: See also changes proposed in PR https://github.com/cypress-io/cypress-documentation/pull/5212

MikeMcC399:
@conversayShawn 

- There are conflicts / overlaps between this PR and https://github.com/cypress-io/cypress-documentation/pull/5212 from @elylucas

MikeMcC399:
@conversayShawn 

>  Any objection to using `latest` as suggested by @emilyrohrbough? This seems like this would be the easiest way to keep from having to consistently update both pages of examples.

Using `container: cypress/browsers:latest` certainly makes the documentation look cleaner.

I would not recommend it for production use though, having been bitten by the use of `latest` tags on several occasions, finding e-mails of CI runs that suddenly broke overnight or at the weekend.

If you want to change to using `latest` in the documentation then I suggest to add a note that this is for convenience in the examples and it is recommended to use specific versions in production.

If  `container: cypress/browsers:latest` is used with [RWA](https://github.com/cypress-io/cypress-realworld-app) it will fail to build, as it needs Node.js 16.16.0 specified in [.node-version](https://github.com/cypress-io/cypress-realworld-app/blob/develop/.node-version).

Each Node.js version is supported for 2 years (see [Node.js release schedule](https://github.com/nodejs/release#release-schedule)). It doesn't seem unreasonable to revise the examples once every two years, although that seems to have been a problem in past years, for whatever reason there is for that.

MikeMcC399:
@conversayShawn 

- @elylucas has described about containers in the latest updates to https://github.com/cypress-io/cypress-documentation/pull/5212. You may find that your PR is no longer necessary.

elylucas (approved):
lgtm, not sure if theres anything outstanding to address though

elylucas:
@conversayShawn I just updated this PR with the latest from master which included my GHA changes. This left with GHA page with relatively few changes but the ones for GitLab are still relevant.

MikeMcC399:
I'm not in favor of using `container: cypress/browsers:latest` however if this change is made, then you need to make sure it is consistent with the commentary text e.g.

image

elylucas:
> I'm not in favor of using `container: cypress/browsers:latest` however if this change is made, then you need to make sure it is consistent with the commentary text e.g.
> 
> image

Ya kinda feel the same way about specifying the latest docker image. On top of what Mike stated about it could lead to broken builds, people will usually want to use a docker image to pin a specific browser. Showing them an example of how to do that would be most beneficial.

docs/guides/continuous-integration/github-actions.mdx
conversayShawn: Updated to match node 16 and chrome 107

mschile (approved):

MikeMcC399:
@conversayShawn 
Maybe you'd like to add GitLab to the PR title, since the scope has been extended?

MikeMcC399 (approved):
LGTM!

PR #5189, opened 2023-04-14
fix: anchor images work properly

elylucas:


jaffrepaul (commented):
So the behavior is still the same for me locally but I see those deploy links you posted seem to be working and I get that same behavior when using the build & deploy scripts. 

I'd like to walk through the additional type checks (the Record,  etc) to make sure I'm grokking everything.

What are we hoping to gain by adding this to Prod? We'd still explore the JS scroll issue, right?

elylucas:
> So the behavior is still the same for me locally but I see those deploy links you posted seem to be working and I get that same behavior when using the build & deploy scripts.
> 
> I'd like to walk through the additional type checks (the Record, etc) to make sure I'm grokking everything.
> 
> What are we hoping to gain by adding this to Prod? We'd still explore the JS scroll issue, right?

yes, this is mostly just a test to see how it behaves in prod

jaffrepaul:
> > So the behavior is still the same for me locally but I see those deploy links you posted seem to be working and I get that same behavior when using the build & deploy scripts.
> > I'd like to walk through the additional type checks (the Record, etc) to make sure I'm grokking everything.
> > What are we hoping to gain by adding this to Prod? We'd still explore the JS scroll issue, right?
> 
> yes, this is mostly just a test to see how it behaves in prod

I'd imagine it would behave like we're seeing in the other methods (deploy links & local serve). If that's what we're hoping to assert, I think this is ok.

PR #5190, opened 2023-04-14
Enterprise ip

conversayShawn:
IPs necessary for installation for on-prem version control when using a restrictive VPN


docs/guides/cloud/gitlab-integration.mdx
```diff
@@ -88,6 +88,13 @@ with more limited access permissions.
    Integration** or **GitLab for Enterprise Integration**.
 9. (Optional) Configure the behavior for each project.
 
+:::warning
+
+If you are running the tests from within a restrictive VPN, see
+[which subdomains to add to your allowlist for Cypress Cloud and your GitLab integration to work properly](../../faq/questions/cloud-faq#Im-working-at-a-restrictive-VPN-Which-subdomains-do-I-have-to-allow-on-my-VPN-for-Cypress-Cloud-to-work-properly)
```
jaffrepaul, line 94:
```suggestion
[which subdomains to add to your allowlist for Cypress Cloud and your GitLab integration to work properly](/faq/questions/cloud-faq#Im-working-at-a-restrictive-VPN-Which-subdomains-do-I-have-to-allow-on-my-VPN-for-Cypress-Cloud-to-work-properly)
```

docs/guides/cloud/github-integration.mdx
```diff
@@ -278,6 +295,13 @@ them into Cypress Cloud, and complete the activation process.
     installed Cypress Cloud GitHub Enterprise integration! You are now ready to
     configure a GitHub Integration for a project.
 
+:::warning
+
+If you are running the tests from within a restrictive VPN, see
+[which subdomains to add to your allowlist for Cypress Cloud and your GitHub Enterprise integration to work properly](../../faq/questions/cloud-faq#Im-working-at-a-restrictive-VPN-Which-subdomains-do-I-have-to-allow-on-my-VPN-for-Cypress-Cloud-to-work-properly)
```
jaffrepaul, line 301:
```suggestion
[which subdomains to add to your allowlist for Cypress Cloud and your GitHub Enterprise integration to work properly](/faq/questions/cloud-faq#Im-working-at-a-restrictive-VPN-Which-subdomains-do-I-have-to-allow-on-my-VPN-for-Cypress-Cloud-to-work-properly)
```

docs/guides/cloud/github-integration.mdx
```diff
@@ -137,11 +137,6 @@ to GitHub.com to complete the installation:
 2. Choose which repositories to associate with the Cypress GitHub App. You can
    choose **All repositories** or **Only select repositories**. with your
    Cypress GitHub App installation.
-   <DocsImage
-     src="/img/dashboard/github-integration/select-gh-repos.jpg"
-     alt="Select GitHub repositories"
-     width={600}
-   />
 
 :::info
```
jaffrepaul, line 142: This info block can use markdown syntax for consistency with the rest of the doc

```suggestion
:::info

All current and _future_ repositories will be included with this installation if
you choose **All repositories**.

:::
```

docs/guides/cloud/github-integration.mdx
```diff
@@ -150,8 +145,23 @@ installation if you choose <strong>All repositories</strong>.
 
 :::
 
+{' '}
```
jaffrepaul, line 148: Sometimes prettier tries to create empty white space in jsx ( {' '} ) if it interprets whitespace in the area in question. The problem is if its superfluous it can end up rendering that code in the browser. You can eliminate this by just making sure there aren't extra lines here and re-lint and this should go away.

docs/guides/cloud/github-integration.mdx
```diff
@@ -150,8 +145,23 @@ installation if you choose <strong>All repositories</strong>.
...
+  alt="Select GitHub repositories"
+  width={600}
+/>
+
 3. Click the **Install** button to complete the installation.
 
+:::warning
+
+If you are running the tests from within a restrictive VPN, see
+[which subdomains to add to your allowlist for Cypress Cloud and your GitHub integration to work properly](../../faq/questions/cloud-faq#Im-working-at-a-restrictive-VPN-Which-subdomains-do-I-have-to-allow-on-my-VPN-for-Cypress-Cloud-to-work-properly)
```
jaffrepaul, line 161:
```suggestion
[which subdomains to add to your allowlist for Cypress Cloud and your GitHub integration to work properly](/faq/questions/cloud-faq#Im-working-at-a-restrictive-VPN-Which-subdomains-do-I-have-to-allow-on-my-VPN-for-Cypress-Cloud-to-work-properly)
```

docs/guides/cloud/github-integration.mdx
```diff
@@ -268,6 +278,13 @@ them into Cypress Cloud, and complete the activation process.
    install the app against all repos or select specific ones, then click the
    **Install** button.
 
+:::info
```
jaffrepaul, line 281: Same here with using the markdown text styling syntax

docs/faq/questions/cloud-faq.mdx
```diff
@@ -233,3 +233,9 @@ Cloud.
 - `https://docs.cypress.io` - **Cypress documentation**
 - `https://download.cypress.io` - **CDN download of Cypress binary**
 - `https://on.cypress.io` - **URL shortener for link redirects**
+
+**If you are using GitHub Enterprise or GitLab for Enterprise (Self-managed),
+you may also need to add the following:**
+
```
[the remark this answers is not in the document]
conversayShawn: For this use case, I was able to confirm both are needed. I added a bit of verbiage to mention "both"

docs/guides/cloud/github-integration.mdx
conversayShawn: This was on the wrong PR. It's been removed and will be adjusted on the correct PR

docs/guides/cloud/github-integration.mdx
```diff
@@ -278,6 +293,13 @@ them into Cypress Cloud, and complete the activation process.
     installed Cypress Cloud GitHub Enterprise integration! You are now ready to
     configure a GitHub Integration for a project.
 
+:::warning
```
jaffrepaul, line 296: Was this intended to be a 'warning' admonition? I think it's more informational. so it wouldn't need the red "danger" style and instead be :::info

Also we can round it out with a "." at the end of the text

docs/guides/cloud/github-integration.mdx
jaffrepaul: Was this the correct comment here? I still see the note related to my first comment. If we're keeping it, it can match the other instance of this on the page using the markdown syntax.

docs/faq/questions/cloud-faq.mdx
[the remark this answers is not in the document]
jaffrepaul: I'm rereading this whole section for better context and I think we can make a couple more edits here. 

First off, the section title reads if you work _at_ a restrictive VPN. This should likely be changed to _with_ unless there's some context here I'm not aware of. 

Also, for this section regarding Dedicated IPs... I'm not familiar with the specifics here, but can you add a little more context? Are these also info that needs to be allowed via the VPN? We say you may need to add them, but perhaps we can be a little more specific if additional steps are required. I might read that and say "add them both where?".

jaffrepaul (changes requested):
I added a couple of comments/questions for clarity.

docs/faq/questions/cloud-faq.mdx
```diff
@@ -233,3 +233,9 @@ Cloud.
 - `https://docs.cypress.io` - **Cypress documentation**
 - `https://download.cypress.io` - **CDN download of Cypress binary**
 - `https://on.cypress.io` - **URL shortener for link redirects**
+
```
jaffrepaul, line 236: I'm re-reading this whole section for better context and think we can make a couple or edits.

First, the section title reads if you work _at_ a VPN, which likely should be changed to _with_ unless theres something here I'm not aware of. 

Also, the section about the Dedicated IPs... we say you might need to add these two. Can you provide more context here? Is that the same as allowing URLs like above? I might read this and say "add these both where?"

jaffrepaul (approved):

PR #5192, opened 2023-04-14
fix: patch docusaurus scroll behavior to fix anchors not working

elylucas:


jaffrepaul (approved):
Seem to be doing the trick.

Do you think it's some kind kind of test for this? Just thinking if anything changes in the future.

elylucas:
> Seem to be doing the trick.
> 
> Do you think it's some kind kind of test for this? Just thinking if anything changes in the future.

ya lets sync next week to discuss

PR #5198, opened 2023-04-18
Update example for deleting videos to use fs

jennifer-shehane:
The 'after:spec' API page uses an older example of deleting videos - this example is incorrect and doesn't run properly. 

The examples should reflect using fs like shown on the Screenshots and Videos guide [here](https://docs.cypress.io/guides/guides/screenshots-and-videos#Control-which-videos-to-keep-and-upload-to-Cypress-Cloud)

This updates the examples on the after:spec API page to use fs and correct examples of deleting videos.


nagash77 (approved):
so long underscore!

PR #5208, opened 2023-04-23
add missed closing quote

maruk0chan:


jaffrepaul:
Thanks for spotting this @maruk0chan. It looks like this PR is failing the linting check. You'll just need to run `npm run lint:fix` and then push up whatever was changed via that script to get it to pass.

Issue #4724, opened 2022-09-11
Example about GitHub Action parallelization will not even run


mwskwong:
### Subject

Guides

### Description

According to the [guide](https://docs.cypress.io/guides/continuous-integration/github-actions#Parallelization), users should create two jobs. And for the worker job, users should specify `install: false` because the install job is supposed to have handled the installation already.

However, the reality is, that by setting `install: false`, the worker job will always fail because `cypress` is not installed. 

In fact, even the RWA doesn't have `install: false` set in the worker job. 
https://github.com/cypress-io/cypress-realworld-app/blob/develop/.github/workflows/main.yml#L99

jennifer-shehane:
This doesn't appear to be relevant anymore

PR #5212, opened 2023-04-25
docs: updates to github action guide

elylucas:
Found some code examples that were invalid (setting install: false on worker jobs would cause the pipeline to error), so made some updates to fix the code and then general cleanup/clarification work.

### DIFF
--- a/docs/guides/continuous-integration/github-actions.mdx
+++ b/docs/guides/continuous-integration/github-actions.mdx
@@ -6,7 +6,7 @@ title: GitHub Actions
 
 ## <Icon name="graduation-cap" /> What you'll learn
 
-- How to run Cypress tests with GitHub Actions as part of CI/CD pipeline
+- How to run Cypress tests with GitHub Actions as part of a CI/CD pipeline
 - How to parallelize Cypress test runs within GitHub Actions
 - How to cache build artifacts between installation jobs and worker jobs
 
@@ -43,11 +43,12 @@ such as the [checkout](https://github.com/marketplace/actions/checkout) and
 [Upload/Download Artifact Actions](https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts)
 actions used below.
 
-The Cypress team maintains the official
-[Cypress GitHub Action](https://github.com/marketplace/actions/cypress-io) for
-running Cypress tests. This action provides npm, pnpm or yarn installation,
-custom caching, additional configuration options and simplifies setup of
-advanced workflows with Cypress in the GitHub Actions platform.
+The official
+[Cypress GitHub Action](https://github.com/marketplace/actions/cypress-io) is
+maintained by Cypress and our community to help ease the setup of Cypress in a
+GitHub Action. The action provides dependency installation (via npm, yarn, or
+pnpm), built-in caching of Node dependencies, and additional configuration
+options for advanced workflows.
 
 ### Version Number Selection
 
@@ -133,7 +134,7 @@ example project and place the above GitHub Action configuration in
 
 [GitHub-hosted runners](https://github.com/actions/runner-images) offer images
 with pre-installed browsers to use for testing. The `ubuntu` and `windows`
-runners each include Google Chrome, Mozilla Firefox and Microsoft Edge
+runners each include Google Chrome, Mozilla Firefox, and Microsoft Edge
 pre-installed. The `macos` runners additionally include Apple Safari. Refer to
 [GitHub Actions Runner Images](https://github.com/actions/runner-images/blob/main/README.md)
 for current details.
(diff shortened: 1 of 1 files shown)


docs/guides/continuous-integration/github-actions.mdx
```diff
@@ -466,14 +407,14 @@ are leveraging three useful features of
...
+   the example above, we use the `group: "UI - Chrome"` option to organize all
+   UI tests for the Chrome browser into a group labeled "UI - Chrome" in the
    [Cypress Cloud](https://on.cypress.io/cloud) report.
 
 ## Cypress Real World Example with GitHub Actions
 
-A complete CI workflow against multiple browsers, viewports and operating
-systems is available in the **Cypress Real World App**.
+A complete CI workflow against multiple browsers, viewports, and operating
+systems are available in the **Cypress Real World App**.
```
elylucas, line 417:
```suggestion
systems is available in the **Cypress Real World App**.
```

elylucas:
closes #4724

docs/guides/continuous-integration/github-actions.mdx
```diff
@@ -163,9 +164,8 @@ Cypress locally and in CI.
 
 Below we extend the previous example by adding the `container` attribute using a
 [Cypress Docker Image](https://github.com/cypress-io/cypress-docker-images)
-built with Google Chrome `107`. This allows us to fix our test to a specific
-browser version without any influence due to browser version changes in the
-GitHub runner image.
+built with Google Chrome `107`. This allows us to specify a browser version
```
emilyrohrbough, line 167:
```suggestion
built with Google Chrome `107`. Specifying a browser version allow our tests to execute
```

docs/guides/continuous-integration/github-actions.mdx
```diff
@@ -163,9 +164,9 @@ Cypress locally and in CI.
 
 Below we extend the previous example by adding the `container` attribute using a
 [Cypress Docker Image](https://github.com/cypress-io/cypress-docker-images)
-built with Google Chrome `107`. This allows us to fix our test to a specific
-browser version without any influence due to browser version changes in the
-GitHub runner image.
+built with Google Chrome `107`. Specifying a browser version allow our tests to
```
MikeMcC399, line 167:
```suggestion
built with Google Chrome `107`. Specifying a browser version allows our tests to
```

docs/guides/continuous-integration/github-actions.mdx
```diff
@@ -312,7 +315,7 @@ jobs:
         with:
           # Disable running of tests within install job
           runTests: false
-          build: yarn build
+          build: npm run build
```
MikeMcC399, line 318: The video in this section refers to the [RWA repo](https://github.com/cypress-io/cypress-realworld-app) and this uses Yarn.

Are you OK with diverging from RWA by specifying npm here?

docs/guides/continuous-integration/github-actions.mdx
```diff
@@ -349,98 +344,45 @@ on: push
 
 jobs:
   install:
-  # ...
+  # ... omitted install job
 
-  ui-chrome-tests:
+  cypress-run:
     runs-on: ubuntu-22.04
-    container: cypress/browsers:node12.18.3-chrome87-ff82
```
MikeMcC399, line 356: If you remove the container, then users may follow this example and run into the open issue 
- https://github.com/cypress-io/cypress/issues/25040

I added the follow text to the GHA [readme](https://github.com/cypress-io/github-action#parallel)

"During staged rollout of a new GitHub-hosted runner version, GitHub may provide a mixture of current and new image versions used by the container matrix. It is recommended to use a [Docker image](#docker-image) in the parallel job run which avoids any Cypress Cloud errors due to browser major version mismatch from the two different image versions. A [Docker image](#docker-image) is not necessary if testing against the default built-in Electron browser because this browser version is fixed by the Cypress version in use and it is unaffected by any GitHub runner image rollout."

MikeMcC399 (changes requested):
Please check especially the removal of the Docker container in the parallel example due to the open issue https://github.com/cypress-io/cypress/issues/25040 which has caused regular customer complaints.

docs/guides/continuous-integration/github-actions.mdx
elylucas: Thanks for the heads up. In this particular example they won't run into the issue because it will use the electron browser. However, I'll expand the section below on using the same containers to include this recommendation if they are using other browsers

jaffrepaul (commented):
LGTM. Mike has a good question about _npm vs yarn_, [here](https://github.com/cypress-io/cypress-documentation/pull/5212/files#r1177911259). We should settle that one before merging.

docs/guides/continuous-integration/github-actions.mdx
elylucas: Yes. We're slowly moving our examples/docs from yarn to npm for simplicity.

jaffrepaul (approved):

PR #5219, opened 2023-04-26
Update examples so that they don't visit websites the user wouldn't control

jennifer-shehane:
- Update visit/origin examples to not visit external sites
- Ensure wording of docs are that Cypress discourages testing sites not under your control.

### DIFF
--- a/docs/api/commands/getallcookies.mdx
+++ b/docs/api/commands/getallcookies.mdx
@@ -60,7 +60,7 @@ redirects back to our site, which sets a session cookie.
 
 ```javascript
 cy.contains('Log in').click()
-cy.origin('https://cypress.io', () => {
+cy.origin('https://example.cypress.io', () => {
   cy.get('[type=password]').type('*****')
   cy.contains('Log in').click()
 })
--- a/docs/api/commands/getallcookies.mdx
+++ b/docs/api/commands/getallcookies.mdx
@@ -60,7 +60,7 @@ redirects back to our site, which sets a session cookie.
 
 ```javascript
 cy.contains('Log in').click()
-cy.origin('https://login.site.com', () => {
+cy.origin('https://cypress.io', () => {
   cy.get('[type=password]').type('*****')
   cy.contains('Log in').click()
 })
--- a/docs/api/commands/getalllocalstorage.mdx
+++ b/docs/api/commands/getalllocalstorage.mdx
@@ -64,7 +64,7 @@ cy.visit('/users', {
 
 cy.getAllLocalStorage().then((result) => {
   expect(result).to.deep.equal({
-    '/users': {
+    'http://localhost:8080': {
       key: 'value',
     },
   })
--- a/docs/api/commands/getalllocalstorage.mdx
+++ b/docs/api/commands/getalllocalstorage.mdx
@@ -37,15 +37,15 @@ Pass in an options object to change the default behavior of
 `cy.getAllLocalStorage()` yields an object where the keys are origins and the
 values are key-value pairs of `localStorage` data.
 
-For example, if `key1` is set to `value1` on `https://origin1.com` and `key2` is
-set to `value2` on `https://origin2.com`, `cy.getAllLocalStorage()` will yield:
+For example, if `key1` is set to `value1` on `https://example.cypress.io` and `key2` is
+set to `value2` on `https://www.cypress-dx.com`, `cy.getAllLocalStorage()` will yield:
 
 ```js
 {
-  'https://origin1.com': {
+  'https://example.cypress.io': {
     key1: 'value1',
   },
-  'https://origin2.com': {
+  'https://www.cypress-dx.com': {
     key2: 'value2',
   },
 }
--- a/docs/api/commands/getalllocalstorage.mdx
+++ b/docs/api/commands/getalllocalstorage.mdx
@@ -57,7 +57,7 @@ For example, if `key1` is set to `value1` on `https://example.cypress.io` and
 ### Get all localStorage
 
 ```javascript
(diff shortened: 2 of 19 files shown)


docs/api/commands/getallcookies.mdx
```diff
@@ -60,7 +60,7 @@ redirects back to our site, which sets a session cookie.
 
 ```javascript
 cy.contains('Log in').click()
-cy.origin('https://login.site.com', () => {
+cy.origin('https://cypress.io', () => {
```
chrisbreiding, line 63: If the goal is to recommend only visiting sites under one's control, should we use localhost examples instead? Users don't control the cypress.io site, so maybe it's not sending the right message to use it in the examples?

docs/api/commands/getalllocalstorage.mdx
```diff
@@ -56,15 +56,15 @@ set to `value2` on `https://other.com`, `cy.getAllLocalStorage()` will yield:
...
+cy.visit('/users', {
   onBeforeLoad(win) {
     win.localStorage.setItem('key', 'value')
   },
 })
 
 cy.getAllLocalStorage().then((result) => {
   expect(result).to.deep.equal({
-    'https://example.com': {
+    '/users': {
```
chrisbreiding, line 67:
```suggestion
    'http://localhost:8080': {
```

The key is the origin the local storage is associated with. Might also want to update above to `cy.visit('http://localhost:8080', {` as well.

docs/api/commands/getallsessionstorage.mdx
```diff
@@ -56,15 +56,15 @@ set to `value2` on `https://other.com`, `cy.getAllSessionStorage()` will yield:
...
+cy.visit('/users', {
   onBeforeLoad(win) {
     win.sessionStorage.setItem('key', 'value')
   },
 })
 
 cy.getAllSessionStorage().then((result) => {
   expect(result).to.deep.equal({
-    'https://example.com': {
+    '/users': {
```
chrisbreiding, line 67:
```suggestion
    'http://localhost:8080': {
```

Same as with the local storage version.

docs/api/commands/getallsessionstorage.mdx
```diff
@@ -37,15 +37,15 @@ Pass in an options object to change the default behavior of
...
+set to `value2` on `https://origin2.com`, `cy.getAllSessionStorage()` will yield:
 
 ```js
 {
-  'https://example.com': {
+  'https://origin1.com': {
     key1: 'value1',
   },
-  'https://other.com': {
+  'https://origin2.com': {
```
chrisbreiding, line 48: If we decide to go with localhost examples, these can be localhost on different ports.

docs/api/commands/getallcookies.mdx
jennifer-shehane: I did think about this. I thought maybe the examples would be seen as ourselves, like Cypress tests to test Cypress which we control? Localhost presented some issues with some examples because we were explicitly mentioning different origins.

Maybe some more examples could be localhost.

docs/api/commands/getallsessionstorage.mdx
jennifer-shehane: Ok, yah I wasn't sure if that would be explicit enough for the different origin examples to be useful for people.

docs/guides/guides/web-security.mdx
```diff
@@ -213,14 +213,14 @@ on an `<a>` that navigates to another superdomain.
 ```html
 <!-- Application code that is served at `localhost:8080` -->
 <html>
-  <a href="https://stackoverflow.com">Stack Overflow</a>
+  <a href="https://cypress.io">Stack Overflow</a>
```
jennifer-shehane, line 216: Fix the content of this link

docs/api/commands/getallsessionstorage.mdx
chrisbreiding: It won't let me leave another code suggestion, but this should be: 

```js
cy.visit('https://example.cypress.io', {
  onBeforeLoad(win) {
    win.sessionStorage.setItem('key', 'value')
  },
})

cy.getAllSessionStorage().then((result) => {
  expect(result).to.deep.equal({
    'https://example.cypress.io': {
      key: 'value',
    },
    },
  })
})
```

The keys for the result of `cy.getAllSessionStorage()` are the full origins, not the paths.

docs/api/commands/getalllocalstorage.mdx
```diff
@@ -56,15 +57,15 @@ set to `value2` on `https://other.com`, `cy.getAllLocalStorage()` will yield:
 ### Get all localStorage
 
 ```javascript
-cy.visit('https://example.com', {
+cy.visit('/users', {
```
chrisbreiding, line 60:
```suggestion
cy.visit('https://example.cypress.io', {
```

docs/api/commands/getalllocalstorage.mdx
```diff
@@ -56,15 +57,15 @@ set to `value2` on `https://other.com`, `cy.getAllLocalStorage()` will yield:
...
   onBeforeLoad(win) {
     win.localStorage.setItem('key', 'value')
   },
 })
 
 cy.getAllLocalStorage().then((result) => {
   expect(result).to.deep.equal({
-    'https://example.com': {
+    'http://localhost:8080': {
```
chrisbreiding, line 68:
```suggestion
    'https://example.cypress.io': {
```

chrisbreiding (changes requested):

jaffrepaul (approved):

chrisbreiding (approved):

PR #5220, opened 2023-04-26
Remove notice about the npm package and the cypress binary versions

jennifer-shehane:
I can't imagine this info being relevant to anyone. This seems like a throwback notice to the time when we had a different npm package version than the binary? What's the value in a user knowing this at the time they install? Maybe someone else's view differs and I'm missing the point of this section - welcome to feedback.


emilyrohrbough (approved):
yeah I would expect this to be the case. This really leads into how many ppl realize you can install Cypress without npm/yarn.

PR #5233, opened 2023-05-09
doc(url.mdx): update description

jofftiquez:
The description on this one is quite confusing. It stated that the command `cy.url()` yields the url as a string which implies that it directly returns the current url as a string. Whereas in reality it returns the Promise like object that actually resolves the url as a string. This confused me and got me stuck for a bit until I found the correct way of doing it from Stackoverflow which is:

```js
cy.url().then(url => {
  // the actual url string
});
```

### DIFF
--- a/docs/api/queries/url.mdx
+++ b/docs/api/queries/url.mdx
@@ -24,7 +24,7 @@ cy.url(options)
 **<Icon name="check-circle" color="green" /> Correct Usage**
 
 ```javascript
-cy.url() // Yields the current URL as a string
+cy.url() // Yields a promise that resolves the current url as a string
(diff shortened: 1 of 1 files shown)


chrisbreiding:
All commands return a promise-like object that resolves a value. Instead of repeating that for every command, we say it _yields_ the value. This is described in the [doc linked](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress#Subject-Management) by the question mark next to the _Yields_ heading on each command doc and in [the Variables and Aliases core concepts doc](https://docs.cypress.io/guides/core-concepts/variables-and-aliases).

PR #5243, opened 2023-05-12
Plugin action test 2

jaffrepaul:
Add a label upon new PRs GH action.

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -13,6 +13,6 @@ jobs:
 
       - name: add label
         uses: actions-ecosystem/action-add-labels@v1
-        if: ${{ startsWith(github.event.pull_request.title, 'plugin') }}
+        if: ${{ contains(github.event.pull_request.title, 'plugin') }}
         with:
           labels: "section: plugins"
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -1,0 +1,18 @@
+name: Pull Request Labeler
+
+on:
+  pull_request:
+    types: opened
+
+jobs:
+  add_label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - name: add label
+        uses: actions-ecosystem/action-add-labels@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,18 @@
...
+on:
+  pull_request:
+    types: opened
+
+jobs:
+  add_label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
```
nagash77, line 12: is there a reason you are checking out the code?  I would remove this step unless you plan to make use of a separate small repo that holds all your automation scripts, etc.

.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,18 @@
...
+
+      - name: add label
+        uses: actions-ecosystem/action-add-labels@v1
```
nagash77, line 15: using a (non-github) 3rd party github action is something I think we try to avoid when possible.  If you are planning on expanding on this I would suggest you check out [Github adding labels via an action](https://docs.github.com/en/actions/managing-issues-and-pull-requests/adding-labels-to-issues) and the [triage_handle_new_comments.yml](https://github.com/cypress-io/cypress/blob/develop/.github/workflows/triage_handle_new_comments.yml) workflow is an example of how to call a script in another repo where you could start out with your straight forward example and give you the flexibility to expand it down the road in a way that is much easier than having to either get a PR in the 3rd party action repo or trying to fight yml syntax and just work in normal javascript.

nagash77 (commented):
Left a couple of things to think about.

Is the goal of this PR to only run in the cypress-documentation repo?  

If you are feeling ambitious and want to make a javascript method that could be reused to add whatever labels we want to whatever issues or PRs we want that would be cool too.

Checkout [isOrgMember.js](https://github.com/cypress-io/release-automations/blob/master/scripts/triage/isOrgMember.js) as an example.  It gets used in the [comment_workflow.js](https://github.com/cypress-io/release-automations/blob/f73f798cc50c8afb5d2b9f34d305cd7b66800eb2/scripts/triage/comment_workflow.js#L24) which is called by the [triage_handle_new_comment](https://github.com/cypress-io/cypress/blob/develop/.github/workflows/triage_handle_new_comments.yml) workflow

PR #5257, opened 2023-05-19
Update aws-codebuild CI guide - set correct env.image property path

PaulGiletich:
buildspec reference: https://docs.aws.amazon.com/codebuild/latest/userguide/batch-build-buildspec.html

you can see that the image field is in the env prop, not in env.variables

### DIFF
--- a/docs/guides/continuous-integration/aws-codebuild.mdx
+++ b/docs/guides/continuous-integration/aws-codebuild.mdx
@@ -186,8 +186,7 @@ batch:
   build-list:
     - identifier: cypress-e2e-tests
       env:
-        variables:
-          IMAGE: public.ecr.aws/cypress-io/cypress/browsers:node12.14.1-chrome85-ff81
+        image: public.ecr.aws/cypress-io/cypress/browsers:node12.14.1-chrome85-ff81
 
 phases:
   install:


docs/guides/continuous-integration/aws-codebuild.mdx
```diff
@@ -186,8 +186,7 @@ batch:
   build-list:
     - identifier: cypress-e2e-tests
       env:
-        variables:
-          IMAGE: public.ecr.aws/cypress-io/cypress/browsers:node12.14.1-chrome85-ff81
+        image: public.ecr.aws/cypress-io/cypress/browsers:node12.14.1-chrome85-ff81
```
MikeMcC399, line 189: Perhaps this is also a good opportunity to update to a Docker image with a supported version of Node.js i.e. `16` or later? See https://github.com/nodejs/release#release-schedule.

Also the version could be aligned with the RWA example in
https://github.com/cypress-io/cypress-realworld-app/blob/develop/buildspec.yml

chrisbreiding (changes requested):

PaulGiletich:
@chrisbreiding good idea, done

chrisbreiding (approved):

chrisbreiding:
Thanks for the contribution, @PaulGiletich!

PR #5286, opened 2023-06-07
Update testing-your-app.mdx

wigsnes:
Help improve the docs, letting users know that there has been changes to the preferred folder used for tests using cy.visit function.

### DIFF
--- a/docs/guides/end-to-end-testing/introduction/testing-your-app.mdx
+++ b/docs/guides/end-to-end-testing/introduction/testing-your-app.mdx
@@ -21,6 +21,16 @@ slug: /guides/end-to-end-testing/testing-your-app
 
 <!-- textlint-enable -->
 
+:::caution
+
+<strong>Use e2e folder instead of integration</strong>
+
+In the video you can see the test files are located in the 'intergration' folder.
+In the newest version of cypress you are not allowed to use 'cy.visit' on component testing.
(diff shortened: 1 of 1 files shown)


wigsnes:
@MikeMcC399

The video is very useful, but using it in the docs can create confusions for new users. This is one of the first things you stumble upon when you go through the docs. I am new to cypress and it happened to me where I did this mistake. It can be hard for new users to understand that the integration folder is used for component testing only. A warning that the video is old might be better. New users would maybe try to follow the video, and just like I did make the same mistake.

MikeMcC399:
@wigsnes 

Thanks for sharing your experience watching the video https://youtu.be/5XQOK0v_YRE

The video even goes back to the first commit for Cypress in the year 2014 so it is quite a historical recording. Even so, the basic explanations being given are still very valid and useful.

If you found this video misleading, it would certainly be good to add a note which makes it clear that details may no longer be correct. The UI also looks quite different these days as well. I'm not sure that the specific point about the change of default folder from `cypress/integration` for legacy configuration to `cypress/e2e` in modern configurations needs to be specifically called out though. If any note is added, I would suggest something much more general.

elylucas:
Thanks @wigsnes and @MikeMcC399. Going to merge this into a branch in the docs repo and make some additional updates and deploy it.

PR #5289, opened 2023-06-08
remove outdated video from testing your app guide

elylucas:
While the video is good, it covers a really old version of Cypress and is confusing to those new to Cypress. The video will still be on the YT channel.


MikeMcC399:
> I think [this one](https://docs.cypress.io/guides/overview/why-cypress#What-youll-learn) is the more outdated, problematic video 😅

The ["What you'll learn" video](https://youtu.be/LcGHiFnBh3Y) is from 2019, so a refresh would certainly be good.
The download numbers / week are 10 x what the video claims.
(Also the volume level of the main presentation is far too low.)

emilyrohrbough (approved):

PR #5294, opened 2023-06-09
fix: Set prose-wrap to preserve

mjhenkes:
Set prose-wrap to the default value of preserve to allow for markdown lines greater than 80 characters.

https://prettier.io/docs/en/options.html#prose-wrap

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,5 +2,5 @@
   "semi": false,
   "singleQuote": true,
   "printWidth": 80,
-  "proseWrap": "always"
+  "proseWrap": "preserve"
 }
--- a/docs/api/commands/intercept.mdx
+++ b/docs/api/commands/intercept.mdx
@@ -1636,6 +1636,5 @@ information about the request and response to the console:
 [match-url]: #Matching-url
 [glob-match-url]: #Glob-Pattern-Matching-URLs
 [arg-method]: #method-String
-[arg-routehandler]:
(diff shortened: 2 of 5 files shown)


mjhenkes:
@MikeMcC399, those changes are the result of running `yarn run lint:fix` after the prettierrc file change.

chrisbreiding (approved):

MikeMcC399:
@mjhenkes 
> those changes are the result of running yarn run lint:fix after the prettierrc file change.

I made the mistaken assumption that `prettier --write` would not make any changes in `*.mdx` files with the setting `"proseWrap": "preserve"`, however I see that is not true!

I was going by the documentation you pointed to:
> "preserve" - Do nothing, leave prose as-is.

mjhenkes:
It does seem to contradict the documentation, but i'm ok with the minor changes it made to make the links more readable. I wonder if there was some sort of order of operations that got changed ie. line limit takes precedence over another rule to `keep links on the same line` and with the relaxed line limit the 'keep links on the same line' rule now applies.

MikeMcC399:
@mjhenkes 
I'm still confused, but unless there are some other unwanted side-effects I would go ahead with the change.

The alternative in Markdown would be to use
https://prettier.io/docs/en/ignore.html#markdown

```
<!-- prettier-ignore -->

<!-- prettier-ignore-start -->
<!-- prettier-ignore-end -->
```

in places which cause formatting issues.

jaffrepaul (approved):

PR #5317, opened 2023-06-15
Runs page imporvements

jaffrepaul:
Small copy & screenshot edits covering the new branch logic.


static/img/guides/core-concepts/cypress-app/recorded-runs.jpg
elylucas, line 1: looks like this image needs to be optimized

elylucas (changes requested):

static/img/guides/core-concepts/cypress-app/recorded-runs.jpg
jaffrepaul: oops, good call

elylucas (approved):

Issue #5305, opened 2023-06-13
Plugin badges overflow out of box


MikeMcC399:
### Description

When I view the plugin list at intermediate viewport widths, some plugin badges are displayed partially out of the box.

For instance at viewport width `1280` to `1439` the `community` badge on the entry for [@bahmutov/cypress-esbuild-preprocessor](https://github.com/bahmutov/cypress-esbuild-preprocessor) is displayed partially out of the box overlapping with the entry on its right side.  

I expect that all plugin badges stay inside their box at all browser widths.

### URL of Issue(s)

https://docs.cypress.io/plugins

### Steps to replicate

1. Open https://docs.cypress.io/plugins on Desktop at full screen width
2. Scroll down to @bahmutov/cypress-esbuild-preprocessor 
3. Gradually reduce the browser screen width to 1400px and notice how the badge `community` is partially outside of the box

### Browser

Google Chrome 114

### Device

- [X] PC
- [ ] Mac
- [ ] iPhone
- [ ] iPad
- [ ] Android Phone
- [ ] Android Tablet

### Additional Information

![plugins overlap](https://github.com/cypress-io/cypress-documentation/assets/66998419/8c4561e0-99fc-4976-b4cb-9f2e41dc5553)

PR #5320, opened 2023-06-18
fix overflowing plugins title and badge

MikeMcC399:
Closes https://github.com/cypress-io/cypress-documentation/issues/5305

This PR adds css property `flex-direction: column` for the class `pluginTitle` in [src/components/plugins-list/style.module.css](https://github.com/cypress-io/cypress-documentation/blob/main/src/components/plugins-list/style.module.css), correcting a display issue on https://docs.cypress.io/plugins.

The plugin badge (`official`, `verified`, `community` or `experimental`) is now always displayed below the name of the plugin and it never overflows out of the plugin's box.

It also corrects a potential issue which has not been apparent so far, that an unhyphenated plugin name longer than 24 characters would also overflow. This is a consequence of combining `h3` with `flex`.

## Before

plugins overlap

## After

plugins no overflow

## Verification

1. Display https://docs.cypress.io/plugins in Google Chrome / Mozilla Firefox and vary the width of the browser. Ensure that all plugin details stay within the plugin's box.

2. Temporarily add a 40 character unhyphenated plugin name (e.g. `A234567890B234567890C234567890D234567890`) and observe that this name stays inside the box independently of the browser's viewport width.

## References

- [MDN: flex-direction](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction)

cc: @elylucas

### DIFF
--- a/src/components/plugins-list/style.module.css
+++ b/src/components/plugins-list/style.module.css
@@ -69,6 +69,7 @@ ul.pluginsList li p {
 @media (min-width: 1280px) {
   .pluginTitle {
     display: flex;
+    flex-direction: column;
     justify-content: space-between;
   }
 


MikeMcC399:
It could also be resolved by removing the `h3`, but then you would need to add some other formatting for the title text, so it seemed that `flex-direction: column` was the simplest and most effective solution.

mike-plummer:
Nice find @MikeMcC399! We could also put a `flex-wrap: wrap` on the `div` but that would result in each badge wrapping at different points which doesn't look great. Only thing I don't like about using `column` is it looks slightly unbalanced due to the large bottom margin on the h3 vs the tiny space between the badge and the `hr`. Maybe we can adjust to balance it out a bit? Thoughts @elylucas ?

MikeMcC399:
@mike-plummer 

> We could also put a `flex-wrap: wrap` on the `div` but that would result in each badge wrapping at different points which doesn't look great.

I also tried that out and didn't like the results.

MikeMcC399:
Accidentally closed. Continued in #5333.

PR #5323, opened 2023-06-19
Update overwrite screenshot section

0neMiss:
The screenshot command doesn't pass or receive these params. Updates so that this code is copyable.


mike-plummer (changes requested):
Hi @0neMiss , thanks for the contribution. Are you referring to the `subject` parameter? This is the yielded value from the previous command if chaining `.screenshot` like so:

```js
// In this instance, the `subject` param of `screenshot` is the `Window` object yielded by `visit`
cy.visit('https://example.cypress.io').screenshot()
```
but is undefined if chained directly off `cy`

```js
// In this instance, the `subject` param of `screenshot` is `undefined`
cy.visit('https://example.cypress.io')
cy.screenshot()
```

It appears this code example is valid, but the param name `file` could be updated to more closely match the `screenshot` docs. LMK if I'm missing something

0neMiss:
Hey @mike-plummer ! Thanks for your response. Sorry I should have left a better description of my issue.

So when copying what is currently in the docs, typescript throws the following errors.

This error is thrown in the parameters of the callback thats passed as the second parameter of `Cypress.Commands.overwrite`
```
No overload matches this call.
  Overload 1 of 2, '(name: "screenshot", fn: CommandFnWithOriginalFn<"screenshot">): void', gave the following error.
    Argument of type '(this: Context, originalFn: CommandOriginalFn<"screenshot">, subject: string, name: Partial<Loggable & Timeoutable & ScreenshotOptions> | undefined, options: Parameters<...>[2]) => void' is not assignable to parameter of type 'CommandFnWithOriginalFn<"screenshot">'.
      Target signature provides too few arguments. Expected 4 or more, but got 3.
  Overload 2 of 2, '(name: "screenshot", fn: CommandFnWithOriginalFnAndSubject<"screenshot", unknown>): void', gave the following error.
    Argument of type '(this: Context, originalFn: CommandOriginalFn<"screenshot">, subject: string, name: Partial<Loggable & Timeoutable & ScreenshotOptions> | undefined, options: Parameters<...>[2]) => void' is not assignable to parameter of type 'CommandFnWithOriginalFnAndSubject<"screenshot", unknown>'.
      Types of parameters 'originalFn' and 'originalFn' are incompatible.
        Type 'CommandOriginalFnWithSubject<"screenshot", unknown>' is not assignable to type 'CommandOriginalFn<"screenshot">'.
          Types of parameters 'fileName' and 'options' are incompatible.
            Type 'Partial<Loggable & Timeoutable & ScreenshotOptions> | undefined' is not assignable to type 'string'.
              Type 'undefined' is not assignable to type 'string'.ts(2769)
```

Then the second error is thrown when passing those params to the `originalFn()`

image

After looking into the type definitions I found that (at least according to the type definitions). 

`originalFn` was expecting two params, a string called `fileName` and a object called `options`.

and that the callback has 3 instead of 4 parameters passed to it.

The first being the original function we are overwriting

image

The second is some string, which based on what `originalFn` was expecting I thought this would be better named as `fileName`.

image


and the  third param passed through this callback is the options. 

image

Hopefully this helps explain what I was trying to fix. So perhaps the type definitions need to be updated to account for when the function is called while chaining if i'm understanding you correctly? 


here is my package.json if you have issues reproducing.

```
{
  "name": "something",
  "version": "0.1.0",
  "private": true,
  "dependencies": {
    "@craco/craco": "^6.2.0",
    "@testing-library/jest-dom": "^5.11.4",
    "@testing-library/react": "^11.1.0",
    "@testing-library/user-event": "^12.1.10",
    "cypress": "^12.14.0",
    "react": "^17.0.2",
    "react-dom": "^17.0.2",
    "react-router-dom": "^5.2.0",
    "react-scripts": "4.0.3",
    "web-vitals": "^1.0.1"
  },
  "scripts": {
    "start": "craco start",
    "build": "craco build",
    "test": "craco test",
    "eject": "react-scripts eject"
  },
  "eslintConfig": {
    "extends": [
      "react-app",
      "react-app/jest"
    ]
  },
  "browserslist": {
    "production": [
      ">0.2%",
      "not dead",
      "not op_mini all"
    ],
    "development": [
      "last 1 chrome version",
      "last 1 firefox version",
      "last 1 safari version"
    ]
  },
  "devDependencies": {
    "autoprefixer": "^9.8.6",
    "nodemon": "^2.0.9",
    "postcss": "^7.0.36",
    "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.2.4"
  }
}

```

0neMiss:
Looked into this a bit more with a debugger. Looks like its an issue with the type definition so i'll try and figure that out instead. Although I do think that we should update this to be `fileName` rather than `name` to match the param name provided here.

image

https://docs.cypress.io/api/commands/screenshot

0neMiss:
I think i've found a fix for the type definition issue and will put up a pr for it tomorrow after I've read through the contributing guide for that repo. Feel free to close this out if you dont think the param name change is warrented. Thanks!

mike-plummer (approved):
@0neMiss Ah, gotcha, didn't register with me you were referring to a Typescript types disagreement. Thanks for chasing it down, looking forward to the fix!

I'm always in favor of tiny (easy to review 😆) docs changes to make things clearer - LGTM 🚀

0neMiss:
Pr is up for the type fix 

https://github.com/cypress-io/cypress/pull/27130

PR #5333, opened 2023-06-22
fix overflowing plugins title and badge

MikeMcC399:
Resubmission of https://github.com/cypress-io/cypress-documentation/pull/5320
Closes https://github.com/cypress-io/cypress-documentation/issues/5305

This PR adds css property `flex-direction: column` for the class `pluginTitle` in [src/components/plugins-list/style.module.css](https://github.com/cypress-io/cypress-documentation/blob/main/src/components/plugins-list/style.module.css), correcting a display issue on https://docs.cypress.io/plugins.

The plugin badge (`official`, `verified`, `community` or `experimental`) is now always displayed below the name of the plugin and it never overflows out of the plugin's box.

It also corrects a potential issue which has not been apparent so far, that an unhyphenated plugin name longer than 24 characters would also overflow. This is a consequence of combining `h3` with `flex`.

## Before

plugins overlap

## After

plugins no overflow

## Verification

1. Display https://docs.cypress.io/plugins in Google Chrome / Mozilla Firefox and vary the width of the browser. Ensure that all plugin details stay within the plugin's box.

2. Temporarily add a 40 character unhyphenated plugin name (e.g. `A234567890B234567890C234567890D234567890`) and observe that this name stays inside the box independently of the browser's viewport width.

## References

- [MDN: flex-direction](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction)

cc: @elylucas

### DIFF
--- a/src/components/plugins-list/style.module.css
+++ b/src/components/plugins-list/style.module.css
@@ -69,6 +69,7 @@ ul.pluginsList li p {
 @media (min-width: 1280px) {
   .pluginTitle {
     display: flex;
+    flex-direction: column;
     justify-content: space-between;
   }
 


MikeMcC399:
@mike-plummer 
I accidentally deleted the wrong fork yesterday when I was cleaning up after another project I was working on closed down. I couldn't re-open the closed PR https://github.com/cypress-io/cypress-documentation/pull/5320, so I have re-submitted here.

Sorry for the inconvenience!

I think this PR is good enough to merge, even if it could possibly be improved on, especially as it is only a cosmetic correction.

mike-plummer (approved):
@MikeMcC399 No problem, happens to all of us from time to time 😆 . I agree, good enough and is a definite improvement - can address any nitpicky stuff separately

PR #5337, opened 2023-06-22
Additional content fixes

jaffrepaul:
Assorted content fixes and removal of unused packages.


package.json
```diff
@@ -26,7 +26,6 @@
   },
   "dependencies": {
     "@docusaurus/core": "^2.4.0",
-    "@docusaurus/plugin-google-tag-manager": "^2.4.0",
```
jaffrepaul, line 29: This is already in the theme and doesn't need to be added as another dep.
https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-google-gtag#installation

docs/guides/continuous-integration/gitlab-ci.mdx
```diff
@@ -276,6 +276,5 @@ A complete CI workflow against multiple browsers, viewports and operating
 systems is available in the Cypress Real World App (RWA).
 
 Clone the <Icon name="github" inline="true" contentType="rwa" /> and refer to
-the
-[.gitlab-ci.yml](https://github.com/cypress-io/cypress-realworld-app/blob/develop/.gitlab-ci.yml)
+the [.gitlab-ci.yml](https://github.com/cypress-io/cypress-realworld-app/blob/develop/.gitlab-ci.yml)
```
MikeMcC399, line 279: The GitLab example has been deactivated. See also PR https://github.com/cypress-io/cypress-realworld-app/pull/1365.

docs/guides/cloud/smart-orchestration/parallelization.mdx
```diff
@@ -223,7 +223,7 @@ runs can be utilized independently of Cypress parallelization.
 
 :::info
 
-## Cross Browser Testing
+<strong>Cross Browser Testing</strong>
```
elylucas, line 226: was there a reason to using html here to make it bold instead of markdown?

elylucas (approved):
Asked a curiosity question but lgtm

docs/guides/cloud/smart-orchestration/parallelization.mdx
jaffrepaul: Yeah it a little hacky but using the markdown syntax adds in these admonitions throws additional line height/padding below the title and makes the overall height taller. I think some some of these were done with html in the old site and when we ported over this became clear so we just went with this pattern to keep things more snug until the redesign is accounted for.

docs/guides/continuous-integration/gitlab-ci.mdx
jaffrepaul: Hadn't seen that. I'll remove this ref. Thanks @MikeMcC399

MikeMcC399 (approved):
@jaffrepaul 

Thanks for removing the section about the disable GitLab CI in RWA, That looks fine now! 👍🏻

PR #5345, opened 2023-06-26
docs: add community frameworks support

shairez:
Wanted to add Qwik as a possible framework to use with component testing
figured it'll be nice to have a place where community supported framework integrations could be listed in

Thanks!

### DIFF
--- a/docs/guides/component-testing/overview.mdx
+++ b/docs/guides/component-testing/overview.mdx
@@ -117,12 +117,11 @@ following development servers and frameworks:
 
 ## Community Supported Frameworks
 
-The following integrations are built and maintained by Cypress community members. 
+The following integrations are built and maintained by Cypress community members.
 
-| Framework                                                                                                             | UI Library  | Bundler    |
-| --------------------------------------------------------------------------------------------------------------------- | ----------- | ---------- |
-| [Qwik](https://github.com/qwikifiers/cypress-qwik) <Badge type="info">Community</Badge>                               | Qwik        |    Vite    |
-
+| Framework                                                                               | UI Library | Bundler |
+| --------------------------------------------------------------------------------------- | ---------- | ------- |
+| [Qwik](https://github.com/qwikifiers/cypress-qwik) <Badge type="info">Community</Badge> | Qwik       | Vite    |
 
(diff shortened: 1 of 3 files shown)


jordanpowell88 (changes requested):
Can you run lint for your changes? It's failing in CI @shairez

mike-plummer:
We have a placeholder section for third-party definitions on [this page](https://docs.cypress.io/guides/component-testing/third-party-definitions#List-of-Framework-Definitions). I think maybe this table should go there, and we can add a link to that spot here - that way there's a way to find them without having to maintain in two different spots.

Thoughts @jordanpowell88 @shairez

shairez:
Yeah me and @jordanpowell88 discussed this exact point before I made the PR.

In terms of user experience - we figured that developers who'll look for their framework integration will probably look for it in the "overview" under "frameworks" rather than in the "how to create an integration" page (but maybe we're wrong)

Linking is a great idea! 
maybe for the sake of UX we can link it the other way around? to the overview page's table from the integration page? (so it'll be easier to find that way for the majority of users who want to use cypress with more frameworks)

WDYT?

mike-plummer:
@shairez I'm fine with this approach, definitely agree it's an easier-to-find location. I think we should update the referenced section with a link pointing here and update the copy a bit. Something like this maybe?

> ## Available Framework Definitions
> You can find a list of available Framework Definitions [here](#TODO).
> 
> If you have created a Framework Definition we would be delighted to mention it in our documentation so other Cypress users on the same framework can find it. [Submit a Pull Request](https://github.com/cypress-io/cypress-documentation/blob/main/CONTRIBUTING.md)!

shairez:
@jordanpowell88 @mike-plummer 

Done 😊

mike-plummer (approved):
LGTM, thanks @shairez !

@jordanpowell88 Apparently I can't approve the preview build for this PR - any chance you have permission to do so and give it a quick 👀 before merge?

mike-plummer:
Sorry for the delay @shairez, and thanks again for your work!

shairez:
No worries at all
Thanks for the help and review @mike-plummer !

PR #5351, opened 2023-06-30
Fix incorrect #Assertions links / Add explicit heading ids

Stacey-V-C:
Fixes links in `force` section descriptions and pre-emptively sets other explicit heading ids

### UPDATED:

Action pages in the api docs with a `force` option contain a description that tries to link to the Assertions section of the page using {url}#Assertions, but this heading no longer works as automatically generated Docusauruses are grabbing the Icon component and adding it to the hash.  

This PR adds explicit heading ids to all headings which have these Icon components - 'Yields', 'Timeouts', 'Assertions', and 'Requirements' sections on all pages that have them.  

In a few cases it also adds explicit heading ids to 'Timeouts' sections in the Notes group when when one also exists in the Rules group to avoid collisions

### ORIGINAL (see below):
Action pages in the api docs with a `force` option contain a description that tries to link to the Assertions section of the page using {url}#Assertions, which is incorrect (possibly outdated?) - the correct hash is #Assertions-Icon-namequestion-circle


mike-plummer:
Hi @Stacey-V-C , thanks for the contribution! I wonder if it might be better to try to fix the destination link value since it seems like this is likely an unintended consequence of the "question mark" icon being added to the header. I think Docusaurus may have built-in support for [customizing the Heading ID](https://docusaurus.io/docs/next/markdown-features/toc#heading-ids) used to generate the anchor link  - see the "explicit heading id" section. We'd just need to verify this works as described and that we don't accidentally generate duplicate values

Stacey-V-C:
Hi @mike-plummer, thanks for the response - I wondered if the extra hash text might be unintentional but noticed it was in a few other places too.  Makes sense if it's automatically generated! 

I'll try and take some time over the weekend to see about the custom heading IDs and look if there are more spots where this is happening - would it be best if I re-use this PR or open a new one?

mike-plummer:
@Stacey-V-C We would certainly appreciate that! Feel free to reuse this PR or open a new one, whichever is easier for you

Stacey-V-C:
@mike-plummer updated - I went ahead and preemptively added this to all heading ids that were getting the `-Icon-name-question-mark` postfix to avoid future collisions and because I think it looks cleaner but please let me know if that doesn't seem correct!

mike-plummer (approved):
LGTM, seems to work as expected when running locally. Thanks for doing such a thorough job! Just need to track down another approver or two to confirm and we can get this merged

PR #5356, opened 2023-07-04
In line 853 by mistake someone wrote "xit" instead of "it "so this test…

romoaldodoliz:
… example was going to fail

In line 853 by mistake someone wrong "xit" instead of it so this test example was going to fail


elylucas:
Hiya @romoaldodoliz , thanks for the PR, but this is actually correct. Using `xit()` is another way to write `it.skip()` and it is in context in this section. Going to close this PR, but thanks again for contributing.

PR #5367, opened 2023-07-08
docs: add lit community framework support

redfox-mx:
Hi! I wanted to add Lit as a possible framework (library) to use with component testing. n.n

Thanks!

### DIFF
--- a/docs/guides/component-testing/overview.mdx
+++ b/docs/guides/component-testing/overview.mdx
@@ -122,6 +122,7 @@ The following integrations are built and maintained by Cypress community members
 | Framework                                                                               | UI Library | Bundler |
 | --------------------------------------------------------------------------------------- | ---------- | ------- |
 | [Qwik](https://github.com/qwikifiers/cypress-qwik) <Badge type="info">Community</Badge> | Qwik       | Vite    |
+| [Lit](https://github.com/redfox-mx/cypress-lit) <Badge type="info">Community</Badge>    | Lit        | Vite    |
 
 ## Component Testing vs. End-to-End Testing
 


lmiller1990:
Thanks - I'll try your library out, assuming no major issues, I'm looking forward to adding `lit` to the list!

lmiller1990:
I tried this out. Few questions, but generally 💯 

- I created a Lit starter project following [the docs](https://lit.dev/docs/tools/starter-kits/). They use a dev server we don't support - "Web Dev Server". I had to choose "Vite" then create an empty Vite config, like you did in your project. This isn't really your fault -- we should have some way to skip selecting a dev server. I'll file an issue for this.
- Any tips on custom elements? My understanding is this is the most popular way to develop with Lit. Here's what I tried.

```ts
// SimpleGreeting.ts
import {LitElement, css, html} from 'lit';
import {customElement, property} from 'lit/decorators.js';

@customElement('simple-greeting')
export class SimpleGreeting extends LitElement {
  // Define scoped styles right with your component, in plain CSS
  static styles = css`
    :host {
      color: blue;
    }
  `;

  // Declare reactive properties
  @property()
  name?: string = 'World';

  // Render the UI as a function of component state
  render() {
    return html`<p>Hello, ${this.name}!</p>`;
  }
}
```

To get this working I did:

```ts
import {html} from 'lit';
import {SimpleGreeting} from './SimpleGreeting';

it('should display content', () => {
  // Need reference to this!
  SimpleGreeting;
  cy.mount(html`<simple-greeting name="Lachlan"></simple-greeting>`);
});
```

I needed to write `SimpleGreeting;` to get it to work (or Vite / Webpack tree shakes and drops the module import). How are you doing Lit with Cypress in larger code bases?

I won't block adding this to the docs, but I'm curious about the answer to these questions, and if there's anything we can do to make it easier to use Lit in Cypress.

lmiller1990 (approved):

lmiller1990:
This looks fine to me. I'm got some general questions, but the library works great - I was able to write some Lit elements and test them.

redfox-mx:
Hi @lmiller1990. Thanks for your time ❤️ 

About:

> I needed to write `SimpleGreeting;` to get it to work (or Vite / Webpack tree shakes and drops the module import). How are you doing Lit with Cypress in larger code bases?

Usually, if you do not need any reference of your class, you can easily import your web component as a side effect module like:

```ts
import './SimpleGreeting';
```

It prevents tree shake for your bundler.

For large codebases, you probably try to keep this behavior because esmodules will be loaded only once by the browser/bundler. You don't need to worry about web component duplication (in most cases). 

For component authors, sometimes you need to write multiple exports (node +10) to keep track of supported tree-shake (eg. utilities) features and no supported tree-shake (eg. web components) parts of your code.


> * Any tips on custom elements? My understanding is this is the most popular way to develop with Lit. Here's what I tried.

No really. Usually, working with typescript and a non-package-based app, you follow a lit example pattern. But if you keep your web components as a single-package-based library, you may intend to follow something like that.

```js
// SimpleGreeting.js # this a tree-shakable module
import {LitElement, css, html} from 'lit';
import {customElement, property} from 'lit/decorators.js';

export class SimpleGreeting extends LitElement {
  static properties = {
    name: { type: String },
  };

  constructor() {
    this.name = 'World';
  }

  // Render the UI as a function of component state
  render() {
    return html`<p>Hello, ${this.name}!</p>`;
  }
}

// index.js # this could be your sideEffect entry point
import { SimpleGreeting } from './SimpleGreeting.js'
customElements.define(SimpleGreeting, 'simple-greeting')
```

So, it depends hahaha


> * I created a Lit starter project following [the docs](https://lit.dev/docs/tools/starter-kits/). They use a dev server we don't support - "Web Dev Server". I had to choose "Vite" then create an empty Vite config, like you did in your project. This isn't really your fault -- we should have some way to skip selecting a dev server. I'll file an issue for this.

Lit teams try to offer "non-opinionated" build tools for web components. And this is "the why" they recommend his scaffolding. Personally, this option only works well for packages, but large applications need a lot more.. (A short discussion [here](https://github.com/lit/lit/discussions/3940))

Optionally you can ommit `vite.config` with the following cypress component-testing configuration 

```ts
import { defineConfig } from 'cypress'

export default defineConfig({
  component: {
    devServer: {
      framework: 'cypress-ct-lit' as any,
      bundler: 'vite',
      viteConfig: {}
    }
  }
})
```

lmiller1990:
Great, these seem fine to me. Might be good to update the docs in your repo to help with these issues (I ran into them when trying them out) but your extension works great, so happy to merge this.

PR #5376, opened 2023-07-13
Retries are not actually enabled by default

ryanpei:


jaffrepaul (approved):

MikeMcC399:
This certainly ties in with the defaults in https://docs.cypress.io/guides/references/configuration#Global

image

also for legacy configuration in https://docs.cypress.io/guides/references/legacy-configuration#Global

so it definitely looks like it should be corrected in the cloud guide. 👍🏻

PR #5399, opened 2023-07-27
Fixed invoke docs incorrectly claiming invoke will wait for promises to resolve.

vala84no:
Before submitting your plugin, please check to see if it fulfills these requirements:

- [] 🚀 Works with the latest major version of Cypress
- [ ] 🛠 Plugin purpose is clearly documented (in a README or docs website)
- [ ] 📝 Well-written documentation - A great example ([link](https://github.com/Fredx87/cypress-keycloak-commands#cypress-keycloak-commands))
- [ ] 🔬 Tested using Cypress - tests using Cypress can act as both example usage _and_ test coverage
- [ ] 👷‍♀️ CI pipeline that's passing (CircleCI and [Github Actions](https://github.com/cypress-io/cypress-tutorial-build-todo/blob/main/.github/workflows) are both free for Open Source)

If you have reason to believe your plugin does not need to meet certain requirements, please feel free to provide justification below:

### DIFF
--- a/docs/api/queries/invoke.mdx
+++ b/docs/api/queries/invoke.mdx
@@ -254,15 +254,14 @@ cy.wrap(english).invoke('greeting').should('equal', 'bye')
 
 - `.invoke()` will wait for the `function` to exist on the subject before
   running.
-- `.invoke()` will wait for the promise to resolve if the invoked `function`
-  returns a promise.
+- `.invoke()` will throw an error if the invoked `function` returns a promise.
+  If you want to invoke a function that returns a promise, use `.then()` instead.
 - `.invoke()` will automatically [retry](/guides/core-concepts/retry-ability)
   until all chained assertions have passed.
 
 ### Timeouts [<Icon name="question-circle"/>](/guides/core-concepts/introduction-to-cypress#Timeouts) {#Timeouts}
 
 - `.invoke()` can time out waiting for assertions you've added to pass.
-- `.invoke()` can time out waiting for a promise you've returned to resolve.
 
 ## Command Log
 


MikeMcC399:
@vala84no 

You have added the plugin checklist which is intended for submitting a new plugin. This is not relevant for this PR, so I suggest you delete this portion of your submission. In other words, delete everything except your last paragraph in the original post.

The history on https://docs.cypress.io/api/commands/invoke#History says "12.0.0 - .invoke() no longer supports promises or async functions" so maybe this should be put explicitly into the text, like with a simple comment "(removed in 12.0.0)" for the two lines in the document which mention promises? That would make the page usable for multiple versions, otherwise it would only be correct for Cypress 12.x and later.

The implications for migration are covered on
https://docs.cypress.io/guides/references/migration-guide#invoke.

docs/api/queries/invoke.mdx
```diff
@@ -254,15 +254,14 @@ cy.wrap(english).invoke('greeting').should('equal', 'bye')
 
 - `.invoke()` will wait for the `function` to exist on the subject before
   running.
-- `.invoke()` will wait for the promise to resolve if the invoked `function`
-  returns a promise.
+- `.invoke()` will throw an error if the invoked `function` returns a promise.
+  If you want to invoke a function that returns a promise, use `.then()` instead.
```
chrisbreiding, line 258:
```suggestion
  If you want to invoke a function that returns a promise, use [`.then()`](/api/commands/then) instead.
```

chrisbreiding (changes requested):

chrisbreiding:
> is it normal to make the current docs apply to old versions (other than the history section)? I assumed old versions of the docs would be available instead.

Currently the docs only apply to the latest version of Cypress and there isn't a publicly published way to view old versions, though we'd like to implement that some day.

MikeMcC399:
@vala84no 

> is it normal to make the current docs apply to old versions (other than the history section)? I assumed old versions of the docs would be available instead.

The website https://docs.cypress.io/ doesn't have any mechanism for selecting documentation for a certain major version of Cypress, like for instance https://docs.npmjs.com/cli/v9/ does with a drop-down list.

@chrisbreiding would be in a better position to answer your question about how this is generally handled in the Cypress documentation.

chrisbreiding:
> The history on [docs.cypress.io/api/commands/invoke#History](https://docs.cypress.io/api/commands/invoke#History) says "12.0.0 - .invoke() no longer supports promises or async functions" so maybe this should be put explicitly into the text, like with a simple comment "(removed in 12.0.0)" for the two lines in the document which mention promises? That would make the page usable for multiple versions, otherwise it would only be correct for Cypress 12.x and later.

The way @vala84no has it implemented is correct. Besides the history section, we don't make references to older versions of Cypress or try to make the docs applicable to multiple mismatching versions.

chrisbreiding (approved):

jaffrepaul (approved):

PR #5402, opened 2023-07-31
docs: install cypress via pnpm

inblossoms:
This PR is related to [#27419](https://github.com/cypress-io/cypress/issues/27419).

### DIFF
--- a/docs/guides/getting-started/installing-cypress.mdx
+++ b/docs/guides/getting-started/installing-cypress.mdx
@@ -6,7 +6,7 @@ title: Installing Cypress
 
 ## <Icon name="graduation-cap" /> What you'll learn
 
-- How to install Cypress via `npm`、`yarn` or `pnpm`
+- How to install Cypress via `npm`, `yarn` or `pnpm`
 - How to install Cypress via direct download
 - How to version and run Cypress via `package.json`
 
--- a/docs/guides/getting-started/installing-cypress.mdx
+++ b/docs/guides/getting-started/installing-cypress.mdx
@@ -6,7 +6,7 @@ title: Installing Cypress
 
 ## <Icon name="graduation-cap" /> What you'll learn
 
-- How to install Cypress via `npm` or `yarn`
+- How to install Cypress via `npm`、`yarn` or `pnpm`
 - How to install Cypress via direct download
 - How to version and run Cypress via `package.json`
 
@@ -77,6 +77,35 @@ yarn add cypress --dev
 
 System proxy properties `http_proxy`, `https_proxy` and `no_proxy` are respected
 for the download of the Cypress binary.
+
+### <Icon name="terminal" /> `pnpm install`
+
+Installing Cypress via [`pnpm`](https://pnpm.io/):
+
+```shell
+git clone https://github.com/cypress-io/github-action
+```
+
+Clone Cypress's GitHub Action repository to the current directory. This will create a folder called "github-action" and copy the contents of the repository into it.
+
+```shell
+cd github-action/examples/basic-pnpm
+```
+
+`/basic-pnpm` will be the location of the later example code.
+
+```shell
+pnpm install --frozen-lockfile && pnpm run test
+```
+
+Install the dependencies and run the test cases in the project
+
+:::info
+
+You need to make sure that you have the `pnpm` environment installed locally.
+If not you need to install it: `npm install pnpm@latest -g`.
+
+:::
 
 ### <Icon name="download" /> Direct download
 
--- a/docs/guides/getting-started/installing-cypress.mdx
+++ b/docs/guides/getting-started/installing-cypress.mdx
@@ -78,27 +78,17 @@ yarn add cypress --dev
 System proxy properties `http_proxy`, `https_proxy` and `no_proxy` are respected
 for the download of the Cypress binary.
 
-### <Icon name="terminal" /> `pnpm install`
+### <Icon name="terminal" /> `pnpm add`
 
 Installing Cypress via [`pnpm`](https://pnpm.io/):
 
 ```shell
-git clone https://github.com/cypress-io/github-action
+cd /your/project/path
 ```
-
-Clone Cypress's GitHub Action repository to the current directory. This will create a folder called "github-action" and copy the contents of the repository into it.
 
 ```shell
-cd github-action/examples/basic-pnpm
+pnpm add cypress -D
 ```
-
-`/basic-pnpm` will be the location of the later example code.
-
-```shell
-pnpm install --frozen-lockfile && pnpm run test
-```
-
-Install the dependencies and run the test cases in the project
 
 :::info
 


MikeMcC399:
Hi @inblossoms 

- The example with github-actions I gave in https://github.com/cypress-io/cypress/issues/27419 was just to demonstrate to you that pnpm works. It's not really suitable here for the main documentation.

- I suggest instead that you submit a documentation issue and ask for Cypress installation using `pnpm` be documented.

inblossoms:
@MikeMcC399

ugh, I misinterpreted your meaning and submitted the PR. 
Thank you again for pointing out the problem.
I wonder if I need to turn it off, but this PR might be useful for users like me.

docs/guides/getting-started/installing-cypress.mdx
```diff
@@ -6,7 +6,7 @@ title: Installing Cypress
 
 ## <Icon name="graduation-cap" /> What you'll learn
 
-- How to install Cypress via `npm` or `yarn`
+- How to install Cypress via `npm`、`yarn` or `pnpm`
```
MikeMcC399, line 9:
```suggestion
- How to install Cypress via `npm`, `yarn` or `pnpm`
```

docs/guides/getting-started/installing-cypress.mdx
```diff
@@ -78,6 +78,35 @@ yarn add cypress --dev
...
+```
+
+Install the dependencies and run the test cases in the project
+
+:::info
+
+You need to make sure that you have the `pnpm` environment installed locally.
+If not you need to install it: `npm install pnpm@latest -g`.
+
+:::
```
MikeMcC399, line 108:
```suggestion
```shell
cd /your/project/path
```

```shell
pnpm install cypress --dev
```

System proxy properties `http_proxy`, `https_proxy` and `no_proxy` are respected
for the download of the Cypress binary.
```

**Comment by inblossoms** [comment]
uh-huh, Do I need to change it to 'pnpm install cypress --dev'? The 'pnpm install cypress --dev' approach is not desirable in my use. It is also because of this way of initializing Cypress that I submit this issue.
It is a good idea to provide the installation steps by tabbing or updating the video, maybe I can give it a try.
I may need to sign the CLA later, as you said, if I want this PR to be accepted.
(I have benefited a lot from talking with you.)

**Comment by MikeMcC399** [comment]
@inblossoms 
> Do I need to change it to 'pnpm install cypress --dev'? The 'pnpm install cypress --dev' approach is not desirable in my use. It is also because of this way of initializing Cypress that I submit this issue.

Installing cypress into `package.json`'s  `devDependencies` aligns to the Getting Started instructions for npm and yarn. Getting Started is not intended to cover all use cases though. Why was 'pnpm install cypress --dev' not right for you?

> It is a good idea to provide the installation steps by tabbing or updating the video, maybe I can give it a try. 

I suggest not to do this in this PR. Getting a PR reviewed by the Cypress team works better the simpler the PR contents are, so if there are more complex things to change, then multiple PRs tends to work better. 

> I have benefited a lot from talking with you.

Thanks for raising this suggestion! Every time somebody has a problem following the documentation it's an opportunity to review and improve the documentation.

PS You should definitely be prepared to sign the CLA as PRs only ever get accepted on this condition. If you do not intend at some stage to sign the CLA then it would not be worth your investing further effort here.

**Comment by inblossoms** [comment]
> Installing cypress into `package.json`'s `devDependencies` aligns to the Getting Started instructions for npm and yarn. Getting Started is not intended to cover all use cases though. Why was 'pnpm install cypress --dev' not right for you?

I don't know why, but it takes more than twenty minutes to install with 'pnpm install cypress -D' (as I recall), but only 10-20 seconds with npm. Moreover, the project completed by pnpm installation lost the test case, so I need to install the patch of the use case again for the test case.

**Comment by MikeMcC399** [comment]
@inblossoms 

[pnpm > install](https://pnpm.io/cli/install#--dev--d) writes:

> --dev, -D
> Only devDependencies are installed and dependencies are removed insofar they were already installed, regardless of the
> NODE_ENV.

With pnpm 8.6.11

```bash
pnpm install cypress -D
```

took about 20 seconds to install.

However I was very surprised when I tried

```bash
pnpm install cypress --dev
```

and I found that it produces the error

```text
 ERROR  Unknown option: 'dev'
For help, run: pnpm help add
```

🙁 

This needs to be following up in https://github.com/pnpm/pnpm I guess.

MikeMcC399:
`pnpm install cypress --save-dev` works, which conflicts with the [pnpm documentation](https://pnpm.io/cli/install#--dev--d).

I suggest to put this PR on hold until `pnpm` has been corrected.

MikeMcC399:
- pnpm documentation issue opened here: https://github.com/pnpm/pnpm.io/issues/438

inblossoms:
> pnpm install cypress --save-dev works, which conflicts with the [pnpm documentation](https://pnpm.io/cli/install#--dev--d).

When I found that I could not install it correctly through 'pnpm', I wondered if it was related to the underlying implementation of pnpm. 
Now I can stop caring so much about it.

docs/guides/getting-started/installing-cypress.mdx
```diff
@@ -78,6 +78,35 @@ yarn add cypress --dev
 System proxy properties `http_proxy`, `https_proxy` and `no_proxy` are respected
 for the download of the Cypress binary.
 
+### <Icon name="terminal" /> `pnpm install`
+
+Installing Cypress via [`pnpm`](https://pnpm.io/):
+
```
jordanpowell88, line 84: Let's simplify this to not use an actual example  and to be more generic and in line with the `npm` and `yarn` installs examples above

Maybe something like:


### <Icon name="terminal" /> `pnpm add`

Installing Cypress via [`pnpm`](https://pnpm.io):

```shell
cd /your/project/path
```
```shell
pnpm add cypress -D
```

:::info

You need to make sure that you have the `pnpm` environment installed locally.
If not you need to install it: `npm install pnpm@latest -g`.

:::

jordanpowell88 (changes requested):

docs/guides/getting-started/installing-cypress.mdx
MikeMcC399: If the bit about `npm install pnpm@latest -g` is added here, then it should also be added for yarn i.e. `npm install yarn@latest -g`, since Node.js only comes with npm installed. GitHub runners are a bit different, they have npm and yarn pre-installed but not pnpm.

inblossoms: 🫤， Oh My use of pnpm was a really elementary mistake！

inblossoms:
Hi @MikeMcC399

add or install, I am very terrible at using pnpm. You're just trying to clear my mind. thanks for you.

jordanpowell88 (approved):

PR #5408, opened 2023-08-03
Fix invalid port number in web-security.mdx

jlherren:
According to [RFC 6335 - 6. Port Number Ranges](https://www.rfc-editor.org/rfc/rfc6335.html#section-6)

"the Dynamic Ports, also known as the Private or Ephemeral Ports,
from 49152-65535 (never assigned)"

so the value 65874 would be outside the permitted range.

### DIFF
--- a/docs/guides/guides/web-security.mdx
+++ b/docs/guides/guides/web-security.mdx
@@ -28,7 +28,7 @@ and we are _mostly_ able to do this.
 - Uses the browser's internal APIs for network level traffic.
 
 When Cypress first loads, the internal Cypress web application is hosted on a
-random port: something like `http://localhost:65874/__/`.
+random port: something like `http://localhost:64874/__/`.
 
 After the first [`cy.visit()`](/api/commands/visit) command is issued in a test,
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@jlherren 

Good catch! According to [RFC 6335 - 6. Port Number Ranges](https://www.rfc-editor.org/rfc/rfc6335.html#section-6) 

"the Dynamic Ports, also known as the Private or Ephemeral Ports,
      from 49152-65535 (never assigned)"

so the value `65874` would be outside the permitted range.

jlherren:
It's not only not permitted, it's also physically impossible to use, since it's not representable as an unsigned 16-bit integer.

nagash77:
Hi @jlherren can you please sign the CLA so that we may accept this PR.

MikeMcC399:
@jlherren 

> It's not only not permitted, it's also physically impossible to use, since it's not representable as an unsigned 16-bit integer.

49152-65535 is `0xC000` - `0xFFFF`

so yes, you're completely right.

nagash77 (approved):

Issue #5293, opened 2023-06-09
Old plugins and use of Legacy section


MikeMcC399:
### Subject

Contributing - Plugins

### Description

The [Plugins list](https://docs.cypress.io/plugins) includes a link "Follow these instructions to [submit your own plugin](https://github.com/cypress-io/cypress-documentation/blob/master/CONTRIBUTING.md#adding-plugins)." as a section in the [CONTRIBUTING](https://github.com/cypress-io/cypress-documentation/blob/master/CONTRIBUTING.md) document - see below under "Adding Plugins" for full text.

It however does not deal with the situation that a listed plugin no longer meets the criteria for new submission.

## Suggestion

Add a new section to [CONTRIBUTING](https://github.com/cypress-io/cypress-documentation/blob/master/CONTRIBUTING.md) which deals with maintaining plugins.

- What are the criteria for plugins to remain on the [Plugins list](https://docs.cypress.io/plugins)?
- What are the criteria for a plugin to be listed in the [Plugins list > Legacy section](https://docs.cypress.io/plugins#Legacy)? For instance is there a minimum version of Cypress which the plugin needs to be compatible with?

## Discussion

The [Plugins list > Legacy](https://docs.cypress.io/plugins#Legacy) section does not define the meaning of "Legacy", so unless there is some text added to the Legacy section, it is unclear what it means when a plugin is listed in this section.

- Legacy could refer to a plugin which runs only on Cypress 9 and below using a [Configuration (Legacy)](https://docs.cypress.io/guides/references/legacy-configuration)
- Legacy could also refer to plugins which are archived, deprecated or otherwise unsupported

## Problematic plugin states

- Plugin's npm registry entry is deprecated
- Plugin is explicitly deprecated in source repo
- Plugin's source repo is archived
- Plugin is not compatible with current major version of Cypress
- Plugin's CI pipeline fails

### @cypress/xpath

- npm package [@cypress/xpath](https://www.npmjs.com/package/@cypress/xpath)

```text
$ npm view @cypress/xpath deprecated
Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
```

- Source https://github.com/cypress-io/cypress/tree/develop/npm/xpath#readme

> @cypress/xpath has been deprecated and is no longer supported.

### @cypress/skip-test

- npm package [@cypress/skip-test](https://www.npmjs.com/package/@cypress/skip-test)

[...]

PR #5416, opened 2023-08-10
docs: remove deprecated links to extend configuration

annamasc:
Extending configuration with [@bahmutov/cypress-extends](https://github.com/bahmutov/cypress-extends) package is deprecated since Cypress 10.0

### DIFF
--- a/docs/guides/references/configuration.mdx
+++ b/docs/guides/references/configuration.mdx
@@ -764,9 +764,5 @@ DEBUG=cypress:cli,cypress:server:specs
   [Cypress.env()](/api/cypress-api/env)
 - [Environment variables](/guides/guides/environment-variables)
 - [Environment Variables recipe](/examples/recipes#Fundamentals)
-- [Extending the Cypress Configuration File](https://www.cypress.io/blog/2020/06/18/extending-the-cypress-config-file/)
-  blog post and
-  [@bahmutov/cypress-extends](https://github.com/bahmutov/cypress-extends)
-  package.
 - Blog post
   [Keep passwords secret in E2E tests](https://glebbahmutov.com/blog/keep-passwords-secret-in-e2e-tests/)


MikeMcC399:
Is there an official statement somewhere that this plugin is deprecated for Cypress 10 and later? I could not find one.

The instructions for https://github.com/bahmutov/cypress-extends in the README are only for legacy Cypress 9 and earlier. There is no mention of Cypress 10 and later.

It seems anyway that it doesn't work on Cypress 10 and even if it worked on Cypress 9 and then Cypress 12 is installed, it fails the migration process from 9 to 12. `TypeError [ERR_INVALID_ARG_TYPE]: The "id" argument must be of type string. Received undefined`.

The plugin should probably also be marked as deprecated on https://docs.cypress.io/plugins#development-tools

https://github.com/cypress-io/cypress-documentation/blob/b4266c27eaadbdd4030a18488a961bcec45c7910/src/data/plugins.json#L292-L298

I agree it makes sense to remove it from the bottom of 
https://docs.cypress.io/guides/references/configuration, even if it remains at the bottom of https://docs.cypress.io/guides/references/legacy-configuration.

jaffrepaul:
Thanks @annamasc!

@MikeMcC399 I'm good to keep the scope of this to removing the config callout. There isn't an established design pattern for showing deprecated or versioned plugins and the number of users on < v10 isn't significant atm to dedicate resources here.

jaffrepaul (approved):

MikeMcC399:
@jaffrepaul 

- I was fine with this PR, in case you thought otherwise!

- I had opened an issue about handling legacy / deprecated plugins https://github.com/cypress-io/cypress-documentation/issues/5293 after I began trying to clean up some items in the plugins list. I put that effort on hold though when I realised how many outdated entries there were and I couldn't do it on my own. The issue was assigned to Ely until he left.

PR #5421, opened 2023-08-17
Add caution against using cypress as name of script

MikeMcC399:
This PR adds a caution / Best Practice message to [Getting Started > Opening the App > Adding npm Scripts](https://docs.cypress.io/guides/getting-started/opening-the-app#Adding-npm-Scripts) to suggest **not** to use `cypress` as a literal script name due to its interference with the Yarn / Cypress CLI.

## Background

Defining a script in `package.json` with the name `cypress` causes ambiguity for Yarn users and should be avoided.

For instance, with the following definition

```json
  "scripts": {
    "cypress": "cypress open"
  }
```

executing

```bash
yarn run cypress verify
```

will be executed instead as

```bash
yarn run cypress open
```

When there is a script and a binary with the same name `yarn run` (Yarn Classic and Modern) selects the script. Yarn Modern offers the `--binaries-only` option to make the choice deterministic. Yarn Classic does not offer this option.

The following command overrides the script definition for Yarn Modern only.

```bash
yarn run --binaries-only cypress verify
```

- This "gotcha" caused https://github.com/cypress-io/github-action/issues/941 in June 2023.

### DIFF
--- a/docs/guides/getting-started/opening-the-app.mdx
+++ b/docs/guides/getting-started/opening-the-app.mdx
@@ -53,6 +53,17 @@ npm run cypress:open
 
 ...and Cypress will open right up for you.
 
+:::caution
+
+<strong>Best Practice</strong>
+
+Don't use `cypress` as the exact name of a script, especially if you use Yarn as package manager.
+Yarn will select the script in preference to the Cypress binary of the same name and
+Cypress CLI commands may not work as expected.
+Use instead a descriptive and non-ambiguous script name such as `cypress:open` or `cypress:run`.
+
+:::
+
 ### CLI tools
 
 By installing Cypress through `npm` you also get access to many other CLI
--- a/docs/guides/getting-started/opening-the-app.mdx
+++ b/docs/guides/getting-started/opening-the-app.mdx
@@ -58,8 +58,8 @@ npm run cypress:open
 <strong>Best Practice</strong>
 
 Don't use `cypress` as the exact name of a script, especially if you use Yarn as package manager.
-Yarn will select the script in preference to the Cypress binary of the same name and
-Cypress CLI commands may not work as expected.
+When running commands on the Cypress binary (e.g `yarn cypress verify`), Yarn will reference the
+script of the same name instead and [Cypress CLI commands](/guides/guides/command-line) may not work as expected.
 Use instead a descriptive and non-ambiguous script name such as `cypress:open` or `cypress:run`.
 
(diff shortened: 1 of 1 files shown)


docs/guides/getting-started/opening-the-app.mdx
```diff
@@ -53,6 +53,17 @@ npm run cypress:open
 
 ...and Cypress will open right up for you.
 
+:::caution
+
+<strong>Best Practice</strong>
+
+Don't use `cypress` as the exact name of a script, especially if you use Yarn as package manager.
+Yarn will select the script in preference to the Cypress binary of the same name and
```
jaffrepaul, line 61: I think removing this bit makes it more succinct and clear what the alert is all about. 

```suggestion
Yarn will select the script to the Cypress binary of the same name and
```

MikeMcC399: @jaffrepaul 

Your suggestion changes the meaning, so I wouldn't want to agree to it.

What I'm trying to explain is that when there is a script and a binary with the same name `yarn run` (Yarn Classic and Modern) selects the script not the binary.

The Yarn Classic documentation doesn't explain this, however [Yarn Modern > `yarn run`](https://v3.yarnpkg.com/cli/run#details) says:

image

If you have a better way of explaining this succinctly then let me know. It's always a challenge to write so that other people will understand, so I appreciate your taking a look at this!

MikeMcC399: Another possibility would be to drop the detailed explanation and just say:

... otherwise Cypress CLI commands may not work as expected.

jaffrepaul: Ah, I read this pretty quickly and while I understood the goal, I had them reversed. The more succinct version works but I think it's good to explain this in an issue with yarn and not cypress. 

This is even more verbose but we could try something like:

Do not use cypress as the exact name of a script, especially if you use Yarn as package manager. When running commands on the Cypress binary(e.g `yarn cypress verify`), Yarn will reference the script of the same name instead and Cypress CLI commands may not work as expected. Use instead a descriptive and non-ambiguous script name such as cypress:open or cypress:run.

MikeMcC399: Thank you! I've used your text which explains the issue in an understandable way.

jaffrepaul (approved):

PR #5453, opened 2023-08-30
Fix "recipe" typo in FAQ

mattxwang:
Found this while reading through the docs. Thanks for making them open-source!


mattxwang:
Hm, CircleCI killed the CI process on their end (is this part of their degraded status issue?). I'm not able to retrigger a run, but I don't think it's a consequence of my change.

jaffrepaul:
Good catch @mattxwang. No worries on the build status, CircleCI seems to be having issues with builds currently. I manually triggered it and we're looking good now.

jaffrepaul (approved):

PR #5460, opened 2023-09-01
docs: add test replay safari troubleshooting

jaffrepaul:
Just what it says on the tin ☝️


docusaurus.config.js
```diff
@@ -149,9 +149,9 @@ const config = {
       // Styles for this are controlled in src/css/announcement-bar.scss
       announcementBar: {
         //give id a unique value to get a new announcement bar to appear
-        id: 'cypress-v13',
+        id: 'cypress-v13+',
```
pstakoun, line 152: intentional change?

pstakoun (approved):

docusaurus.config.js
jaffrepaul: yep. but unrelated to the safari bit. I'll pull it from this PR

PR #5461, opened 2023-09-01
chore: bump circle resources

jaffrepaul:
- bump CirceCI resources to the next tier to help with resource failures
- remove cypress orb for now (nothing custom to test, only docusaurus + markdown content)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,11 +1,8 @@
 version: 2.1
 
-orbs:
-  cypress: cypress-io/cypress@1.26.0
-
 executors:
   node-executor:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,10 +1,8 @@
 version: 2.1
 
-orbs:
```
jaffrepaul, line 3: not currently running cypress tests on the repo. the old version of the site had ineffective tests and now there isn't anything custom here so we don't need to test docusaurus in this case.

jennifer-shehane (approved):
Should be good.

PR #5463, opened 2023-09-04
Add required archlinux packages for Cypress

MuhammadSawalhy:
I was trying to run Cypress in Arch (WSL). I had lots of trouble using Arch in a WSL and then using Cypress in this thing. So it will be helpful to make it easier for other Arch users and include the required packages in the documentation.


docs/guides/getting-started/installing-cypress.mdx
```diff
@@ -214,6 +214,12 @@ on your system.
 apt-get install libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb
 ```
 
+#### Arch
+
+```shell
+pacman -S gtk2 gtk2 alsa-lib xorg-server-xvfb libxss nss libnotify
```
mschile, line 220: `gtk2` is listed twice, is one of them supposed to be something else?

MuhammadSawalhy: Oh sorry, it should be `gtk3`!

PR #5469, opened 2023-09-07
docs: add example to expand the usage of the get method

hlawuleka:


docs/api/queries/get.mdx
```diff
@@ -143,6 +143,12 @@ cy.get('form').within(() => {
 })
 ```
 
+#### Find an anchor element by it's label/text, remove some attribute and click it
```
MikeMcC399, line 146:
```suggestion
#### Find an anchor element by its label/text, remove some attribute and click it
```

grammar correction

docs/api/queries/get.mdx
```diff
@@ -143,6 +143,12 @@ cy.get('form').within(() => {
...
+
+```javascript
+cy.get("a", { name: "My link text" }).invoke("removeAttr", "target").click();
```
MikeMcC399, line 149: other examples use single quotes

docs/api/queries/get.mdx
```diff
@@ -143,6 +143,12 @@ cy.get('form').within(() => {
 })
 ```
 
+#### Find an anchor element by its label/text, remove some attribute and click it
+
```
marktnoonan, line 147: It would be good to an add explanation of why this is useful. Something like:

```suggestion

The following may be useful to change remove a `target` of `"_blank"` before clicking a link that would otherwise launch a new tab/window: 

```

docs/api/queries/get.mdx
```diff
@@ -143,6 +143,12 @@ cy.get('form').within(() => {
...
+```javascript
+cy.get('a', { name: 'My link text' }).invoke('removeAttr', 'target').click();
```
marktnoonan, line 149: The syntax here is not correct - it looks like a mix of Cypress and Testing Library syntax, the following would work though:

```suggestion
cy.contains('a', 'hello').invoke('removeAttr', 'target').click();
```

marktnoonan (changes requested):
I made a couple of comments:

- The syntax doesn't seem right in the example
- The code could use some more context to explain why you might want to do this

Otherwise, this looks good! I would be happy to approve after these changes are made.

docs/api/queries/get.mdx
MikeMcC399: @marktnoonan 

If the example is changed from `cy.get()` to `cy.contains()`, would it then need to be added to https://github.com/cypress-io/cypress-documentation/blob/main/docs/api/queries/contains.mdx instead?

jennifer-shehane:
@hlawuleka Thanks for the contribution! Could you please sign our CLA? Also can you address the feedback? Otherwise we'll need to close this PR.

PR #5523, opened 2023-10-06
docs(gh action): use correct workflow yaml syntax

schlich:
Even though this is a stub, it should have the correct syntax with the dash indicating the mapping for "uses"


mschile (approved):

mschile:
Hi @schlich 👋, thanks for the contribution!

PR #5562, opened 2023-11-14
CYCLOUD-1740 multi repo best practices

jaffrepaul:
Add Cloud project best practices around having tests in a different repo than the source code.


docs/guides/cloud/account-management/projects.mdx
```diff
@@ -299,6 +299,35 @@ See also
 [Highlight your project's test status with Cypress README badges](https://www.cypress.io/blog/2020/09/02/highlight-your-projects-test-status-with-cypress-readme-badges/)
 announcement.
 
+## Best Practices
+
+### Multi-repo Implementation
+
+There is a sceario where you may have Cypress tests in a repo separate from the application code itself. This creates a sub-optimal [Branch Review](guides/cloud/branch-review) experience considering the commit and branch interpretation of each git directory.
```
tbiethman, line 306:
```suggestion
There is a scenario where you may have Cypress tests in a repo separate from the application code itself. This creates a sub-optimal [Branch Review](guides/cloud/branch-review) experience considering the commit and branch interpretation of each git directory.
```

...though I wonder if we need to explicitly call out branch review here; it'd be problematic for really any sort of VCS integration we have, including statuses/PR comments, etc.

jaffrepaul: Makes sense. I figured it may be useful to callout a practical area in the product this would impact but we can make it more general to VC. 

@pstakoun any other thoughts so I can batch changes?

docs/guides/cloud/account-management/projects.mdx
```diff
@@ -299,6 +299,35 @@ See also
...
+   - Link the GitHub integration to the Application code repo
+   - Override the `COMMIT_INFO_*` environment variables used by the test recording to reflect the state of the checked out Application code repo
+   - Record any runs related to the Application code repo to **App Project**
+   - Capture any Cypress config/tests repo related identifiers (short sha/release number) as tags decorated on the run
+
+3. For the **Test Project**:
+
+   - Link the GitHub integration to the Cypress config/tests repo
+   - Ensure any runs recorded to **Test Project** are related to test code changes in the Cypress config/tests repo
+     - the application code repo state that the tests are executed against here could likewise be captured as tags here, if necessary.
```
pstakoun, line 329:
```suggestion
     - The application code repo state that the tests are executed against here could likewise be captured as tags here, if necessary.
```

docs/guides/cloud/account-management/projects.mdx
pstakoun: I agree with making it more general and also mention Branch Review as an example of a feature that is impacted.

jaffrepaul: @pstakoun  I took out the direct mention of branch review in the last commit. Are you saying we should still leverage it in some way?

pstakoun: nah this is fine. I'd just maybe link to this section from within the branch review docs

PR #5563, opened 2023-11-15
docs(selectFile): readFile & fixture should use `null` instead of `{ encoding: null }`

ronny1020:
The `null` should not wrapped in the `{ encoding: null }`, which would cause a type error.

reference: https://docs.cypress.io/api/commands/fixture#Syntax


emilyrohrbough (approved):

emilyrohrbough:
@ronny1020 nice catch! thank you! 🙏🏻

PR #5585, opened 2023-12-20
Remove references to cypress's chromium dls

jennifer-shehane:
This service is no longer maintained: https://github.com/cypress-io/chromium-downloads


MikeMcC399:
https://vikyd.github.io/download-chromium-history-version/ appears not to show the categories 

- canary
- stable
- dev
- beta

like https://chromium.cypress.io/ did.

It looks like there isn't any convenient exact replacement for https://chromium.cypress.io/ 🙁

jennifer-shehane:
@MikeMcC399 Yah, there are several tools and other ways to track down upcoming versions. Those are likely not as necessary for our users, so I do think this tool addresses most of the user facing concerns.

PR #5596, opened 2024-01-08
Replace "irregardless" with "regardless".

AndreiRailean:
While "irregardless" _is_ a word, it is probably too quirky for international readers, likely used here by mistake.


jennifer-shehane (approved):

jennifer-shehane:
@AndreiRailean Thanks for the contribution!

PR #5601, opened 2024-01-09
update dependencies (install Docusaurus 3)

elevatebart:
- [x] Update all dependencies (prettier, docusaurus, remark, unist,... )
- [x] Rewrite the MDX plugins for compat with updated version of remark
- [x] Use sync version of prettier as prettier has become async
- [x] Remove all MDX1 hacks from the codebase

**Caveat**: some customized header hashes cannot be parsed in DS3. See [Yields](https://docs.cypress.io/api/commands/type#Yields) has disapeared [here](https://github.com/cypress-io/cypress-documentation/blob/06f89295902d4f9f40b455cab8af36d4c61da831/docs/api/actions/type.mdx#L98). The end result seems the same. so I am not worried.


jaffrepaul (approved):
I'm good with all this. Nice job working through the remark stuff 💪 

Two q's just for my edification:
- Why does the prettier need to work sync?
- I didn't quite understand the header hashes caveat you mentioned. Did something not translate in v3?

elevatebart:
> I'm good with all this. Nice job working through the remark stuff 💪
> 
> Two q's just for my edification:
> 
> * Why does the prettier need to work sync?
The `visit` function that remark is using to change the content and load the config files for instance needs to use `visitors` that are synchronous. The maintainer or remark and MDX is very convinced of it. If a function is async, any function that depends on it needs to be async too. So we cannot use `prettier.format` if it's async. Believe me I tried. Good thing there is a `@prettier/sync` library maintained by the prettier team already.
 
> * I didn't quite understand the header hashes caveat you mentioned. Did something not translate in v3?
Before, when an icon was in a header, the Icon name was made part of the ID. We went around it with adding stuff at the end of header lines.

```md
### Yields [<IconHelp>](https://example-docs.org/yields) {#Yields}
```

Problem is: on some pages the "acorn" parser did not like it one bit. I removed it where it was problematic and the hash have not changed. The one above for instance is still "Yields"

PR #5602, opened 2024-01-10
Remove comment from AAD cypress.env.json example

oflisback:
Removing unnecessary comment from json example which if included in the actual cypress.env.json file causes cypress to not start (only showing loading spinner) without error message.


jennifer-shehane (approved):

jennifer-shehane:
@oflisback Thanks for the contribution!

PR #5613, opened 2024-01-19
Update description of testFiles option

sjlehn:
Updated the description of the `testFiles` option to indicate that it is relative to the `integrationFolder` path. It took me a while to realize that this path wasn't relative to the project root.


jennifer-shehane:
@sjlehn integrationFolder is only available in Cypress versions prior to 10.0 in our legacy config: https://docs.cypress.io/guides/references/legacy-configuration

Issue #5614, opened 2024-01-24
Bun download fails with no Node.js


MikeMcC399:
### Subject

Installing Cypress > System requirements > Node.js

### Description

- Some users are attempting to install Cypress using Bun (see https://github.com/cypress-io/cypress/issues/28722) and if there is no Node.js installed Cypress produces an error message

> TypeError: undefined is not an object
> The Cypress App could not be downloaded.

- Cypress `13.6.3`
- Bun `1.0.25`
- Ubuntu 22.04

A pure [Bun](https://bun.sh/) environment has no [Node.js](https://nodejs.org/) present.

If Node.js is a firm prerequisite for Cypress, then the documentation section [Installing Cypress > System requirements > Node.js](https://docs.cypress.io/guides/getting-started/installing-cypress#Nodejs) should be updated to state that Node.js is required.

https://bun.sh/docs say:

> At its core is the Bun runtime, a fast JavaScript runtime designed as a drop-in replacement for Node.js.

so if Cypress is not compatible with installing Cypress using a pure Bun environment as the drop-in replacement for Node.js, then the documentation should possibly also mention Bun.

jennifer-shehane:
Added something to docs here: https://github.com/cypress-io/cypress-documentation/pull/5615

PR #5615, opened 2024-01-24
docs: Add note about Cypress requiring Node.js to install

jennifer-shehane:
Close https://github.com/cypress-io/cypress-documentation/issues/5614


MikeMcC399 (approved):
Thank you @jennifer-shehane 

- This resolves the issue #5614 without mentioning Bun
- Support of Bun can be made a separate topic / statement, if required

PR #5617, opened 2024-01-26
Fix bold markdown in retry-ability.mdx

p2635:
URL: https://docs.cypress.io/guides/core-concepts/

image

I just noticed some text that is not probably in bold on the live website, this is a minor PR to fix that.

I'm not sure if this fix is already in a separate branch. This is my first time contributing, so forgive me if I am missing any steps here.


jennifer-shehane (approved):
Thank you!

PR #5626, opened 2024-02-02
fix: design (and remove mistaken yarn.lock)

elevatebart:
- [x] remove jank when switching to dark mode
- [x] keep underline inside the code box when commands are linked
- [x] make the Table of Contents fully opaque on the right side to avoid contrast issues


jaffrepaul (commented):
Two q's:

- What was the jank with dark mode? I didn't notice a difference between this and prod.
- I'm not getting the underline on hover for linked `code` any more, was that intentional?

ryanjwilke:
@elevatebart This looks good. I would be in favor of doing an `underline` on hover for all links. At least links inside the content area. That would be fine.

@jaffrepaul Here's a video of the "jank" Bart is referring to where you see the elements shifting in the content area when you toggle the light and dark mode.

https://github.com/cypress-io/cypress-documentation/assets/211597/541b53b6-1705-41ed-8eae-2cb52301020d

elevatebart:
The link thing was not intentional.
Must have been a glitch between dev and prod. I will fix it and resubmit.

jaffrepaul (changes requested):
There's a slight difference now between `code` that have anchor links inside of the admonitions vs. regular page content. 

I recall wrestling with this last time around before I landed on that current implementation. But you can see the underline is happening on the code display (in a bit of a funny way) in the admonition, but has no hover state otherwise. 

Ideally these are consistent with a hover state that keeps the affordance that these are clickable. https://share.cleanshot.com/zmv2JR11

jaffrepaul (approved):

PR #5628, opened 2024-02-04
fix autoCancelAfterFailures in ModuleAPI

MikeMcC399:
This PR fixes the pipe (vertical bar) rendering in the `autoCancelAfterFailures` entry in the [Guides > Module API > Options](https://docs.cypress.io/guides/guides/module-api#Options) table. The `mdx` Markdown source was attempting to use the HTML Character Entity Number `&#124` for "vertical bar". The `|` character (which is a otherwise a table control character in Markdown) is escaped as `\|` instead for correct rendering.

~~Also, since `false` is a value, not a type, the type is changed to show `boolean` for consistency with the other table entries.~~

---
BEFORE
before autoCancelAfterFailures

---
AFTER
after autoCancelAfterFailures


MikeMcC399:
Preview change on https://deploy-preview-5628--benevolent-cat-040f48.netlify.app/guides/guides/module-api#Options

docs/guides/guides/module-api.mdx
```diff
@@ -56,31 +56,31 @@ node e2e-run-tests.js
...
-| `testingType`             | _string_               | Specify the type of tests to execute; either `e2e` or `component`. Defaults to `e2e`                                                                                                                     |
+| Option                    | Type                  | Description                                                                                                                                                                                              |
+| ------------------------- | --------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `autoCancelAfterFailures` | _number_ \| _boolean_ | Specify the number of failures to cancel a run being recorded to the Cloud or `false` to disable auto-cancellation.                                                                                      |
```
emilyrohrbough, line 61:
```suggestion
| `autoCancelAfterFailures` | _number_ \| _false_ | Specify the number of failures to cancel a run being recorded to the Cloud or `false` to disable auto-cancellation.                                                                                      |
```

emilyrohrbough: False is the correct option here if a number is not specified

MikeMcC399: Thanks for the correction. I was unfamiliar with the concept of TypeScript literal types.
I will revert to using `_false_` as a Type, so the only correction in this PR will be the rendering of the pipe symbol.

emilyrohrbough (approved):

emilyrohrbough:
Thanks Mike!

PR #5634, opened 2024-02-06
fix: document should update when path changes

elevatebart:


elevatebart:
@emilyrohrbough I fixed an error where when one clicks on a link it did not switch to the page.

It was actually a problem of not expanding sections containing active items.

emilyrohrbough:
@elevatebart The navigation is now correctly highlighting, but we have log the scrolling nav item into view behavior we previously had. See

https://github.com/cypress-io/cypress-documentation/assets/14099737/15f5bd0a-c817-4c23-9488-4e48d649a188

elevatebart:
Hello @emilyrohrbough I just updated docmenu with the proper scrolling behavior.

It should work better now.

emilyrohrbough (approved):

Issue #5630, opened 2024-02-04
Multiple broken anchors


MikeMcC399:
### Description

When I executed `npm run build` I expected no broken anchors to be reported.

### URL of Issue(s)

https://docs.cypress.io/

### Steps to replicate

```shell
git clone https://github.com/cypress-io/cypress-documentation
cd cypress-documentation
npm install
npm run build
```

### Additional Information

- PR https://github.com/cypress-io/cypress-documentation/pull/5601, merged on Jan 16, 2024, updated Docusaurus from `^2.4.1` to `3.1.1`.
- Broken anchor link checking is a new feature of [docusaurus@3.1.0](https://github.com/facebook/docusaurus/releases/tag/v3.1.0). This means the broken anchors were previously not detected.

Bad anchor links generally cause reader confusion, as they will typically call up the intended page and then display it at the wrong place, leaving the reader to search and / or scroll for the relevant information.

<details>
<summary>npm run build log (original state)</summary>

```text
[INFO] Docusaurus found broken anchors!

Please check the pages of your site in the list below, and make sure you don't reference any anchor that does not exist.
Note: it's possible to ignore broken anchors with the 'onBrokenAnchors' Docusaurus configuration, and let the build pass.

Exhaustive list of all broken anchors found:
- Broken anchor on source page path = /api/commands/children:
   -> linking to /guides/core-concepts/introduction-to-cypress#Default-Assertions
- Broken anchor on source page path = /api/commands/closest:
   -> linking to /guides/core-concepts/introduction-to-cypress#Default-Assertions
- Broken anchor on source page path = /api/commands/contains:
   -> linking to /faq/questions/using-cypress-faq#How-do-I-get-an-element-s-text-contents
   -> linking to /guides/core-concepts/introduction-to-cypress#Default-Assertions
- Broken anchor on source page path = /api/commands/eq:
   -> linking to /guides/core-concepts/introduction-to-cypress#Default-Assertions
- Broken anchor on source page path = /api/commands/filter:
   -> linking to /guides/core-concepts/introduction-to-cypress#Default-Assertions
- Broken anchor on source page path = /api/commands/find:
   -> linking to /guides/core-concepts/introduction-to-cypress#Default-Assertions
- Broken anchor on source page path = /api/commands/first:
   -> linking to /guides/core-concepts/introduction-to-cypress#Default-Assertions
- Broken anchor on source page path = /api/commands/fixture:
[...]

PR #5666, opened 2024-02-19
Remove instructions on running headless tests without Xvfb

timvdalen:
As already written in this bit of the docs, `ELECTRON_RUN_AS_NODE` was never tested.
Having tested it, I'm now sure it _doesn't_ work, and I think removing it from the docs is clearer than having it in there with this note.

Essentially, when `ELECTRON_RUN_AS_NODE` is set to `1`, the `Cypress` binary starts to act as if it's a `node` binary:

2024-02-19 09_23_19-Window

When that happens, when cypress tries to call it with set-up params, `node` responds that it does not know those params:

2024-02-19 09_23_55-Window

Perhaps this used to work in the past, but seeing the 'maybe' in the docs gave me some hope that I would be able to get cypress to run without Xvfb.

Also reported in https://github.com/cypress-io/cypress/issues/23636

### DIFF
--- a/docs/guides/continuous-integration/introduction.mdx
+++ b/docs/guides/continuous-integration/introduction.mdx
@@ -613,30 +613,6 @@ crashing.
 
 :::
 
-### Running headless tests without Xvfb
-
-:::caution
-
-Using the `ELECTRON_RUN_AS_NODE=1` env var is experimental and not fully tested,
-so may not work in all environments.
-
-:::
-
-Chromium based browsers and Firefox can spawn without Xvfb when run via the
-`--headless` flag. If you're testing against either of those browsers using the
-`--browser` flag, you can opt out of Cypress spawning an X11 server by setting
-the environment variable
-[`ELECTRON_RUN_AS_NODE=1`](https://www.electronjs.org/docs/api/environment-variables#electron_run_as_node).
-
-:::caution
-
-Electron cannot be run without an X11 server. Cypress's default browser is
-Electron and won't be able to launch if you set this environment variable.
-Likewise, Cypress's interactive mode (running via `cypress open`) is run via
-Electron and cannot be opened without an X11 server.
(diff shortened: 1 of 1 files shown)


MikeMcC399 (commented):
@timvdalen 

Removing this section entirely would break a link to the [cypress@8.0.0](https://docs.cypress.io/guides/references/changelog#8-0-0) changelog.

The changelog says:

-  Setting the environment variable `ELECTRON_RUN_AS_NODE` now starts Cypress as  a normal Node.js process rather than an Electron process. See [Running headless tests without Xvfb](/guides/continuous-integration/introduction#Running-headless-tests-without-Xvfb) for more details. Addresses [#16505](https://github.com/cypress-io/cypress/issues/16505).

The PR which introduced this feature was:

- https://github.com/cypress-io/cypress/pull/16537

You mentioned also

- https://github.com/cypress-io/cypress/issues/23636 which remains unresolved

The Cypress.io team needs to weigh in on this. It depends on whether there is hope that this feature can be fixed or whether it is permanently broken. If it is temporarily broken, then perhaps it would be better just to expand on the note in
[Continuous Integration > Introduction > Running headless tests without Xvfb](https://docs.cypress.io/guides/continuous-integration/introduction#Running-headless-tests-without-Xvfb) which says:

> Using the ELECTRON_RUN_AS_NODE=1 env var is experimental and not fully tested, so may not work in all environments.

MikeMcC399:
- If https://github.com/cypress-io/cypress/issues/19868 were to be resolved, that would ease the pressure about having a workaround for at least this one common scenario.

jennifer-shehane (approved):
We'll remove for now.

MikeMcC399:
- I'll add the new broken anchor link to https://github.com/cypress-io/cypress-documentation/issues/5630

PR #5730, opened 2024-02-28
fix: wrong args order in introduction-to-cypress.mdx

degletnour:
+ removed unused index arg.


jennifer-shehane (commented):
The first arg returned here is the index. Removing the index arg results in this code failing and not returning the el here.

PR #5740, opened 2024-03-05
fix: update calledwith assertions argument list syntax

### LINKED ISSUE #5736: Misleading argument description in .should('have.been.calledWith', ['arg1, 'arg2'])
### Subject

guides/references/assertions

### Description

Screenshot 2024-03-02 181146

https://docs.cypress.io/guides/references/assertions#Sinon-Chai

The current document states that the arguments should be passed in as an array, while the actual behavior expects a list of arguments.

For example, the following assertion will fail when running the test.

```js
        //wrong assertion with array
        cy.get('@open').should('have.been.calledOnceWithExactly', [
           `${Cypress.config('baseUrl')}/?space=e7bf21ca`,
           `inkstain-e7bf21ca`,
         ]);
```

Changing to argument list will pass.

```ts
        cy.get('@open').should(
          'have.been.calledOnceWithExactly',
          `${Cypress.config('baseUrl')}/?space=e7bf21ca`,
          `inkstain-e7bf21ca`
        );
```

t-benze:
This is to update the description regarding the calledWith assertion arguments syntax.

(Closes #5736)

### DIFF
--- a/docs/guides/references/assertions.mdx
+++ b/docs/guides/references/assertions.mdx
@@ -126,31 +126,31 @@ You will commonly use these chainers after using DOM commands like:
 These chainers are used on assertions with [`cy.stub()`](/api/commands/stub) and
 [`cy.spy()`](/api/commands/spy).
 
-| Sinon.JS property/method | Assertion                                                                                                                   |
-| ------------------------ | --------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


jennifer-shehane (approved):
Looks good. Main Sinon docs here: https://sinonjs.org/releases/v17/assertions/

PR #5741, opened 2024-03-06
Adding pixeleye to plugins.json

AlfieJones:
I've added [Pixeleye](https://pixeleye.io/home) to the plugin definitions. 

You can see an example of a cypress & pixeleye project [here](https://github.com/pixeleye-io/pixeleye/tree/main/examples/cypress).

I've added it with a community badge, but I would love to get it verified. What would be the steps to make that happen?

### DIFF
--- a/src/data/plugins.json
+++ b/src/data/plugins.json
@@ -1077,6 +1077,13 @@
           "badge": "verified"
         },
         {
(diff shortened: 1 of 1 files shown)


jennifer-shehane (approved):
Thanks!

PR #5758, opened 2024-03-21
fix formatting for "Install binary"

MikeMcC399:
## Issue

[References > Advanced Installation > Install binary](https://docs.cypress.io/guides/references/advanced-installation#Install-binary) is displaying `shell` as part of the command line, when it is supposed to be an `.MDX` formatting instruction, for instance:

> Install a version different than the default npm package. `shell CYPRESS_INSTALL_BINARY=2.0.1 npm install cypress@2.0.3`

This is an issue that was introduced unnoticed some years ago.

## Changes

Revert to the original use of `shell` as a triple backtick styler in the [References > Advanced Installation > Install binary](https://docs.cypress.io/guides/references/advanced-installation#Install-binary) section.

### DIFF
--- a/docs/guides/references/advanced-installation.mdx
+++ b/docs/guides/references/advanced-installation.mdx
@@ -25,11 +25,17 @@ Cypress is installed. To override what is installed, you set
 **This is helpful if you want to:**
 
 - Install a version different than the default npm package.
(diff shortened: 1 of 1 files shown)


jennifer-shehane (approved):
Yah, this is nicer. 

Before:
Screenshot 2024-03-21 at 11 26 11 AM


After:
Screenshot 2024-03-21 at 11 26 18 AM

Issue #5444, opened 2023-08-28
Operating System support lists vendor unsupported versions


MikeMcC399:
### Subject

[Getting Started > Installing Cypress > System requirements > Operating System](https://docs.cypress.io/guides/getting-started/installing-cypress#Operating-System)

### Description

[System requirements > Operating System](https://docs.cypress.io/guides/getting-started/installing-cypress#Operating-System) lists OS versions which are no longer under vendor support.

Should the current list below be updated to list only operating systems which are still supported by their respective vendors?

#### Current list

- macOS 10.9 and above (Intel or Apple Silicon 64-bit (x64 or arm64))
- Linux Ubuntu ~~12.04~~ 20.04 and above, Fedora 21 and Debian 8 (x86_64 or Arm 64-bit (x64 or arm64)) (see Linux Prerequisites down below)
- Windows ~~7~~ 10 and above (64-bit only)

## Vendor lifecycle information

Note: resolved OS entries in the list below are marked in the respective check-box:

- [x] According to https://endoflife.date/macos macOS 10.9 service status ended on Dec 1, 2016. macOS 12 is the earliest version still supported. Edit: updated to minimum macOS 10.15.

- [x] Ubuntu 20.04 is the earliest version still supported in its Standard Support phase according to https://ubuntu.com/about/release-cycle. Ubuntu 14.04, 16.04 & 18.04 are in their Extended Security Maintenance phase. Ubuntu 12.04 is now unsupported. Edit: Changed to minimum Ubuntu 20.04.

- [x] Fedora ~~37~~ 38 is the earliest version still supported and Fedora 21 entered end-of-life on Dec 1, 2015 according to https://docs.fedoraproject.org/en-US/releases/eol/.

- [x] Debian 8 is supported in Extended Long Term Support according to https://wiki.debian.org/LTS/Extended, however it is not compatible with Node.js 18. A minimum of Debian 10 is needed. Edit: changes to Debian 10.

- [x] Windows 7 last support date was January 2023 (see https://learn.microsoft.com/en-us/lifecycle/products/windows-7). Edit: Changed to minimum Windows 10.

- [ ] A Windows server system is missing from the list.

## GitHub current images

See [GitHub Actions Runner Images](https://github.com/actions/runner-images) for list of currently available images.

## Notes

https://endoflife.date/ is also a convenient non-primary source for many end-of-life dates.

jaffrepaul:
[...]

PR #5766, opened 2024-03-26
Update requirements to Debian 10 and above

MikeMcC399:
- relates to issue #5444

This PR updates the Debian version in the list of supported operating systems under the section [Getting Started > Installing Cypress > System requirements > Operating System](https://docs.cypress.io/guides/getting-started/installing-cypress#Operating-System)

from

- Debian 8 (x86_64 or Arm 64-bit (x64 or arm64))

to

- Debian 10 (x64 or arm64) and above

## Background

[Getting Started > Installing Cypress > System requirements > Node.js](https://docs.cypress.io/guides/getting-started/installing-cypress#Nodejs) lists Node.js `18`, `20` and above as being supported.

The [Node.js 18 announcement](https://nodejs.org/en/blog/announcements/v18-release-announce) provides detailed information in the [Toolchain and Compiler Upgrades](https://nodejs.org/en/blog/announcements/v18-release-announce#toolchain-and-compiler-upgrades) section which lists:
> Prebuilt binaries for Linux are now built on Red Hat Enterprise Linux (RHEL) 8 and are compatible with Linux distributions based on glibc 2.28 or later, for example, Debian 10, RHEL 8, Ubuntu 20.04.

Executing `ldd --version` on Debian `8.11.1` (jessie) shows that GLIBC `2.19` is installed, which is below GLIBC `2.28` or later required by Node.js `18`. Attempting to run `node` or `npm` from Node.js `18.19.1` LTS on Debian `8.11.1` (jessie) fails with fatal error messages related to the GLIBC version requirements not being fulfilled.

[Debian “buster” (10)](https://www.debian.org/releases/buster/) is therefore the earliest version which is supported by Node.js `18` and is compatible with the prebuilt binaries of Node.js `18`. Debian `10` is under [Long Term Support](https://wiki.debian.org/LTS) until June 30, 2024.

### DIFF
--- a/docs/guides/getting-started/installing-cypress.mdx
+++ b/docs/guides/getting-started/installing-cypress.mdx
@@ -152,8 +152,7 @@ Cypress is a desktop application that is installed on your computer. The desktop
 application supports these operating systems:
 
 - **macOS** 10.9 and above _(Intel or Apple Silicon 64-bit (x64 or arm64))_
-- **Linux** Ubuntu 20.04 and above, Fedora 21 and Debian 8 _(x86_64 or Arm
-  64-bit (x64 or arm64))_ (see [Linux Prerequisites](#Linux-Prerequisites) down
+- **Linux** Ubuntu 20.04 and above, Fedora 21 and Debian 10 and above _(x64 or arm64)_ (see [Linux Prerequisites](#Linux-Prerequisites) down
   below)
 - **Windows** 10 and above _(64-bit only)_
 


MikeMcC399:
Preview

https://deploy-preview-5766--benevolent-cat-040f48.netlify.app//guides/getting-started/installing-cypress#Operating-System

jennifer-shehane (approved):
@MikeMcC399 Thanks!

PR #5779, opened 2024-04-02
Update example CYPRESS_INSTALL_BINARY

### LINKED ISSUE #5762: Non-working advanced installation example for mismatched binary
### Subject

[References > Advanced Installation > Install binary](https://docs.cypress.io/guides/references/advanced-installation#Install-binary)

## Issue

[References > Advanced Installation > Install binary](https://docs.cypress.io/guides/references/advanced-installation#Install-binary) uses [cypress@2.0.1](https://docs.cypress.io/guides/references/changelog#2-0-1) and [cypress@2.0.2](https://docs.cypress.io/guides/references/changelog#2-0-2) as an example in the usage of the `CYPRESS_INSTALL_BINARY` environment variable.

> Install a version different than the default npm package.
    `CYPRESS_INSTALL_BINARY=2.0.1 npm install cypress@2.0.3`
  
1. Cypress `2.x` are legacy versions [released](https://docs.cypress.io/guides/references/changelog#2-1-0) in 2018. These are not compatible with currently supported Ubuntu environments. Cypress `2.0.3` does not run on Ubuntu `20.04.06` LTS with Node.js `18.x` (the lowest currently supported versions) and fails to start. See also https://github.com/cypress-io/cypress-documentation/pull/5528 which identified [Cypress `3.4.1`](https://docs.cypress.io/guides/references/changelog#3-4-1) as the last version to require the obsolete `libgconf-2-4`, so essentially [Cypress `3.5.0`](https://docs.cypress.io/guides/references/changelog#3-5-0) is the lowest version which can be used in any examples.

2. The use-case for installing a Cypress binary version which does not match the version of the Cypress npm module is unclear and not explained.

### Remove

If there is no valid use-case for installing a Cypress binary version which does not match the version of the Cypress npm module, then remove the example.

### Update

1. If there is a valid use-case for this installation option, then the use-case should be explained.
2. If the example is to remain in the documentation, then update to current non-legacy `13.x` versions.

#### Test of update
Executing the following on Ubuntu `22.04` causes the Cypress binary `13.7.1` to be placed in the cache folder for `13.7.0`.

```shell
CYPRESS_INSTALL_BINARY=13.7.1 npm install cypress@13.7.0
```

The cache folder created is `~/.cache/Cypress/13.7.0`.

Running Cypress with `npx cypress open` results in a warning

```text
$ npx cypress open
Found binary version 13.7.1 installed in: /home/mike/.cache/Cypress/13.7.0/Cypress

⚠ Warning: Binary version 13.7.1 does not match the expected package version 13.7.0

  These versions may not work properly together.
```

Judging by the warning, this is not something which should be done, making it questionable whether the example should remain in the documentation at all, even if the version is updated to a current non-legacy version.

MikeMcC399:
- closes #5762

## Issue

The example in [References > Advanced Installation > Install binary](https://docs.cypress.io/guides/references/advanced-installation#Install-binary) uses [cypress@2.0.1](https://docs.cypress.io/guides/references/changelog#2-0-1) and [cypress@2.0.2](https://docs.cypress.io/guides/references/changelog#2-0-2) shows the usage of the `CYPRESS_INSTALL_BINARY` environment variable.

> Install a version different than the default npm package.
```shell
CYPRESS_INSTALL_BINARY=2.0.1 npm install cypress@2.0.3
```

Cypress `2.x` are legacy versions which are not compatible with [currently supported versions of Ubuntu](https://docs.cypress.io/guides/getting-started/installing-cypress#Operating-System) where a minimum of Ubuntu `20.04` is described.

## Change

The example is updated to

```shell
CYPRESS_INSTALL_BINARY=13.7.0 npm install cypress@13.7.1
```

## Comments

Caution: using this optional method will result in a warning. In the above case this will be:

```text
⚠ Warning: Binary version 13.7.0 does not match the expected package version 13.7.1

  These versions may not work properly together.
```

Executing the above installation command causes the Cypress binary named under `CYPRESS_INSTALL_BINARY` to be cached under the version directory using the version from `npm install`. This can affect other projects intending to use the same `npm install` version. They will then also use the alternate version from `CYPRESS_INSTALL_BINARY`.

To reset the cache so that the binary version matches the cache directory name, execute

```shell
npx cypress install
```

### DIFF
--- a/docs/guides/references/advanced-installation.mdx
+++ b/docs/guides/references/advanced-installation.mdx
@@ -26,7 +26,7 @@ Cypress is installed. To override what is installed, you set
 
 - Install a version different than the default npm package.
   ```shell
-  CYPRESS_INSTALL_BINARY=2.0.1 npm install cypress@2.0.3
+  CYPRESS_INSTALL_BINARY=13.7.0 npm install cypress@13.7.1
   ```
 - Specify an external URL (to bypass a corporate firewall).
   ```shell


MikeMcC399:
Preview

https://deploy-preview-5779--benevolent-cat-040f48.netlify.app//guides/references/advanced-installation#Install-binary

PR #5790, opened 2024-04-09
Analytics and API

danjohansenconsulting:
Baseline structure for two new features about to deploy in Cypress Cloud:  enterprise reporting and data extract APIs.  This PR reorganizes the analytics docs pages into a nested structure.  The new features are marked as "coming soon" and the full capabilities / details will be added later this week.  This PR will secure the URL links and the Cloud product is going to be coded to reference those via on links.


netlify.toml
```diff
@@ -319,7 +319,7 @@
 
 [[redirects]]
   from = "/guides/dashboard/analytics"
-  to = "/guides/cloud/analytics"
+  to = "/guides/cloud/analytics/overview"
```
elevatebart, line 322: we should add `guides/cloud/analytics` to this to avoid unexpected 404s

```suggestion
  to = "/guides/cloud/analytics/overview"
  status = 301
  force = true
  
 [[redirects]]
  from = "/guides/cloud/analytics"
  to = "/guides/cloud/analytics/overview"
```

elevatebart (changes requested):
Super quick change necessary to avoid unwanted 404.

elevatebart (approved):

PR #5791, opened 2024-04-09
Add cypress-runner-themes to Cypress plugins page

dingraham:
[cypress-runner-themes](https://github.com/dingraham/cypress-runner-themes) is a Cypress plugin that provides alternative themes for the Cypress Test Runner

It replaces the now deprecated `cypress-dark` plugin since it supports Cypress v10+

Thanks!

### DIFF
--- a/src/data/plugins.json
+++ b/src/data/plugins.json
@@ -157,6 +157,13 @@
           "badge": "official"
         },
         {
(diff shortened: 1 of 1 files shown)


jennifer-shehane (approved):
Thanks! This looks like a great tool for accessibility.

Issue #5720, opened 2024-02-26
Docusaurus reports false positive anchor link errors for Plugins page


MikeMcC399:
### Subject

[Plugins](https://docs.cypress.io/plugins)

### Description

- Following on from the resolved issue https://github.com/cypress-io/cypress-documentation/issues/5630, all broken anchor errors reported by Docusaurus `npm run build` have been fixed, with the exception of false positives related to bookmarks on the [Plugins](https://docs.cypress.io/plugins) page https://docs.cypress.io/plugins.

For example

```text
- Broken anchor on source page path = /api/cypress-api/custom-commands:
   -> linking to /plugins#custom-commands
```

This is however a false positive, as the link https://docs.cypress.io/plugins#custom-commands brings up the correct page and the correct section. The target HTML is valid:

```html
<h2 id="custom-commands">Custom Commands</h2>
```

## Logs

```
[INFO] Docusaurus found broken anchors!

Please check the pages of your site in the list below, and make sure you don't reference any anchor that does not exist.
Note: it's possible to ignore broken anchors with the 'onBrokenAnchors' Docusaurus configuration, and let the build pass.

Exhaustive list of all broken anchors found:
- Broken anchor on source page path = /api/cypress-api/custom-commands:
   -> linking to /plugins#custom-commands
- Broken anchor on source page path = /api/cypress-api/custom-queries:
   -> linking to /plugins#custom-commands
- Broken anchor on source page path = /api/table-of-contents:
   -> linking to /plugins#custom-commands
- Broken anchor on source page path = /examples/recipes:
   -> linking to /plugins#authentication
- Broken anchor on source page path = /faq/questions/using-cypress-faq:
   -> linking to /plugins#email
   -> linking to /plugins#preprocessors
- Broken anchor on source page path = /guides/core-concepts/writing-and-organizing-tests:
   -> linking to /plugins#preprocessors
- Broken anchor on source page path = /guides/overview/why-cypress:
   -> linking to /plugins#visual-testing
- Broken anchor on source page path = /guides/references/best-practices:
   -> linking to /plugins#email
- Broken anchor on source page path = /guides/tooling/visual-testing:
   -> linking to /plugins#visual-testing
```

MikeMcC399:
- Thanks to @jaffrepaul for fixing this issue in PR #5796 💪🏻 !

PR #5796, opened 2024-04-11
add plugin anchor workaround

jaffrepaul:
Docusaurus is flagging these anchors as broken because of a timing issue as the plugin page is dynamically generated. We've set the build process to fail if legit broken anchors are caught. This workaround swaps markdown links with anchor tags which avoid flagging them as broken. Neither a glorious or long term solution but one that bandwidth allows and will improve upon broken anchors not making it into a build.


MikeMcC399:
- Related to issue https://github.com/cypress-io/cypress-documentation/issues/5720

PR #5824, opened 2024-05-09
updated new-spec-added-confirmation.png and spec-list-with-new-spec.png

Oliver-Looney:
Updated: https://docs.cypress.io/guides/end-to-end-testing/writing-your-first-end-to-end-test#Add-a-test-file

1. Updated text that said default test would fail when it passes
2. Updated screenshot showing default spec
3. Updated surrounding screenshots to match cypress & chrome versions

<table>
<tr>
 <td>Before</td>
 <td>After</td>
<tr>
 <td>
Screenshot 2024-05-09 at 11 44 34
</td>
 <td>
Screenshot 2024-05-09 at 11 44 45
</td>
</tr>
<tr>
 <td>
Screenshot 2024-05-09 at 11 06 07
</td>
 <td>
Screenshot 2024-05-09 at 11 06 51
</td>
<tr>
 <td>
Screenshot 2024-05-09 at 11 08 25
</td>
 <td>
Screenshot 2024-05-09 at 11 07 57
</td>
</table>


MikeMcC399:
@Oliver-Looney 

You're right that this section is out of date!

Also the text has changed in the previous screen. It no longer says "Create new empty spec" and instead it says just "Create new spec", so it would make sense to also replace the screenshot for https://docs.cypress.io/img/guides/end-to-end-testing/writing-your-first-end-to-end-test/create-new-empty-spec.png.

image

jennifer-shehane (approved):
Yah, it did used to do this, but we updated it since then since that was kind of an awful experience. Thanks for updating the docs!

PR #5840, opened 2024-06-06
Update writing-your-first-end-to-end-test.mdx

gilzow:
Updates test description to match previous test specifics against example.cypress.io

Current version of example.cypress.io/commands/actions (as of 20240606 - see screenshot below) does not contain an input element with a `data-testid` attribute. Additionally, from the test we built previously, we're targeting an element with a class of `action-email`.

Updates the description section to reflect the test as written.

image


jennifer-shehane:
Thanks for the contribution!

Issue #5859, opened 2024-07-01
Remove centOS 7 references


MikeMcC399:
### Subject

Guides

### What would you like?

Review and remove references to unsupported CentOS Linux 7.

- https://docs.cypress.io/guides/cloud/test-replay#Troubleshooting
- https://docs.cypress.io/guides/getting-started/installing-cypress#CentOS

### Why is this needed?

Open source community support has now ceased:

https://blog.centos.org/2023/04/end-dates-are-coming-for-centos-stream-8-and-centos-linux-7/

> **CentOS Linux 7 End of Life: June 30, 2024**
>
> RHEL 7 reaches [End of Maintance on June 30, 2024](https://access.redhat.com/support/policy/updates/errata#Life_Cycle_Dates). After June 30, 2024, there will be no updates published for CentOS Linux 7.

MikeMcC399:
Possibly references to CentOS should be replaced by references to CentOS Stream? Or perhaps it would be better to focus on Fedora and/or RHEL?

Fedora, CentOS Stream and RHEL are interrelated, see [What's Great About CentOS Stream](https://blog.centos.org/2021/12/introducing-centos-stream-9/).

MikeMcC399:
Just as a rough guide to how often each Linux OS is mentioned, here is the result from a search of open issues on https://github.com/cypress-io/cypress/issues:

| OS           | Number |
| ------------ | ------ |
| Linux        | 172    |
| Ubuntu       | 146    |
| Debian       | 36     |
| Arch Linux   | 14     |
| CentOS       | 6      |
| Fedora       | 5      |
| RHEL         | 4      |
| Alpine Linux | 6      |
| Amazon Linux | 1      |
| FreeBSD      | 1      |
| suse         | 0      |

jennifer-shehane:
Yah i haven't seen this centos issue reported in a while. It used to come up a LOT which necessitated adding it to the docs. It's likely ok to remove?

MikeMcC399:
Submitted the following two PRs for removal:

- #5860 - centOS Stream 9 works out of the box with Cypress, so it's not necessary to specify any additional packages to install
- #5861 - the troubleshooting steps only reference CentOS 7 and 8, and both these versions are now unsupported

MikeMcC399:
Resolved

PR #5860, opened 2024-07-02
Remove obsolete centOS Linux prerequisites

MikeMcC399:
- contributes to resolution of issue #5859

## Issue

CentOS Linux 7 reached end of life (EOL) on June 30, 2024 (see [What to know about CentOS Linux EOL](https://www.redhat.com/en/topics/linux/centos-linux-eol)).

CentOS is listed on [Getting Started > Installing Cypress > System requirements > Linux Prerequisites > CentOS](https://docs.cypress.io/guides/getting-started/installing-cypress#CentOS). Cypress cannot support operating systems which are themselves no longer supported. Additionally [Node.js states](https://nodejs.org/en/blog/announcements/v18-release-announce#toolchain-and-compiler-upgrades)
> Node.js does not support running on operating systems that are no longer supported by their vendor.

## Change

Remove the obsolete CentOS section [Getting Started > Installing Cypress > System requirements > Linux Prerequisites > CentOS](https://docs.cypress.io/guides/getting-started/installing-cypress#CentOS).

Although RedHat does not consider CentOS Stream to be the replacement for CentOS (the goals are different), I did test CentOS Stream 9. On [CentOS Stream release 9](https://centos.org/centos-stream/) Cypress `13.13.0` works out of the box. It is only necessary to install Node.js LTS then the full set of scaffolded Cypress specs runs without error. Therefore there is no longer any need to specify prerequisites for the current CentOS (Stream) release.

### DIFF
--- a/docs/guides/getting-started/installing-cypress.mdx
+++ b/docs/guides/getting-started/installing-cypress.mdx
@@ -223,12 +223,6 @@ apt-get install libgtk2.0-0t64 libgtk-3-0t64 libgbm-dev libnotify-dev libnss3 li
 pacman -S gtk2 gtk3 alsa-lib xorg-server-xvfb libxss nss libnotify
 ```
 
-#### CentOS
-
-```shell
-yum install -y xorg-x11-server-Xvfb gtk2-devel gtk3-devel libnotify-devel GConf2 nss libXScrnSaver alsa-lib
-```
-
 #### Amazon Linux 2023
 
 ```shell


MikeMcC399:
Preview

https://github.com/MikeMcC399/cypress-documentation/blob/remove/centos-prereq/docs/guides/getting-started/installing-cypress.mdx#linux-prerequisites

jennifer-shehane (approved):

PR #5880, opened 2024-07-26
docs: Note that GHA doesn't support any caching for pnpm

bmulholland:
Ideally there'd be an official recommendation or something here, but of course I can't speak to that.

See also https://github.com/cypress-io/github-action/issues/1044

### DIFF
--- a/docs/guides/continuous-integration/github-actions.mdx
+++ b/docs/guides/continuous-integration/github-actions.mdx
@@ -219,8 +219,9 @@ When working with actions that have multiple jobs, it is recommended to have an
 initial "install" job that will download any dependencies and build your app,
 and then cache these assets for use later by subsequent jobs.
 
-The Cypress GitHub Action will automatically cache and restore your Node
-dependencies for you.
+For Yarn and npm, the Cypress GitHub Action will automatically cache and restore your Node
+dependencies for you. The Cypress Github Action does not do any caching for pnpm, and
+[advises not to use any caching of any form](https://github.com/cypress-io/github-action?tab=readme-ov-file#pnpm).
 
 For build assets, you will need to cache and restore them manually.
 


MikeMcC399:
This may be a bug rather than missing functionality. It needs further investigation.

- The original PR https://github.com/cypress-io/github-action/pull/586 intended to cache dependencies. Possibly there has been a change since the original support of pnpm, perhaps in the caching mechanisms on the side of GitHub or on the side of pnpm affecting how dependencies are stored.

bmulholland:
AFAICT the issue is that, while deps are cached, Cypress installs browsers in a place that is cached. The official cypress.io recommendation is to cache that browser install location, too, but then the two can get out of sync. When they do, you may have the package cache, but not the browser cache, and then the job would fail.

There's a silly-obvious approach of "give me a command to run that fetches the browsers if they don't exist," which is what playwright has. Cypress seems fixed on maintaining the magical "npm install"-only experience, and that leads to this problem.

MikeMcC399:
@bmulholland 

- Cypress doesn't install browsers.

- The `cypress-io/github-action` can use browsers made available through GitHub Actions images, such as [ubuntu-22.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#browsers-and-drivers), through Docker images such as [cypress/browsers](https://github.com/cypress-io/cypress-docker-images/tree/master/browsers) or browsers installed by other means.

- I've opened a new issue https://github.com/cypress-io/github-action/issues/1227 and hope that we can make some progress there in providing guidance on caching in pnpm environments.

bmulholland:
Yeah sorry, I mis-typed. The "browser" part is actually the Cypress binary. The substance of my comment remains the same.

MikeMcC399:
@bmulholland 

> Yeah sorry, I mis-typed. The "browser" part is actually the Cypress binary. The substance of my comment remains the same.

Perhaps you could edit your post and substitute the correct terms in that case? I may have misunderstood you even given the substitution, but `pnpm cypress install` is the command that will install the Cypress binary if it is missing.

A caching recommendation has been re-instated. See 
https://github.com/cypress-io/github-action/blob/master/README.md#pnpm

MikeMcC399:
This PR adds text to an example using npm, not pnpm, so it would need more explanation to cover npm, Yarn and pnpm at this point.

I would defer to https://github.com/cypress-io/github-action/blob/master/README.md for all detailed information and try to keep the description on the docs site simple.

The comment about cache advisory is no longer true:

```markdown
The Cypress Github Action does not do any caching for pnpm, and
[advises not to use any caching of any form](https://github.com/cypress-io/github-action?tab=readme-ov-file#pnpm).
```

My preference would be to not pursue this PR further.

PR #5881, opened 2024-07-26
docs: Note for CI missing docs that pnpm isn't supported

bmulholland:
See also https://github.com/cypress-io/cypress-documentation/pull/5880

### DIFF
--- a/docs/guides/continuous-integration/introduction.mdx
+++ b/docs/guides/continuous-integration/introduction.mdx
@@ -370,6 +370,11 @@ minutes off install time by preventing a large binary download.
   Cypress caches. Consider using `yarn install --frozen-lockfile` as an
   [`npm ci`](https://docs.npmjs.com/cli/ci) equivalent.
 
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
@MikeMcC399 thoughts?

MikeMcC399:
@jennifer-shehane 

This PR was proposing to make changes to https://docs.cypress.io/guides/continuous-integration/introduction#Caching. 
The section does not cover specifics for the Cypress JavaScript GitHub Action (`cypress-io/github-action`) and is intended for general use.

- There was never any restriction on caching across builds with pnpm **without** `cypress-io/github-action`.

- The temporary restriction on manually caching the pnpm store across builds was lifted with the resolution of https://github.com/cypress-io/github-action/issues/1227. The action `cypress-io/github-action` in any case caches the Cypress binary and the npm cache across builds and this cannot be disabled.

I suggest to close this PR.

PR #5886, opened 2024-08-01
Improve WebKit linux dependencies instruction integration

MikeMcC399:
## Issue

On [Guides > Launching Browsers](https://docs.cypress.io/guides/guides/launching-browsers) the [Linux Dependencies](https://docs.cypress.io/guides/guides/launching-browsers#Linux-Dependencies) are separated from the [WebKit (Experimental)](https://docs.cypress.io/guides/guides/launching-browsers#WebKit-Experimental) installation instructions which need them.

## Change

Move [Linux Dependencies](https://docs.cypress.io/guides/guides/launching-browsers#Linux-Dependencies) up to belong with the installation instructions and refer to them explicitly.

PR #5892, opened 2024-08-06
docs: correct WebKit dependencies installation order Linux

MikeMcC399:
## Issue

- The documentation change for [WebKit (Experimental)](https://docs.cypress.io/guides/guides/launching-browsers#WebKit-Experimental) in PR https://github.com/cypress-io/cypress-documentation/pull/5886 concerning Linux dependencies may lead to problems with changing versions.
- The Playwright dependencies must be installed with the same version of Playwright as used for `playwright-webkit`

## Change

- Change the installation order to first install `playwright-webkit` then install Playwright dependencies for Linux.
- Add the information that the dependencies must be reinstalled if the version of `playwright-webkit` is changed.

### DIFF
--- a/docs/guides/guides/launching-browsers.mdx
+++ b/docs/guides/guides/launching-browsers.mdx
@@ -178,12 +178,12 @@ these steps:
 
 1. Add `experimentalWebKitSupport: true` to your
    [configuration](/guides/references/configuration) to enable the experiment.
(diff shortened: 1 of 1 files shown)


MikeMcC399:
This change was based on a misconception and needs more research. Unfortunately the Playwright documentation is not very forthcoming. I may resubmit if appropriate.

PR #5896, opened 2024-08-16
fix: code block with incorrect language

sumimakito:
According to the context, the below code block should be for a JSON file. However, it is rendered as `cypress.config.js/ts`:

Screenshot 2024-08-16 at 14 44 28

### DIFF
--- a/docs/guides/tooling/reporters.mdx
+++ b/docs/guides/tooling/reporters.mdx
@@ -226,9 +226,7 @@ Then add the separate `reporter-config.json` file (defined in your
 configuration) to enable `spec` and `junit` reporters and direct the `junit`
 reporter to save separate XML files.
 
-:::cypress-config-example
-
-```ts
+```json
 {
   reporterEnabled: 'spec, mocha-junit-reporter',
   mochaJunitReporterReporterOptions: {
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@sumimakito 

There is special pre-processing used on this page and your change fails formatting checks, so I suspect that it may not be necessary to change. You could check how `:::cypress-config-example` is supposed to work.

sumimakito:
Hi @MikeMcC399, thanks for responding. It turned out that it was caused by the incorrect format in the JSON code block (unquoted keys and single quotes around values). I've fixed the issue and updated the PR.

From [the documentation](https://github.com/cypress-io/cypress-documentation/blob/04b99f007c25f100811583705970a2bb5ec5cd49/plugins/cypressRemarkPlugins/src/cypressConfigExample/readme.md), `:::cypress-config-example` directive wraps a JavaScript object with `defineConfig(...)` and marks it as a default export, but in this PR, I was intended to render the JSON configuration with a proper JSON code block instead of a JavaScript/TypeScript block rendered by the `:::cypress-config-example` directive.

MikeMcC399:
@sumimakito 

I understand now that you are correcting an issue in https://docs.cypress.io/guides/tooling/reporters#Command-line-3 which is supposed to display an example `reporter-config.json` and this example was wrongly generated / formatted. Sorry for my misunderstanding!

sumimakito:
@MikeMcC399 No problem. Thanks for helping check the CI results anyway!

jennifer-shehane (approved):
@sumimakito Thanks for the fix!

Issue #5412, opened 2023-08-04
CT Supported Frameworks with wildcard "+" version


MikeMcC399:
### Subject

Component Testing > Overview > Supported Frameworks

### Description

[Supported Frameworks](https://docs.cypress.io/guides/component-testing/overview#Supported-Frameworks) lists framework versions with a base version and a "+" to indicate later versions. 

![image](https://github.com/cypress-io/cypress-documentation/assets/66998419/345be5d5-c487-4f7b-9dfc-7498d33123d9)

This is not consistent with the way that Cypress detects frameworks, which works on discrete versions, rather than minimum versions.

~~For instance, Svelte with Vite lists `Svelte 3+`, which implies also Svelte 4, however Cypress `12.17.3` displays a message saying that svelte `4.1.2` is not supported.~~ Edit: Outdated comment, since Svelte 4 is now supported. The general problem raised is however still relevant.

![image](https://github.com/cypress-io/cypress/assets/66998419/b517749d-bdfb-47ea-ae37-5fe2788f0192)

The version list should be aligned to the way that Cypress determines which framework versions are supported. If there is a minimum and maximum version permitted, then this should be listed in the table.

Currently the table is misleading.

MikeMcC399:
Cypress source code [packages/scaffold-config/src/dependencies.ts](https://github.com/cypress-io/cypress/blob/develop/packages/scaffold-config/src/dependencies.ts) is related to the framework dependencies which are supported.

jordanpowell88:
Are you proposing we document the versions differently? If so, what do you suggest?

MikeMcC399:
@jordanpowell88 

The suggestion would be to replace the "+" notation with version ranges, for example

Create React App 4+ | React 16+ | Webpack 4+

would become

Create React App 4-5  | React 16-18 | Webpack 4-5

so that the table mirrors the versions which the Cypress app displays during scaffolding.

Svelte 3+ would be changed to Svelte 3 and when 4 is supported it would be changed to Svelte 3-4.

MikeMcC399:
The table, with no upper bound for supported versions, continues to be misleading as new frameworks are released without Cypress providing support.

Currently the following issues are requesting updated support:

- https://github.com/cypress-io/cypress/issues/28347
- https://github.com/cypress-io/cypress/issues/28185

PR #5899, opened 2024-08-18
Replace CT Supported Frameworks "+" notation with fixed version ranges

MikeMcC399:
- Closes https://github.com/cypress-io/cypress-documentation/issues/5412

## Issue

[Supported Frameworks](https://docs.cypress.io/guides/component-testing/overview#Supported-Frameworks) lists framework versions with a base version and a "+" to indicate later versions.

This is not consistent with the way that Cypress detects frameworks, which works on discrete versions, rather than minimum versions. Versions which are not verified by Cypress may not work. Leaving the highest supported version open is misleading.

## Change

The table [Supported Frameworks](https://docs.cypress.io/guides/component-testing/overview#Supported-Frameworks) is reworked to conform to the supported versions specified in https://github.com/cypress-io/cypress/blob/develop/packages/scaffold-config/src/dependencies.ts.

### DIFF
--- a/docs/guides/component-testing/overview.mdx
+++ b/docs/guides/component-testing/overview.mdx
@@ -101,19 +101,19 @@ Cypress currently has official mounting libraries for
 [Svelte](/guides/component-testing/svelte/overview) and support for the
 following development servers and frameworks:
 
-| Framework                                                                                                             | UI Library  | Bundler    |
-| --------------------------------------------------------------------------------------------------------------------- | ----------- | ---------- |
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Preview

https://github.com/MikeMcC399/cypress-documentation/blob/ct-framework-versions/docs/guides/component-testing/overview.mdx#supported-frameworks

MikeMcC399:
@jennifer-shehane 

> Thanks!

You're welcome! I held back on attempting to resolve the associated issue since it was assigned to @jaffrepaul , however it seems that he is no longer available.

PR #5913, opened 2024-09-04
Change name from Run Cancellation to Auto Cancellation

danjohansenconsulting:
this was some old verbiage still in-use and we call it auto cancellation in all other places


jennifer-shehane (approved):
You intending to update the actual slug?

danjohansenconsulting:
> You intending to update the actual slug?

Not positive yet.  That was my initial plan but the more I look at it the url of run-cancellation is still okay.  It's just our name of the feature in Cloud that is most critical to correct.  

Either way, it was going to be a 2-step deploy if we did change the URL because we'd have to update the https://on.cypress.io/auto-cancellation onlink also

PR #5942, opened 2024-10-06
docs: update plugin.json to include cypress-accessibility

rowellx68:

PR #5950, opened 2024-10-11
docs: rename cypress-accessibility plugin to cypress-a11y-report

rowellx68:
Renamed the plugin to avoid any confusion between it and the official offering from Cypress. See PR https://github.com/cypress-io/cypress-documentation/pull/5942 for reference.


jennifer-shehane (approved):
@rowellx68 Thanks!

PR #5955, opened 2024-10-17
docs(cy.stub): Update syntax to use callsFake() instead of deprecated third argument

sadie100:
While using Cypress, I encountered a deprecated notice when following the official documentation for `cy.stub(object, method, replacerFn)`. Upon investigation, I found that [this syntax has been deprecated in Sinon.JS since version 3.0.0](https://sinonjs.org/releases/v19/stubs/). As Cypress is currently using Sinon.JS version 7.3.2(as per package.json), I determined that this was a deprecated specification.

So I made changes in document:
1. Removed the deprecated `cy.stub(object, method, replacerFn)` syntax from the documentation.
2. Updated related examples to use the new recommended syntax `cy.stub(object, method).callsFake(replacerFn)`.

These changes will help Cypress users write more up-to-date and future-proof test code, avoiding potential deprecation warnings or errors.

Please review and let me know if any further changes or clarifications are needed.

### DIFF
--- a/docs/api/commands/stub.mdx
+++ b/docs/api/commands/stub.mdx
@@ -19,7 +19,6 @@ assertion. It is not retryable, chainable, or timeout-able.
 ```javascript
 cy.stub()
 cy.stub(object, method)
-cy.stub(object, method, replacerFn)
 ```
 
 ### Usage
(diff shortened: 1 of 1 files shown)


MikeMcC399:
- This PR seems to be an duplicate of PR https://github.com/cypress-io/cypress-documentation/pull/5592 (which apparently was never reviewed).

PR #5971, opened 2024-10-25
More cleanup

jennifer-shehane:
- Do not bold some changelog entries in UI Cov
- Reroute ui-coverage to main ui-cov doc

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1106,6 +1106,12 @@
   force = true
 
 [[redirects]]
+  from = "/accessibility"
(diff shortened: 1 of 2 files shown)


netlify.toml
```diff
@@ -1099,6 +1099,12 @@
   status = 301
   force = true
 
+[[redirects]]
```
marktnoonan, line 1102: can we do the same for `/accessibility`?

marktnoonan (approved):

PR #5991, opened 2024-11-07
docs: add defaultBrowser config option

alexsch01:
For https://github.com/cypress-io/cypress/pull/30517


MikeMcC399 (commented):
Wouldn't it make more sense to put this in the https://docs.cypress.io/app/references/configuration#Browser section?

MikeMcC399 (commented):
What about https://docs.cypress.io/app/references/module-api#Options ?

Shouldn't there be an entry there as well?

- (I may have been misled due to the comment in https://github.com/cypress-io/cypress/pull/30517#pullrequestreview-2430146621)

alexsch01:
> Wouldn't it make more sense to put this in the https://docs.cypress.io/app/references/configuration#Browser section?

changed, I think that makes sense

MikeMcC399 (changes requested):
Linting is failing, so 

```shell
npm run lint:fix
```

needs to be run

docs/app/references/configuration.mdx
```diff
@@ -146,6 +146,7 @@ For more options regarding screenshots, view the
 
 | Option                  | Default | Description                                                                                                                                                                                                                                                                                                                                           |
 | ----------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `defaultBrowser`        | `null`  | The default browser to launch if the "--browser" command line option is not provided. This option only affects the first browser launched; changing this option after the browser is already launched will have no effect.                                                                                                                            |
```
mschile, line 149:
```suggestion
| `defaultBrowser`        | `null`  | The default browser to launch if the "--browser" command line option is not provided. This option only affects the first browser launch; changing this option after the browser is already launched will have no effect.                                                                                                                              |
```

mschile (approved):

PR #6024, opened 2024-12-02
Rework CYPRESS_RUN_BINARY instructions

MikeMcC399:
## Issue

[Advanced Installation > Run binary](https://docs.cypress.io/app/references/advanced-installation#Run-binary) includes instructions for using the environment variable `CYPRESS_RUN_BINARY` to override where the npm module finds the Cypress binary.

1. The instructions to run Cypress do not work as written. The package manager prefix `npx`, `yarn` or `pnpm` is missing and there is no hint that this would be needed.
2. There are no instructions about how to unzip the Cypress binary `cypress.zip`. Depending on how the extract or unzip operation is performed, there may be an extra directory level created. On Linux, for instance, this could mean that it is necessary to set the environment variable to `CYPRESS_RUN_BINARY=~/Downloads/cypress/Cypress/Cypress` instead of `CYPRESS_RUN_BINARY=~/Downloads/Cypress/Cypress`. This is quite confusing considering the multiple use of the names `cypress` and `Cypress` as directories and `Cypress` as an executable.
3. If the unzipped binary is located in a read-only location, then Cypress may fail to verify and therefore fail to run (see issue https://github.com/cypress-io/cypress/issues/30684). There is no mention of this issue or a possible workaround.
4. As written, the cautionary message, advising against exporting the `CYPRESS_RUN_BINARY` environment variable, does not make a lot of sense.

## Change

The instructions in [Advanced Installation > Run binary](https://docs.cypress.io/app/references/advanced-installation#Run-binary) for using the environment variable `CYPRESS_RUN_BINARY` are reworked.

### DIFF
--- a/docs/app/references/advanced-installation.mdx
+++ b/docs/app/references/advanced-installation.mdx
@@ -186,32 +186,56 @@ ensure this, consider exporting this environment variable. For example, in a
 Setting the environment variable `CYPRESS_RUN_BINARY` overrides where the npm
 module finds the Cypress binary.
 
-`CYPRESS_RUN_BINARY` should be a path to an already unzipped binary executable.
-The Cypress commands `open`, `run`, and `verify` will then launch the provided
-binary.
+`CYPRESS_RUN_BINARY` should be a path to an already unzipped Cypress binary executable.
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Preview

- https://github.com/MikeMcC399/cypress-documentation/blob/rework/run_binary/docs/app/references/advanced-installation.mdx#run-binary

PR #6027, opened 2024-12-04
doc: Cypress 13.16.1 changelog

cacieprins:


docs/app/references/changelog.mdx
```diff
@@ -2,6 +2,14 @@
 title: Changelog
 ---
 
+## 13.16.1
+
+_Released 12/04/2024_
+
+**Bugfixes:**
+
+- During recorded or parallel runs, execution will fail if Cypress is unable to confirm the creation of an instance instead of skipping the spec. Addresses [#30628](https://github.com/cypress-io/cypress/issues/30628).
```
mschile, line 11: Can you have a Cloud parallel run without recording?

```suggestion
- During recorded runs, execution will fail if Cypress is unable to confirm the creation of an instance instead of skipping the spec. Addresses [#30628](https://github.com/cypress-io/cypress/issues/30628).
```

jennifer-shehane (approved):

docs/app/references/changelog.mdx
cacieprins: Yep. Without a confirmed instance for a spec, the app will not run that spec. This was intended behavior prior to 13.16.1, evidenced by error messages in version control

PR #6029, opened 2024-12-05
[WIP] supplementalElements UI Coverage configuration

pstakoun:


emilyrohrbough:
I'd have to read it again to try to understand the intent, but the grouping behavior needs to be outlined. If only one element is found, is it an element or always a group?

PR #6031, opened 2024-12-09
Slack, Teams, and Data Extract API changes

danjohansenconsulting:
Corresponding with our November releases, we have added the following capabilities and need the docs to match:

* Enterprise reporting offers filtering by Commit Branch in the UI
* Data Extract API allow for filtering by commit branch, has some data now available near real-time, and added spec path and test names to some endpoints as requested by customers.
* Slack integration has added the ability to configure when a message is sent based on Run Groups
* Microsoft Teams integration has added the ability to configure when a message is sent based on Run Groups and Tags.  The "beta" status of Teams integration is being removed and it is now official.

Where applicable, screen captures shown in our Docs will be updated in the December release because of more UI changes coming in the next few weeks.


danjohansenconsulting:
Starting the draft period.  We are finalizing the filtering in a Cloud release tomorrow and once that launches we will finish out this documentation push.

jennifer-shehane (approved):
Looks good!

Issue #6054, opened 2024-12-20
Orphaned Test Replay limitations link


MikeMcC399:
## Issue

The target https://docs.cypress.io/faq/questions/cloud-faq#Is-everything-captured-and-replayed-in-Test-Replay was removed by PR https://github.com/cypress-io/cypress-documentation/pull/5985. It was last present in 3f2c22c.

It is used in:

- https://docs.cypress.io/accessibility/core-concepts/how-it-works#Powered-by-Test-Replay
  > Because Cypress Accessibility uses data captured through Cypress Test Replay, it is subject [Test Replay limitations](https://docs.cypress.io/faq/questions/cloud-faq#Is-everything-captured-and-replayed-in-Test-Replay).

- https://docs.cypress.io/ui-coverage/get-started/introduction#How-it-Works
  > Because Cypress Accessibility uses data captured through Cypress Test Replay, it is subject to any [limitations of Test Replay](https://docs.cypress.io/faq/questions/cloud-faq#Is-everything-captured-and-replayed-in-Test-Replay) data captured and browser support.

Clicking on the link leaves the reader stranded with no content discernible which matches the theme: "limitations of Test Replay".

It used to contain the following content:

---

**Is everything captured and replayed in Test Replay?**

Our aim is to create an impactful debugging experience that covers the most ground for users. We will continue to expand support. Currently Test Replay does not support the following:

#### DOM elements

- Video ([see issue](https://github.com/cypress-io/cypress/issues/29903))
- Audio ([see issue](https://github.com/cypress-io/cypress/issues/29905))
- Objects that do not have `type="image/svg+xml"` ([see issue](https://github.com/cypress-io/cypress/issues/29904))
- Shadow DOM with `slotAssignment="manual"` ([see issue](https://github.com/cypress-io/cypress/issues/29107))
- Shadow DOM canvas ([see issue](https://github.com/cypress-io/cypress/issues/30438))

#### Browsers

- WebKit browser (Safari) ([see issue](https://github.com/cypress-io/cypress/issues/28895))
- Firefox browser ([see issue](https://github.com/cypress-io/cypress/issues/28894))

#### Other

- Server sent events ([see issue](https://github.com/cypress-io/cypress/issues/29908))
- Web sockets ([see issue](https://github.com/cypress-io/cypress/issues/29907))
- localStorage & sessionStorage ([see issue](https://github.com/cypress-io/cypress/issues/29909))
[...]

PR #6055, opened 2024-12-20
Convert https://docs.cypress.io/ links to local

MikeMcC399:
## Issue

Source links stored in `*.mdx` files with the published domain name `https://docs.cypress.io/...` are not portable and may produce unexpected results when testing locally under `localhost`.

Secondly, Docusaurus does not report errors for links with external domains, meaning errors may go undetected. It only checks local links.

### Missing target bookmarks

The following links resolve to a page, however the corresponding bookmark is missing from the page and so, when the link is used, the cursor is placed at the top of the page, leaving the reader to figure out what was intended:

- https://docs.cypress.io/faq/questions/cloud-faq#Is-everything-captured-and-replayed-in-Test-Replay resolves to https://docs.cypress.io/cloud/faq#Is-everything-captured-and-replayed-in-Test-Replay. The bookmark `Is-everything-captured-and-replayed-in-Test-Replay` is missing. This is logged as a separate issue https://github.com/cypress-io/cypress-documentation/issues/6054.
- https://docs.cypress.io/app/core-concepts/open-mode#Clicking-on-Commands. The bookmark `Clicking-on-Commands` is missing.

## Change

- Remove the external part of URLs in `*.mdx` file links containing `https://docs.cypress.io/...`

- Change https://docs.cypress.io/app/core-concepts/open-mode#Clicking-on-Commands to the internal link `/app/core-concepts/open-mode#Command-Log`

### DIFF
--- a/docs/api/cypress-api/require.mdx
+++ b/docs/api/cypress-api/require.mdx
@@ -15,7 +15,7 @@ in browser-targeted code.
 
 :::caution
 
-Using `Cypress.require()` within the [`cy.origin()`](https://docs.cypress.io/api/commands/origin)
+Using `Cypress.require()` within the [`cy.origin()`](/api/commands/origin)
 callback requires enabling the
 [`experimentalOriginDependencies`](/app/references/experiments#End-to-End-Testing) option in the Cypress configuration.
 
--- a/docs/app/continuous-integration/overview.mdx
+++ b/docs/app/continuous-integration/overview.mdx
@@ -235,7 +235,7 @@ Cypress Docker images are available from:
 
 - [cypress/base](https://github.com/cypress-io/cypress-docker-images/tree/master/base) is the entry-level Cypress Docker image, allowing you to test in the Electron browser, built in to Cypress.
   It contains a complete Linux (Debian) operating system, together with the
-  [prerequisite operating system packages](https://docs.cypress.io/app/get-started/install-cypress#UbuntuDebian) for Cypress,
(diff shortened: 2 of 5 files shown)


jennifer-shehane (approved):
Yah I noticed some of these slipping into a recent PR. We definitely want them local so the checks run. Might be good to look for on.cypress.io links also.

MikeMcC399:
@jennifer-shehane 

> Yah I noticed some of these slipping into a recent PR. We definitely want them local so the checks run. Might be good to look for on.cypress.io links also.

It's easy to forget to strip off the domain when copying links. I was guilty of adding one of these links for Docker, which is now corrected.

I did look at `on.cypress.io` and there is a lot of use of this, so it looked to me like it was strategy and not a mistake. I can take a look at converting these links in the new year, if that is what you would like to have done.

jennifer-shehane:
@MikeMcC399 If the result on the on.cypress.io link results in a link back to the docs itself, it should be a relative link. There may be some outside links that on.cypress.io reroutes to (like to cloud.cypress.io), those should be left.

MikeMcC399:
@jennifer-shehane 

> If the result on the on.cypress.io link results in a link back to the docs itself, it should be a relative link. There may be some outside links that on.cypress.io reroutes to (like to cloud.cypress.io), those should be left.

Understood. That is what I was assuming.

MikeMcC399:
- on.cypress.io links addressed in PR #6061

Issue #6092, opened 2025-01-28
Add explicit package manager support list and installation instructions


MikeMcC399:
## What would you like?

1. Add a section for `Package Manager` in the [Get Started > Install Cypress > System requirements](https://docs.cypress.io/app/get-started/install-cypress#System-requirements) section
2. Add the following list of supported package managers:
   - [npm](https://docs.npmjs.com/) `8.6.0` and above (corresponds to npm version packaged with Node.js `18.0.0`)
   - [Yarn Classic](https://classic.yarnpkg.com/lang/en/) `1.22.22` and above
   - [Yarn Modern](https://yarnpkg.com/) `4.x` and above
        for Component testing, with [`nodeLinker: "node-modules"`](https://yarnpkg.com/configuration/yarnrc#nodeLinker) configuration only
   - [pnpm](https://pnpm.io/) `8.x` and above
3. Move any generic installation instructions, which are not Cypress specific, down to this section.
4. Separate out Cypress-specific installation instructions and keep these in the [Get Started > Install Cypress](https://docs.cypress.io/app/get-started/install-cypress#Install) section.

## Why is this needed?

1. Although npm, Yarn and pnpm are listed as supported, there is no information about which versions are supported.
2. Descriptions in [Get Started > Install Cypress](https://docs.cypress.io/app/get-started/install-cypress#Install) mix up the reference to specific package managers. Generic and Cypress-specific instructions are put together instead of being clearly separated.
3. Yarn Classic and Yarn Modern are different products with a combination of similarities and major incompatibilities, so it is not clear, when referring to Yarn, whether Yarn Classic or Yarn Modern or both are meant.
4. Instructions for how to install Yarn are missing. If these instructions are added, they should take account of the fact that [Yarn 1 Classic](https://classic.yarnpkg.com/lang/en/docs/install/) and [Yarn Modern](https://yarnpkg.com/getting-started/install) are to be installed differently from each other.
5. Instructions for how to install pnpm using npm are incorrect if Corepack is enabled. Refer to [pnpm Installation](https://pnpm.io/installation) for full instructions.

## Other

Related to:

- https://github.com/cypress-io/cypress-documentation/issues/5394

MikeMcC399:
@jennifer-shehane 

I plan to submit a PR to implement the above suggestions.

PR #6093, opened 2025-01-28
Add package manager list and installation instructions

MikeMcC399:
- closes https://github.com/cypress-io/cypress-documentation/issues/6092

## Issue

[Get Started > Install Cypress](https://docs.cypress.io/app/get-started/install-cypress) information for using package managers: npm, Yarn and pnpm is scattered and not at the same level of detail for each package manager. Developments in package management usage, such as the introduction of Corepack, are not covered. When Yarn is mentioned there is no distinction made between Yarn 1 Classic and Yarn Modern. For the latter there is no indication about which `nodeLinker` modes are supported and which do not work.

## Change

- Generalize package manager installation and refer back to each package manager's own documentation.
- For Yarn Modern prefer `node-modules` and exclude [Yarn Plug'n'Play](https://yarnpkg.com/features/pnp) for Component Testing usage.

## Corepack comment

- Since [Corepack](https://nodejs.org/docs/latest/api/corepack.html) is currently considered as experimental by the Node.js documentation site, do not refer to Corepack explicitly.
  - npm does not use Corepack.
  - Yarn Classic was released before Corepack and does not document its use at all.
  - Yarn Modern has its own separate [Corepack](https://yarnpkg.com/corepack) page.
  - pnpm has a [Corepack section](https://pnpm.io/installation#using-corepack) as part of the [Installation page](https://pnpm.io/installation)

## pnpm 10.x comment

- Delay adding any specific information for pnpm 10.x until the documentation and releases have stabilized.

### DIFF
--- a/docs/app/get-started/install-cypress.mdx
+++ b/docs/app/get-started/install-cypress.mdx
@@ -1,5 +1,5 @@
 ---
-title: 'Install using npm, yarn, or pnpm | Cypress Documentation'
+title: 'Install using npm, Yarn, or pnpm | Cypress Documentation'
 description: 'A step-by-step guide on how to install Cypress. Learn the requirements, installation process, and get started with Cypress for end-to-end and component testing'
 sidebar_label: Install Cypress
 sidebar_position: 30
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Preview

https://github.com/MikeMcC399/cypress-documentation/blob/add/package-manager-support-installation/docs/app/get-started/install-cypress.mdx

jennifer-shehane (approved):

Issue #5946, opened 2024-10-10
Remove deprecated plugins


MikeMcC399:
### Subject

[Plugins](https://docs.cypress.io/plugins)

### Description

Remove deprecated plugins from [Plugins](https://docs.cypress.io/plugins)

- which are based on repos which have been archived or which are otherwise declared as unsupported
- which support only legacy versions of Cypress `v9` or lower
- which have not been updated since the release of Cypress `v10` in June 2022

#### Preprocessors

| Plugin                           | Status                                 |
| -------------------------------- | -------------------------------------- |
| cytorus                          | Archived Sep 3, 2024                   |
| cypress-app-watcher-preprocessor | Unmaintained. Last update Feb 2019     |
| cypress-eslint-preprocessor      | Unmaintained. Last update Nov 2018     |
| cy-rollup                        | Unmaintained. Cypress v8               |
| cypress-esbuild-preprocessor     | Unmaintained. Last update Oct 22, 2020 |

#### Development Tools

| Plugin                           | Status                   |
| -------------------------------- | ------------------------ |
| @cypress/instrument-cra          | Archived Jan 31,2023     |
| cypress-dark                     | Unmaintained. Cypress v8 |
| cypress-protobuf                 | Unmaintained. Cypress v4 |
| knapsack-pro-cypress             | Archived Jun 13, 2023    |
| cypress-autostub                 | Unmaintained. Cypress v4 |
| cypress-select-tests             | Archived May 28, 2021    |
| cypress-autorecord               | Unmaintained. Cypress v6 |
| @cypress/fiddle                  | Archived Feb 28, 2023    |
| cypress-extends                  | Unmaintained. Cypress v9 |
| @cypress/browserify-preprocessor | Archived Jan 31, 2023    |

#### Custom Commands

| Plugin                         | Status                   |
| ------------------------------ | ------------------------ |
| cypress-unfetch                | Unmaintained. Cypress v3 |
| cypress-redux                  | Unmaintained. Cypress v3 |
| cypress-pipe                   | Unmaintained. Cypress v4 |
| cypress-plugin-stripe-elements | Unmaintained. Cypress v8 |
| @cypress/skip-test             | Archived Jan 31, 2023    |
| cypress-websocket-testing      | Unmaintained. Cypress v3 |
| cypress-iframe                 | Unmaintained. Cypress v3 |
| cypress-rest-graphql           | Unmaintained. Cypress v7 |
[...]

PR #6102, opened 2025-02-14
Add cypress-laravel plugin

NoelDeMartin:
I had already registered this plugin in #2250, but it was removed without my knowledge in #5946.

I also realized it had been marked as "deprecated" before, also without my knowledge. I suppose it was marked as such because it had Cypress 3 as a dependency, but that's something [I fixed 3 months ago](https://github.com/NoelDeMartin/cypress-laravel/releases/tag/v0.3.0). Would have fixed it sooner if anyone bothered to open an issue about it in my repo :).

### DIFF
--- a/src/data/plugins.json
+++ b/src/data/plugins.json
@@ -52,7 +52,14 @@
           "link": "https://github.com/mammadataei/cypress-vite",
           "keywords": ["vite"],
           "badge": "community"
-        }
+        },
+        {
+          "name": "cypress-laravel",
+          "description": "Add commands and hooks to test Laravel applications.",
+          "link": "https://github.com/noeldemartin/cypress-laravel",
+          "keywords": ["php", "laravel"],
+          "badge": "community"
+        },
       ]
     },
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
@NoelDeMartin Yes, that's likely why. We can add it back.

src/data/plugins.json
```diff
@@ -52,7 +52,14 @@
...
           "badge": "community"
-        }
+        },
+        {
+          "name": "cypress-laravel",
+          "description": "Add commands and hooks to test Laravel applications.",
+          "link": "https://github.com/noeldemartin/cypress-laravel",
+          "keywords": ["php", "laravel"],
+          "badge": "community"
+        },
```
MikeMcC399, line 62:
```suggestion
        }
```

MikeMcC399: otherwise there is a build failure

jennifer-shehane (approved):

PR #6105, opened 2025-02-18
Add Plugins: cypress-plugin-grep-boxes & cypress-cli-select

dingraham:
Adding two new, fantastic plugins that support expanded grep/test select functionality. 

- [cypress-plugin-grep-boxes](https://github.com/dennisbergevin/cypress-plugin-grep-boxes)
- [cypress-cli-select-plugins](https://github.com/dennisbergevin/cypress-cli-select)

@dennisbergevin

### DIFF
--- a/src/data/plugins.json
+++ b/src/data/plugins.json
@@ -147,11 +147,25 @@
           "keywords": ["auditory", "ui", "results", "duration"],
           "badge": "community"
         },
-        {
+      {
           "name": "cypress-plugin-last-failed",
-          "description": "A companion Cypress plugin for `cy-grep` that re-runs the last failed test(s).",
+          "description": "A companion Cypress plugin for `cy-grep` that re-runs the last failed test(s).",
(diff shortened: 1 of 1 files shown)


MikeMcC399:
@dingraham / @dennisbergevin 

It would be great if the new plugins are updated from Cypress `13.x` to use and test under the current major Cypress `14.x` version!

It's part of the requirement list on https://github.com/cypress-io/cypress-documentation/blob/main/CONTRIBUTING.md#adding-plugins

dennisbergevin:
> @dingraham / @dennisbergevin
> 
> It would be great if the new plugins are updated from Cypress `13.x` to use and test under the current major Cypress `14.x` version!
> 
> It's part of the requirement list on https://github.com/cypress-io/cypress-documentation/blob/main/CONTRIBUTING.md#adding-plugins

Hi @MikeMcC399, thanks for the reminder, I've gone ahead and updated both plugins so that Cypress is up to date with the latest version.

PR #6133, opened 2025-03-26
feat: (ui-coverage) add examples for protocol-agnostic views and viewFilters

astone123:
Update UI coverage docs to show examples of specifying views and view patterns without a specific protocol.

See related PR https://github.com/cypress-io/cypress-services/pull/10437

### DIFF
--- a/docs/partials/_viewfilters.mdx
+++ b/docs/partials/_viewfilters.mdx
@@ -137,3 +137,32 @@ https://cypress.io/error/500
 ```
 https://cypress.io/home
 ```
+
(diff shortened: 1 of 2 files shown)


marktnoonan (approved):
Looks good. I might like to add some more explainer text, but I think that would make sense across this whole page and should come in a new PR, this correctly documents the current behavior.

marktnoonan (changes requested):
(Blocking for now as I noticed the related PR for the functionality hasn't merged yet.)

PR #6138, opened 2025-03-31
chore: Node.js 18/23 removal

jennifer-shehane:
Goes with https://github.com/cypress-io/cypress/pull/31409


MikeMcC399:
If Node.js 18.x will no longer be supported. Node.js `20.x` LTS becomes the lowest supported version.

The lowest `20.x` LTS version is [Node.js v20.9.0](https://nodejs.org/en/blog/release/v20.9.0) which ships with npm `10.1.0`. So this means that the npm version should be bumped from `8.6.0` to `10.1.0` to match.

https://github.com/cypress-io/cypress-documentation/blob/0e8c8aefb9628c4dc76f204cc55bef4ee59e34b6/docs/app/get-started/install-cypress.mdx?plain=1#L144

([Node.js v20.0.0](https://nodejs.org/en/blog/release/v20.0.0) shipped with npm `9.6.4` but I don't think this needs to be considered because Node.js doesn't support older non-LTS releases.)

jennifer-shehane:
@MikeMcC399 Thanks for the detailed review, as always! I updated the lowest npm version supported as well.

MikeMcC399:
@jennifer-shehane 

> Thanks for the detailed review, as always! I updated the lowest npm version supported as well.

Thanks also for making the tweak!

PR #6145, opened 2025-04-08
Correct parameter usage for click() in shadow.mdx

MatthewSbar:
The documentation for `Shadow` suggests that it's valid to pass `{position: 'top'}` to `cy.click()`, however `cy.click()`'s documentation suggests the `position` parameter should be passed as a standalone string outside of the `options` object, and the `options` object does not include a field for `position`.

- Update the suggestion to pass `'top'` to `cy.click()` using the API defined in `cy.click()`'s documentation.

See
https://docs.cypress.io/api/commands/click


jennifer-shehane (approved):
Indeed, this is correct. Thank you!

Issue #6147, opened 2025-04-11
pnpm side effects cache not compatible with Cypress postinstall script


MikeMcC399:
## Issue

pnpm implements a [side-effects-cache](https://pnpm.io/npmrc#side-effects-cache) that uses and caches the results of (pre/post)-install hooks. This does not work if the effects of running the hook are outside the package's module. This is the case for the Cypress `postinstall` hook which installs the Cypress binary in a separate cache location (see [Binary cache](https://docs.cypress.io/app/references/advanced-installation#Binary-cache)). pnpm believes that it is unnecessary to run the Cypress `postinstall` hook and skips it if the Cypress npm module is already present in the pnpm store cache.

This is not documented on https://docs.cypress.io/ and it is only briefly referred to on https://pnpm.io/settings#sideeffectscache.

This can cause Cypress to report the error:

> The cypress npm package is installed, but the Cypress binary is missing.

## Suggestion

Add configuration and troubleshooting information for the pnpm side effects cache.

Ideally pnpm would convert to consistently support using `pnpm-workspace.yaml`, so that Cypress could recommend using:

```yaml
pnpm config set side-effects-cache false --location project
pnpm --allow-build cypress add --save-dev cypress
```

There are however multiple related issues (see below) concerning configuration and pnpm `10.x` is still evolving in this respect.

I will hold back before attempting to submit any documentation change PRs for this.

## Related issues

- https://github.com/pnpm/pnpm/issues/9123
- https://github.com/pnpm/pnpm/issues/9394
- https://github.com/pnpm/pnpm/issues/9396
- https://github.com/pnpm/pnpm/issues/9399

MikeMcC399:
[pnpm@10.8.1](https://github.com/pnpm/pnpm/releases/tag/v10.8.1), release Apr 14, 2025 has resolved several configuration issues, so this Cypress documentation enhancement can now be moved forward.

MikeMcC399:
- PR #6149 is submitted to add the side effects cache instructions to the installation configuration
- A separate PR will follow to add related troubleshooting steps

MikeMcC399:
The main part of this issue is resolved through PR https://github.com/cypress-io/cypress-documentation/pull/6149

https://docs.cypress.io/app/get-started/install-cypress#pnpm-Configuration

![Image](https://github.com/user-attachments/assets/fb8a1248-09a7-4467-8719-cded9db86e5f)

Troubleshooting should be handled separately if necessary.

PR #6149, opened 2025-04-14
extend pnpm config instructions

MikeMcC399:
- partially resolves issue https://github.com/cypress-io/cypress-documentation/issues/6147

## Issue

pnpm implements a [side effects cache](https://pnpm.io/npmrc#side-effects-cache) that uses and caches the results of (pre/post)-install hooks. This does not work if the effects of running the hook are outside the package's module. This is the case for the Cypress `postinstall` hook which installs the Cypress binary in a separate cache location (see [Binary cache](https://docs.cypress.io/app/references/advanced-installation#Binary-cache)). pnpm believes that it is unnecessary to run the Cypress `postinstall` hook and skips it if the Cypress npm module is already present in the pnpm store cache.

This is not documented on https://docs.cypress.io/ and it is only briefly referred to on https://pnpm.io/settings#sideeffectscache.

Cypress will not run if the Cypress binary is not installed.

## Change

Recommend the configuration related commands:

```yaml
pnpm config set side-effects-cache false --location project
pnpm --allow-build cypress add --save-dev cypress
```

and rework the section https://docs.cypress.io/app/get-started/install-cypress#pnpm-Configuration.

### DIFF
--- a/docs/app/get-started/install-cypress.mdx
+++ b/docs/app/get-started/install-cypress.mdx
@@ -157,10 +157,23 @@ is preferred. Cypress [Component Testing](/app/core-concepts/testing-types#What-
 
 #### pnpm Configuration
 
+The following configuration options enable Cypress to execute its `postinstall` script so it can install the Cypress binary into the [binary cache](/app/references/advanced-installation#Binary-cache).
+If these configuration options are not set, then Cypress may skip the `postinstall` script execution and Cypress will not run.
+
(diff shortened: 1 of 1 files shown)


MikeMcC399:
Preview

https://github.com/MikeMcC399/cypress-documentation/blob/pnpm-configuration/docs/app/get-started/install-cypress.mdx#pnpm-configuration

jennifer-shehane (approved):

jennifer-shehane (approved):

PR #6191, opened 2025-05-21
Fix changelog 14.4.0

MikeMcC399:
Changes [Changelog](https://docs.cypress.io/app/references/changelog) for Cypress `14.4.0`

1. Change header `14.4.0` to second level, so that it appears in the right-hand navigation and is available as a bookmark
2. Remove one space after `(`
3. Change https://docs.cypress.io/api/node-events/browser-launch-api#Add-browser-extensions to relative URL `/api/node-events/browser-launch-api#Add-browser-extensions`

### DIFF
--- a/docs/app/references/changelog.mdx
+++ b/docs/app/references/changelog.mdx
@@ -8,13 +8,13 @@ sidebar_label: Changelog
 
 # Changelog
 
-# 14.4.0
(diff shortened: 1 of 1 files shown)


jennifer-shehane (approved):
thanks!

PR #6236, opened 2025-07-31
Update UI Coverage documenation

marktnoonan:
* Update element identification and interaction commands docs to reference addition ways of managing what is considered interactive, and what commands count towards coverage.
* Fix incorrect docs for significant attributes format.


docs/partials/_significantattributes.mdx
```diff
@@ -112,12 +58,7 @@ button: type="submit", disabled="disabled"
 
 ```json
 {
-  "significantAttributes": [
-    {
-      "selector": "*",
-      "attributes": ["aria-*"]
-    }
-  ]
+  "significantAttributes": ["aria-*"]
```
mike-plummer: IIRC this doesn't support wildcarding

docs/ui-coverage/core-concepts/interactivity.mdx
```diff
@@ -19,6 +19,16 @@ UI Coverage determines interactivity based on a combination of HTML semantics, [
...
 
+### Customizing interactive elements
+
+Custom interactive elements that do not meet the criteria above can also be declared with a `data-cy-ui-interactive="include"` attribute, used as follows:
+
+```html
+<div data-cy-ui-interactive="include">Custom interactive widget</div>
+```
+
+It most cases we recommend **not** using this override, and instead updating the HTML to be something the browser would already consider to be interactive. This will likely produce better overall behavior, including for accessibility purposes. `data-cy-ui-interactive` is a fallback for situations where that may not be possible.
```
mike-plummer, line 30: Personally I would word this a bit more strongly - something like "use of this attribute is a strong signal there is an issue with the HTML structure of your application which likely has real-world impacts for users, particularly those using screen readers or with disabilities. If possible we recommend updating your HTML to better convey the interactive nature of this element; this would improve the functionality of your application for these users as well as automatically detect the element in the report. As a fallback for scenarios where this isn't possible......."

marktnoonan: I'm not so sure about this. Since we can now support requiring custom commands and the "assert" command - you might have a totally valid reason to want to change the definition of what we consider interactive without it implying the elements themselves are broken.

Still might need another pass at this though - the framing of everything in terms of "interactions" might not be the most accurate, long term. Will think on it a bit.

docs/partials/_significantattributes.mdx
jennifer-shehane: I'm not totally sure if this API exactly matches what unique-selector accepts, but if it does, it would be something like`['attribute:aria-label', 'attribute:role', 'id', 'class']`

mike-plummer: @jennifer-shehane That's abstracted away by UICov logic here https://github.com/cypress-io/cypress-services/blob/develop/discovery/packages/common/src/elementProcessingUtils.ts#L72

marktnoonan: yeah I didn't verify on the wildcarding, I'll remove this from the PR

PR #6242, opened 2025-08-10
docs: Add best practices for element interaction patterns and command chaining

### LINKED ISSUE #6189: Best practice - then() usage vs chained commands
### Description

Hi,

I wanted to clarifiy some best practices around then() function and 'classical' chaining.  
What is the best practice for example when interacting with an input where we want to:
- focus
- remove current text content
- write a new value
- and blur

I ask that becasue, it is a common practice and i read several ways to do it but some of them seems unstable:

1. First option: The most classic is to make the commands successively as some of them cannot be chained (type(), focus(), clear()...) So it will be:

`
cy.get('form #stripe-payment-element').focus();
cy.get('form #stripe-payment-element').clear();
cy.get('form #stripe-payment-element').type('new value');
cy.get('form #stripe-payment-element').blur();
`

2. Second option: use the then() function to wrap all actions. So it will be:

`
cy.get('form #stripe-payment-element').then(($input) => {
        cy.wrap($input).focus();
        cy.wrap($input).clear();
        cy.wrap($input).type(value);
        cy.wrap($input).blur();
  });
`

This last case seems to be, as what i read, the most appropriate as:
- it reduce risks of "detached DOM errors"
- the input if found once  and not re-searched 3 times

Is it the best approach ? But also the most performant too ? 

Documentation is really well done, but with so many posibilities, sometimes it is hard to choose the best approach :)

Hope my question is clear!

ziad404:
closes #6189


docs/app/core-concepts/best-practices.mdx
```diff
@@ -951,6 +951,184 @@ cy.wait('@getUsers') // <--- wait explicitly for this route to finish
...
+    cy.get('[data-cy="email-input"]')
+      .should('be.visible')
+      .focus()
+      .clear()
+      .type('user@example.com')
+      .should('have.value', 'user@example.com')
+      .blur()
+  })
+})
+```
```
MikeMcC399: This proposed addition to the documentation contradicts other information about chaining.

The example code does not follow the rules set out in [Actions should be at the end of chains, not the middle](https://docs.cypress.io/app/core-concepts/retry-ability#Actions-should-be-at-the-end-of-chains-not-the-middle)

and the "unsafe to chain" comments in:

- [.focus](https://docs.cypress.io/api/commands/focus)
- [.clear](https://docs.cypress.io/api/commands/clear)
- [.type](https://docs.cypress.io/api/commands/type)

MikeMcC399 (changes requested):
@ziad404 

I added a comment about chaining, although I don't have the background to comment on your complete documentation section. The chaining part would need to be revised in order to be correct.

MikeMcC399:
@ziad404 

Are you intending to improve this PR or have you dropped it now?

docs/app/core-concepts/best-practices.mdx
```diff
@@ -951,6 +951,185 @@ cy.wait('@getUsers') // <--- wait explicitly for this route to finish
...
+
+#### Option 1: Command Chaining (Recommended)
+
+```js
+// RECOMMENDED: Use separate cy.get() or cy.wrap() for each action command
+cy.get('[data-cy="input-field"]').focus()
+cy.get('[data-cy="input-field"]').clear()
+cy.get('[data-cy="input-field"]').type('new value')
+cy.get('[data-cy="input-field"]').blur()
+```
```
MikeMcC399: This no longer makes sense, since the title is Command Chaining and the examples do not use chaining

MikeMcC399 (changes requested):
Apart from the comment about chaining, your PR fails to build, so you would need to correct both issues.

MikeMcC399 (commented):
@ziad404 

You have addressed my previous comments. Now your article section needs to be reviewed by a subject matter expert, so hopefully the Cypress.io team will respond to you.

docs/app/core-concepts/best-practices.mdx
```diff
@@ -951,6 +951,193 @@ cy.wait('@getUsers') // <--- wait explicitly for this route to finish
 cy.get('table tr').should('have.length', 2)
 ```
 
+## <Icon name="angle-right" /> Element Interaction Patterns and Command Chaining
+
+:::danger
+
+<Icon name="exclamation-triangle" color="red" /> **Anti-Pattern:** Using `then()`
+unnecessarily or re-querying the same element multiple times when chaining would
+work.
```
cacieprins, line 960:
```suggestion
<Icon name="exclamation-triangle" color="red" /> **Anti-Pattern:** Using `then()`
unnecessarily or re-querying the same element multiple times when chaining is
inefficient and reduces test clarity.
```

docs/app/core-concepts/best-practices.mdx
```diff
@@ -951,6 +951,193 @@ cy.wait('@getUsers') // <--- wait explicitly for this route to finish
...
+  if (initialValue !== '') {
+    cy.wrap($input).clear().type('new value')
+  }
+})
+```
+
+2. **Perform conditional logic based on element state:**
+
+```js
+cy.get('[data-cy="toggle"]').then(($toggle) => {
```
cacieprins, line 1046: Unfortunately, conditional actions and/or assertions introduce flake, reduce clarity, and make debugging more difficult. They're widely considered an anti-pattern in tests, and serve as a "code smell," hinting at opportunities for quality improvement:

- If they are necessary, you may be breaking test isolation (itself an anti-pattern), where the application state is carried over from a previous test.
- You probably need more granular tests

There are some edge cases where you may want to do this, like in before/beforeEach, but there are probably better ways to construct the state you wish to act upon in the test.

We should not call them out in our docs as a recommended pattern, IMHO.

cacieprins (changes requested):

PR #6250, opened 2025-08-16
Remove reference to `experimentalInteractiveRunEvents` flag

jkjustjoshing:
It looks like this flag doesn't exist at the link destination. I'm not 100% sure it's gone, but the docs link is dead.

I figured I'd submit this PR even though I'm not sure - if I'm correct, this can be approved and merged. If not, consider this a bug report and update the PR to reflect the true state of the flags.

### DIFF
--- a/docs/api/node-events/after-spec-api.mdx
+++ b/docs/api/node-events/after-spec-api.mdx
@@ -15,12 +15,6 @@ The `after:spec` event fires after a spec file is run. When running cypress via
 
 <WarningSetupNodeEvents />
 
-:::caution
-
-⚠️ When running via `cypress open`, the `after:spec` event only fires if the
-[experimentalInteractiveRunEvents flag](/app/references/configuration#Experiments)
-is enabled.
-
 :::
 
 :::cypress-config-plugin-example


MikeMcC399:
@jkjustjoshing 

This looks like a misinterpretation

https://docs.cypress.io/api/node-events/after-spec-api#Syntax

image

https://docs.cypress.io/app/references/configuration#Experiments

image

https://docs.cypress.io/app/references/experiments

image

MikeMcC399:
I suggest to close this PR as the experimental option `experimentalInteractiveRunEvents` is valid and should not be removed from the documentation.

jkjustjoshing:
Ah I understand. I saw `experimentalRunAllSpecs` on https://docs.cypress.io/app/references/configuration, which made me think that that page had experimental and non-experimental flags on it. 

And honestly I didn't notice the "Experimental" section. Can I recommend a `:target` styling to highlight the linked-to section while navigating?

jennifer-shehane (changes requested):
This does still exist. The link should be: /app/references/experiments

docs/api/node-events/after-spec-api.mdx
```diff
@@ -15,12 +15,6 @@ The `after:spec` event fires after a spec file is run. When running cypress via
 
 <WarningSetupNodeEvents />
 
-:::caution
-
-⚠️ When running via `cypress open`, the `after:spec` event only fires if the
-[experimentalInteractiveRunEvents flag](/app/references/configuration#Experiments)
```
emilyrohrbough, line 21: @jkjustjoshing thanks for flagging this! the correct path is `/app/references/experiments` for https://docs.cypress.io/app/references/experiments

PR #6286, opened 2025-10-01
chore: run prettier --fix on staged .md files

cacieprins:
This will fix minor errors, like incorrect offsets in tables, before they hit the PR stage.


jennifer-shehane (approved):
Thank you! 😭

PR #6287, opened 2025-10-04
fix: MASH to SMASH THAT START BUTTON

Uanela:
I natively speak portuguese but also fluent in English, I saw "MASH THAT START BUTTON" and became like what is now? is it some word I haven't learned yet...

Search a little bit and became like naaah it was a mistake, it must be "SMASH".


MikeMcC399:
@Uanela 

If you want to continue to suggest that the word is changed, then you would need to sign the CLA.

If you want to withdraw your suggestion, then in that case you should close your PR.

PR #6299, opened 2025-10-15
App quality content updates

marktnoonan:
Wrapping a couple of content changes together here in the App Quality area:

* Release of `data-a11y-ignore` attribute
* Publication of required CI environment variables for each supported provider for the Results API. This will help customers with getting started, debugging their code, and passing the right data through in more complicated setups that don't work out of the box.


docs/partials/_results-api-env-vars.mdx
```diff
@@ -0,0 +1,163 @@
+The `@cypress/extract-cloud-results` helper cross-references some environment variables from where it is executed with ones that were present when a Cypress Cloud run was recorded. This allows for automatically detecting the correct Cloud run when the Results API is invoked from the same CI context as a given run (as is the case in the above examples).
```
jennifer-shehane, line 1: This doc has a pretty big potential for becoming stale. It'd be great if we could somehow reference code to prevent this.

marktnoonan: Thanks for the review! Yeah I agree this is tricky. So far these haven't changed in the last year+. I was thinking we could link to this as the source of truth in the cloud internal readme for this package so that at least there is a cross reference if we ever do update these.

PR #6337, opened 2025-12-10
Experimental fast visibility

cacieprins:
Expands the visibility section in "interacting with elements" to include the "fast" visibility algorithm heuristics/semantics. Cross-references with links, and links to the main github issue tracking this experiment.

PR #6345, opened 2026-01-02
Adjust experimentalFastVisibility documentation

wlsf82:
Removed unnecessary code block markers from the documentation that make a sample test file look like the config file.

Cc. @jennifer-shehane

### DIFF
--- a/docs/app/references/changelog.mdx
+++ b/docs/app/references/changelog.mdx
@@ -8,6 +8,19 @@ sidebar_label: Changelog
 
 # Changelog
 
+## 15.8.2
+
+_Released 01/06/2026_
+
+**Bugfixes:**
+
+- Fixed an issue where the ffprobe path was not properly configured for video operations. The path is now set at module load time, ensuring it's available for all video operations. Upgraded `@ffprobe-installer/ffprobe` from `1.1.0` to `2.1.2` to support darwin-arm64 (Apple Silicon). Addressed in [#33136](https://github.com/cypress-io/cypress/pull/33136).
+- Fixed an issue where `test:after:run` and `test:after:run:async` events were not firing after both a `before all` and `after all` hook failed. Addressed in [#33172](https://github.com/cypress-io/cypress/pull/33172).
+
+**Dependency Updates:**
+
+- Upgraded `@cypress/request` to `3.0.10`. This removes the [CVE-2025-15284](https://security.snyk.io/vuln/SNYK-JS-QS-14724253) vulnerability being reported in security scans. Addressed in [#33188](https://github.com/cypress-io/cypress/pull/33188).
+
 ## 15.8.1
 
 _Released 12/18/2025_


docs/app/references/experiments.mdx
```diff
@@ -168,8 +168,6 @@ Enable fast visibility by setting `experimentalFastVisibility` to `true` in your
```
MikeMcC399, line 168:
```suggestion

:::

```

MikeMcC399: The config block above is otherwise not properly terminated.

```text
[WARNING] Docusaurus found 1 unused Markdown directives in file "docs/app/references/experiments.mdx"
- :::cypress-config-example (161:1)
Your content might render in an unexpected way. Visit https://github.com/facebook/docusaurus/pull/9394 to find out why and how to fix it.
```

wlsf82: The point of the PR is that a sample test file looks like a config file, but it shouldn't.

MikeMcC399: @wlsf82 

> The point of the PR is that a sample test file looks like a config file, but it shouldn't.

I understand, however there is a related error in the same file and in the same section that was wrongly added by the same PR (https://github.com/cypress-io/cypress-documentation/pull/6337), so I'm suggesting to correct both errors in your PR here because they are related.

wlsf82: I see, thanks for clarifying.

MikeMcC399:
@wlsf82 

It's looking right now and with no build errors 👍🏻 

image

MikeMcC399 (approved):
LGTM!

You could optionally add a blank line after line 169. That is cosmetic only - since it lints and works already without it. Somehow GitHub lost the blank line when you accepted my suggestion.

docs/app/references/experiments.mdx
```diff
@@ -166,10 +166,9 @@ Enable fast visibility by setting `experimentalFastVisibility` to `true` in your
 }
 ```
 
+:::
```
wlsf82, line 169:
```suggestion
:::

```

MikeMcC399:
CI is failing due to known issue https://github.com/cypress-io/circleci-orb/issues/538 which occurs sporadically and can be ignored.

CircleCI has added a fix to their image build process, however it needs Node.js to cut a new release first (expected next week), and then the Cypress Orb can get updated to a default higher Node.js version as well. This will probably take a couple of weeks to run through the waterfall of updates.

wlsf82:
> CI is failing due to known issue [cypress-io/circleci-orb#538](https://github.com/cypress-io/circleci-orb/issues/538) which occurs sporadically and can be ignored.
> 
> CircleCI has added a fix to their image build process, however it needs Node.js to cut a new release first (expected next week), and then the Cypress Orb can get updated to a default higher Node.js version as well. This will probably take a couple of weeks to run through the waterfall of updates.

@MikeMcC399, could you re-run it to unblock the PR?

Hardanish-Singh (approved):

PR #6351, opened 2026-01-16
docs: Studio updates based on feedback

jennifer-shehane:


docs/app/guides/cypress-studio.mdx
```diff
@@ -84,19 +90,19 @@ Studio will automatically create a new test definition for you.
...
 
-You can extend and update existing tests using Cypress Studio.
+To extend an existing test:
 
 1. Run the spec in Cypress
 2. Hover over the test in the Command Log
 3. Click **Edit in Studio**
 
-Studio runs your test up to the last command and lets you keep building from there.
+Studio loads your test and runs it up to the last command, ready for you to either edit the code directly or record new actions and assertions from that point.
```
marktnoonan, line 105: Would we consider documenting a pattern to comment out the rest of your test if you want to start recording from a specific point in the timeline? This would work fine when making assertions, but not work well if you record actions that would invalidate later steps in the test. You could do the commenting/uncommenting in the editor and save the test.

marktnoonan (approved):

docs/app/guides/cypress-studio.mdx
jennifer-shehane: Yeah, I was shying away from taking a stance on the approach to editing the middle of a test. It may be warranted to have an actual recommendation, I'm not sure what the best recommendation is atm.

PR #6361, opened 2026-01-30
improve docs for run complete delay and run time settings

emilyrohrbough:


docs/cloud/account-management/projects.mdx
```diff
@@ -201,17 +209,37 @@ to learn more.
...
 
-You can lengthen the timeout if you are experiencing runs that have passing / failing tests but are not being
-captured by Cloud due to a timeout. You can shorten the timeout if you need faster feedback.
+- Very long-running end-to-end tests
+- Tests that include operations that take significant time (such as large file uploads)
+- Operations with unpredictable completion times
+
+The default is **90 minutes**. You can lengthen the timeout if your tests or CI need
+more time to send results, or shorten it if you want faster feedback when a run has
+stopped sending data.
```
jennifer-shehane: @emilyrohrbough I would say that the use case is more often someone wanting to set it shorter so that their runs can quickly time out. If someone needs to set it higher than 90 minutes, they should be looking at other ways to optimize their run honestly as that's a really gracious amount of time.

emilyrohrbough: good point

PR #6470, opened 2026-05-28
docs: clarify cypress.config.js format depends on package.json type field

### LINKED ISSUE #6076: Cypress tsconfig.json configuration with `"type": "module"`?
## What would you like?

Document how to set up Cypress with TypeScript when `package.json` contains `"type": "module"`.

## Why is this needed?

[Tooling > TypeScript](https://docs.cypress.io/app/tooling/typescript-support) states that Cypress tests can be written in TypeScript and includes a section [Configure tsconfig.json](https://docs.cypress.io/app/tooling/typescript-support#Configure-tsconfigjson).

The Node.js [type](https://nodejs.org/docs/latest/api/packages.html#type) field can be defined as `module`, `commonjs` or left undefined. If the field `type` is defined as `module`, then `.js` files are treated as [CommonJS](https://nodejs.org/docs/latest/api/modules.html).

## Other

According to https://www.typescriptlang.org/docs/handbook/modules/reference.html#the-module-compiler-option

> node16 and nodenext are the only correct module options for all apps and libraries that are intended to run in Node.js v12 or later, whether they use ES modules or not.

Using the following in `tsconfig.json` appears to work

```text
"module": "nodenext"
```

however I don't have experience in TypeScript to want to be the author of a documentation change.

- Also https://github.com/cypress-io/cypress-documentation/issues/5262 is still open and waiting for an experienced TypeScript user to update the documentation to cover Cypress non-legacy (Cypress 10 and above) use with TypeScript.

### LINKED ISSUE #6385: Improve documentation about scaffolding `cypress.config.js` (CommonJS / ESM)
### Description

A Cypress project's `package.json` field [type](https://nodejs.org/api/packages.html#type) can be configured with optional values `commonjs`  and `module` (ESM). The file `cypress.config.js` is configured during scaffolding according to the value:

| `type`     | Cypress config         |
| ---------- | ---------------------- |
| undefined  | `require` for CommonJS |
| `commonjs` | `require` for CommonJS |
| `module`   | `import` for ESM       |

This is not mentioned in the section [Quick Configuration](https://docs.cypress.io/app/get-started/open-the-app#Quick-Configuration).

Also, the Cypress Configuration documentation section [Intelligent Code Completion](https://docs.cypress.io/app/references/configuration#Intelligent-Code-Completion) seems to equate CommonJS to JavaScript and ESM to TypeScript.

This could be better explained.

At this time, scaffolding does not recognize TypeScript. This is a different issue, described in https://github.com/cypress-io/cypress/issues/32442

jennifer-shehane:
- closes https://github.com/cypress-io/cypress-documentation/issues/6385
- Closes https://github.com/cypress-io/cypress-documentation/issues/6076

Explain that when Cypress scaffolds cypress.config.js, the module syntax
(CommonJS vs ESM) is determined by the `type` field in package.json —
not by whether the project uses JavaScript or TypeScript.

- Quick Configuration section: add table showing how package.json `type`
  maps to config file syntax
- Configuration File section: add the same table and clarify that .mjs/.cjs
  extensions are a way to fix the format regardless of package.json
- Intelligent Code Completion section: add an info callout explaining that
  the cypress.config.js tab shows CommonJS, and projects with "type":"module"
  use ESM syntax instead (same as the .ts tab)

https://claude.ai/code/session_01Wq8kUMxa9ch9cjiodfEJk6

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime or configuration behavior impact.
> 
> **Overview**
> This PR updates Cypress docs so **`cypress.config.js` module syntax is tied to `package.json` `"type"`**, not to JS vs TS.
> 
> **Open the app** and **Configuration** now include the same mapping table (`"module"` → ESM, `"commonjs"` or unset → CommonJS). The configuration page also reframes **`.mjs` / `.cjs`** as ways to force module format regardless of `package.json`, and adds an **info callout** before config examples noting the JS tab shows CommonJS unless `"type": "module"` (then it matches the TS/ESM tab).
> 
> **TypeScript support** adds guidance for ESM projects: set `"module": "nodenext"` in the Cypress `tsconfig.json`, with a sample config and a note that Vite-style projects can inherit bundler settings instead.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d46bef47cbca36f4ac2d251bdcdd1b48f6617a61. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/app/tooling/typescript-support.mdx
```diff
@@ -70,6 +70,34 @@ with the following configuration:
...
+
+If your project sets `"type": "module"` in `package.json`, Node.js treats `.js`
+files as ES modules. In this case, add `"module": "nodenext"` to your Cypress
+`tsconfig.json` so TypeScript generates module-compatible output:
+
+```json title="cypress/tsconfig.json"
+{
+  "compilerOptions": {
+    "target": "es6",
+    "lib": ["es6", "dom"],
```
AtofStryker: Should we update the recommended targets here, as well as on our typescript docs page? The browsers we support should be within the ES2024 limits
```suggestion
    "target": "ES2024",
    "lib": ["ES2024", "dom"],
```

docs/app/get-started/open-the-app.mdx
```diff
@@ -114,6 +114,15 @@ to review. For more information about the generated config check out the
...
 just scroll down and hit "Continue".
 
+The format of the scaffolded `cypress.config.js` is determined by the
+[`type`](https://nodejs.org/api/packages.html#type) field in your project's
+`package.json`:
+
+| `package.json` `type`   | Config file syntax                    |
+| ----------------------- | ------------------------------------- |
+| `"module"`              | ESM (`import`/`export default`)       |
+| `"commonjs"` or not set | CommonJS (`require`/`module.exports`) |
```
AtofStryker, line 124: does TypeScript need to be accounted for here?

docs/app/references/configuration.mdx
```diff
@@ -47,6 +57,16 @@ automatic code completion for configuration in many popular code editors. While
 it's not strictly necessary for Cypress to parse your configuration, we
 recommend wrapping your config object with `defineConfig()` like this:
 
+:::info
+
+The `cypress.config.js` tab below shows CommonJS syntax (`require`/`module.exports`),
+which applies when `"type"` in `package.json` is `"commonjs"` or not set. If
+your project sets `"type": "module"`, your `cypress.config.js` will instead use
+ESM syntax (`import`/`export default`) — the same syntax shown in the
```
AtofStryker, line 65: probably implied, but the syntax may be the same but behavior may be different depending on whats in the config. I don't think there is anything for us to do here just something to be aware of

AtofStryker:
closing in favor of #6559

PR #6740, opened 2026-07-21
docs(ui-coverage): add "what makes a view" diagram

jennifer-shehane:
## Summary

Adds an SVG diagram to the **"What makes a view"** section of the UI Coverage Views core-concept page (`docs/ui-coverage/core-concepts/views.mdx`) illustrating the two sources a view can come from.

## Why

The "What makes a view" distinction was a two-bullet list that's easier to grasp visually:

- **End-to-end tests** — many unique URLs are consolidated into fewer URL-pattern views (e.g. `/users/1`, `/users/2` → `/users/*`).
- **Component tests** — each spec file maps one-to-one to a view named by its file path.

The diagram shows the many-to-one consolidation on one side and the one-to-one spec-to-view mapping on the other.

It follows the conventions already established by the existing core-concepts SVGs (`how-ui-coverage-fits-together.svg`, `element-grouping-before-after.svg`): a bordered card background so it reads on both light and dark site themes, a descriptive `aria-label` baked in for screen readers, and `noBorder` on the `DocsImage` so it doesn't get a doubled frame. Colors were checked for WCAG AA contrast, and the diagram doesn't rely on color alone to convey meaning.

Closes #

## Notes for reviewers

- One new asset: `static/img/ui-coverage/core-concepts/view-sources.svg`.
- A second diagram ("How the UI Coverage score is calculated") was included in earlier revisions of this branch but has been **removed**. The UI Coverage PM flagged that its model was inaccurate: the real run score counts interactive nodes distinct across the whole run (an element on several views counts once), collapses element groups to a single point, and also factors in untested links and empty views — so per-view element counts don't sum to a run total, as that diagram implied. Rather than ship a misleading visual, it was dropped from this PR.
- Separately: the existing prose in "Views in your coverage report" ("the total number of tested elements divided by the total number of interactive elements… each view contributes in proportion to how many elements it contains") has the same simplification. That text predates this PR and is left unchanged here; happy to follow up with the PM in a separate change.

https://claude.ai/code/session_01ToW1pUPQFhL59dMorwoK83


static/img/ui-coverage/core-concepts/weighted-coverage-score.svg
marktnoonan, line 1: Haven't reviewed updates in any detail, so apologies if I am missing context here, but I want to flag that this diagram doesn't seem right as I was passing. The score is usually more complicated because the same element selectors are often present on more than one page. The total number of elements in the application is almost always **smaller** than the total if you added up all the elements on each view.

Overall score also depends a lot on the total number of untested links compared to the number of views.

IIRC the formula is "total tested elements count + total views count" divided by "total views count + total untested links count + total untested elements count + total tested elements count".

jennifer-shehane: I'm trying to settle on something that's accurate but simplified (not necessarily precise in the exact deep details), let me take another look given your feedback.

If I can't show it simplified - might just skip it, so lets see.

jennifer-shehane: Ended up removing

PR #6790, opened 2026-08-06
docs: replace `experimentalMemoryManagement` with `manageBrowserMemory` (Cypress 16)

mschile:
Documents the Cypress 16 breaking change from https://github.com/cypress-io/cypress/pull/34516 — `experimentalMemoryManagement` is removed and replaced by the new public `manageBrowserMemory` option, which defaults to `true`. Closes https://github.com/cypress-io/cypress/issues/34385.

Targets `release/16.0.0`, not `main` — this is a v16-only change.

### Changes

- **`docs/app/references/experiments.mdx`** — removed the `experimentalMemoryManagement` row from the experiments table. The 12.4.0 row in the historical changelog table at the bottom of the page is left alone, since it is accurate about the past.
- **`docs/app/references/configuration.mdx`** — added `manageBrowserMemory` to the Global options table with a default of `true`. Deliberately *not* added to the Test Configuration overridable-options list, since the option has no `overrideLevel` and cannot be changed per test, per suite, or via `Cypress.config()`.
- **`docs/app/references/migration-guide.mdx`** — new section under "Migrating to Cypress 16.0" modeled on the sibling `experimentalFastVisibility` and `experimentalSourceRewriting` sections. It covers the rename, the new `true` default, and the trap that matters most: there is no alias and no value remapping, so anyone who was setting `experimentalMemoryManagement: false` to opt out must now set `manageBrowserMemory: false` or memory management silently turns on for them.
- **`docs/app/references/error-messages.mdx`** — reworded the browser-crash remediation. It previously told users to try enabling `experimentalMemoryManagement`; it now notes memory management is on by default and points at `manageBrowserMemory`. This matches the bullet removed from `RENDERER_CRASHED` in the code PR. The `numTestsKeptInMemory` sentence that follows is unchanged.
- **`docs/app/guides/test-performance.mdx`** — rewrote the "Manage memory accumulation" section rather than find-and-replacing the option name, and corrected a long-standing inaccuracy in it. The section claimed memory management works "by clearing the browser's heap after each test" and that the cost is "paid on every test transition." It isn't: Cypress samples memory on an interval and forces a garbage collection before the next test **only** when a sample crossed the threshold (see `packages/server/lib/browsers/memory/index.ts`). The section now describes that accurately. Since the default is flipped, the old opt-in advice is replaced with how to opt out. Also updated the priority-checklist entry and the cross-reference from the CI resources section.

`docs/app/references/changelog.mdx` is untouched — those are historical release notes.

`cypress.config.ts` is also untouched. An earlier revision of this PR dropped `experimentalMemoryManagement` and `experimentalFastVisibility` from it, but this repo still runs Cypress 15, so those options are valid here today and removing them would change the docs site's own test behavior. They should be cleaned up as part of upgrading this repo to Cypress 16, not here.

### Verified

`npm run build` passes, which exercises Docusaurus' `onBrokenLinks: 'throw'` / `onBrokenAnchors: 'throw'`, so every new cross-reference resolves. `npm run lint:frontmatter` and `prettier --check` on the changed `.mdx` files both pass.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime impact; reviewers may want to fix the configuration history row that says `manageBrowserMemory` was renamed to itself.
> 
> **Overview**
> Documents the Cypress 16 change where **`experimentalMemoryManagement`** is removed in favor of the public **`manageBrowserMemory`** option (default **`true`** on Chromium).
> 
> **Configuration & migration:** Adds **`manageBrowserMemory`** to the global options table and a dedicated section describing threshold-based GC (not per-test heap clears). Removes the experiment from **`experiments.mdx`**. Adds a **Migrating to Cypress 16.0** section covering opt-out via **`manageBrowserMemory: false`** when users previously set **`experimentalMemoryManagement: false`**.
> 
> **Cross-links:** Updates browser-crash guidance in **`error-messages.mdx`**, trims the test-performance priority checklist (drops the old memory-management item and renumbers), removes the standalone “Manage memory accumulation” section, and points “Keep Cypress up to date” / CI resource tuning at **`manageBrowserMemory`** instead of the experiment.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f82de419092bc832bf6f6c046525b69fa84d90c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jennifer-shehane (approved):
I made some updates, moved things around a bit.

PR #6806, opened 2026-08-13
docs: add the `forceHttp1` configuration option

### LINKED ISSUE #33847: Add `forceHttp1` config option and remove `CYPRESS_INTERNAL_DISABLE_PROXY`
## Summary

Add a `forceHttp1` boolean config option that selects between the CDP-based network path and the legacy HTTP/1 MITM proxy path. In Cypress 16, `forceHttp1: false` (the CDP path) is the **default**. Setting `forceHttp1: true` forces the legacy MITM proxy, which downgrades AUT traffic to HTTP/1.

This replaces the internal `CYPRESS_INTERNAL_DISABLE_PROXY=1` env var: `CYPRESS_INTERNAL_DISABLE_PROXY=1` ≡ `forceHttp1: false`. The env var is removed entirely.

`forceHttp1` is **deprecated at introduction** (same pattern as `injectDocumentDomain`): the long-term plan is to support the CDP path on Chrome and Firefox, at which point the MITM escape hatch is removed. Firefox CDP support is **not** in the MVP — Firefox falls back to the MITM path in 16. WebKit support still requires a spike to determine feasibility.

## Scope

**In:**

- New `ConfigOption` in `packages/config/src/options.ts`: `forceHttp1`, `validation: validate.isBoolean`, `defaultValue: false`, `requireRestartOnChange: 'server'`. Available in e2e and component testing (root-level and overridable per testing type).
- Deprecation warning when `forceHttp1: true` is set, following the `injectDocumentDomain` breaking-options pattern (`isWarning: true` + a named error template in `@packages/errors`, e.g. `FORCE_HTTP1_DEPRECATION`). The warning keys off the value — explicitly setting `false` (the default) does not warn.
- Type surfaced in `cli/types/cypress.d.ts` (marked `@deprecated`) and `@packages/types`.
- Replace the env read in `packages/server/lib/util/is-proxy-disabled.ts` with the resolved config value. All existing call sites keep working through the same util.
- Remove `CYPRESS_INTERNAL_DISABLE_PROXY` from `CYPRESS_RESERVED_ENV_VARS` in `packages/config/src/project/utils.ts` and from all CI exports.
- Browsers without CDP-path support (Firefox, WebKit in the MVP): **no error**. Log a debug message (`cypress:server:*` namespace) that the browser doesn't use the CDP path yet and fall back to the MITM path for that browser. The effective mode is per-browser: CDP path only when `forceHttp1 === false` **and** the browser is Chromium-family.
- Driver test coverage for `forceHttp1: true`, mirroring `injectDocumentDomain`:
  - `packages/driver/cypress.config-forceHttp1.ts` — spreads the base config, sets `forceHttp1: true`, narrows `specPattern` to network-affected specs.
  - `package.json` script (`cypress:run:force-http1`) and a `force-http1` boolean parameter on the `run-driver-integration-tests` CI job, recorded as its own group (`5x-driver-force-http1-chrome`).
- CI cleanup: delete the `*-cdp-remediated` jobs, the `*-cdp` opt-in twins, the `run-proxy-disabled-tests` pipeline parameter, the `halt-unless-proxy-disabled-enabled` command, and the `disable-proxy` job parameters. The default jobs now exercise the CDP path; the `force-http1` driver variant plus the Firefox/WebKit jobs cover the MITM path.
- Documentation (cypress-io/cypress-documentation): configuration-page entry for `forceHttp1`, plus a dedicated page explaining what the option does (CDP interception vs MITM proxy, browser support matrix, deprecation status and removal plan, when to set `true`).
- Changelog entry in `cli/CHANGELOG.md` for Cypress 16.

**Out:**

- Firefox CDP-path support (planned long-term, separate work).
- WebKit CDP-path support (needs a spike to determine feasibility before any commitment).

## Implementation notes

- The server currently decides proxy mode once, globally, at startup. Because Firefox/WebKit fall back per-browser, the MITM proxy machinery must still be available when a non-Chromium browser launches — the mode decision moves from server start to browser launch. `requireRestartOnChange: 'server'` covers config-file edits in open mode (restart banner).
- The `injectDocumentDomain` deprecation warning fires on key presence; for `forceHttp1` the warning should fire only when the value is `true`.
- Driver/spec code that currently gates on proxy-disabled via `Cypress.expose('PROXY_DISABLED')` can gate on `Cypress.config('forceHttp1')` + browser family instead.

## Acceptance criteria

- Given no config, `cypress run --browser chrome` uses the CDP path (`forceHttp1` resolves `false`).
- Given `forceHttp1: true`, all browsers use the MITM proxy path; behavior matches Cypress 15 defaults — and a deprecation warning is printed stating the option will be removed in a future version once the CDP path is supported long-term on Chrome and Firefox.
- Given `forceHttp1: false` (or unset) and `--browser firefox` (or webkit), Cypress launches successfully on the MITM path and emits a debug message; no error, exit code unchanged.
- Changing `forceHttp1` in the config file during `cypress open` prompts a server restart before taking effect.
- `forceHttp1: 'true'` (string) fails config validation with the standard message.
- The option is settable in both `e2e` and `component` blocks.
- The driver `force-http1` CI job runs the narrowed spec subset with `forceHttp1: true` and passes.
- `CYPRESS_INTERNAL_DISABLE_PROXY` no longer appears anywhere in the repo.

## Deliverables

- [x] `ConfigOption` entry + deprecation warning + config unit tests (`@packages/config`, `@packages/errors`)
- [x] Types in `cli/types/cypress.d.ts` (`@deprecated`) + `@packages/types`
- [x] `is-proxy-disabled.ts` reads resolved config; env var removed from `CYPRESS_RESERVED_ENV_VARS`; server unit tests updated
- [x] Per-browser fallback + debug message; unit test for the Firefox/WebKit fallback
- [x] `cypress.config-forceHttp1.ts` + `cypress:run:force-http1` script + `force-http1` driver CI job
- [x] CircleCI cleanup (remediated jobs, opt-in parameter, `disable-proxy` parameters, env exports)
- [x] `cli/CHANGELOG.md` entry
- [x] cypress-documentation PR: config entry + dedicated `forceHttp1` page (including deprecation notice)

AtofStryker:
## Release

Target release: Cypress App 16.0.0

Documents the new [`forceHttp1`](https://github.com/cypress-io/cypress/issues/33847) configuration option and the network path change that comes with it: Chrome, Chromium, and Edge now intercept test traffic on the native browser network instead of routing it through the Cypress HTTP/1.1 proxy. Firefox and WebKit have no native browser network path yet, and Electron is deprecated as a test browser, so all three continue to use the legacy path.

## Included changes

- [x] **Changelog entry** — 16.0.0 bullets in `docs/app/references/changelog.mdx` for the network path change and the new option.
- New guide, `docs/app/guides/native-network-interception.mdx` — what changed, which browsers are affected, how to use the legacy path, the behavior differences with before/after examples, and how to write tests that pass in every browser.
- `docs/app/references/configuration.mdx` — `forceHttp1` row in the Global options table, a `### forceHttp1` notes section, and a History entry.
- `docs/app/references/migration-guide.mdx` — "Chrome, Chromium, and Edge use the native browser network" under Migrating to Cypress 16.0.
- `docs/api/commands/intercept.mdx` — corrects the `httpVersion` docs and cross-links the guide.
- Two partials (`_force-http1-deprecation.mdx`, `_native-browser-network-browsers.mdx`) for the deprecation notice and browser matrix, which each render in more than one place.

## Notes for reviewers

**`forceHttp1` is not implemented yet.** The corresponding Cypress PR for [#33847](https://github.com/cypress-io/cypress/issues/33847) is still in review, so this should merge alongside or after it.

The `httpVersion` change in `intercept.mdx` corrects copy that already shipped in #6801. That text said the property is unreported in Chromium-based browsers; Electron is Chromium-based and stays on the legacy path, so it still reports a value.

Behavior differences and examples are drawn from the contracts pinned by the driver tests, so they describe shipped behavior rather than intent.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime or test code; risk is limited to doc accuracy and timing relative to the not-yet-merged `forceHttp1` implementation.
> 
> **Overview**
> Documents **Cypress 16** network changes: Chrome, Chromium, and Edge intercept traffic on the **native browser network** (HTTP/2/3 in production-like conditions), while Firefox, WebKit, and Electron stay on the **legacy HTTP/1.1 proxy path**.
> 
> Adds a full **Native network interception** guide with browser matrix, deprecated **`forceHttp1`** escape hatch, and `cy.intercept()` behavior deltas (headers, `httpVersion`, cache/revalidation, timeouts, etc.) plus cross-browser test patterns. Updates **changelog**, **configuration** (`forceHttp1` global option + section), **migration guide**, and **`intercept.mdx`** (`httpVersion` wording + note). Introduces shared MDX partials wired through **`MDXComponents.js`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb699534ca7c1e050d5c1ea6e3ebd931136d48d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/app/guides/native-network-interception.mdx
```diff
@@ -0,0 +1,300 @@
...
+
+:::
+
+Changing `forceHttp1` restarts the Cypress server, and the option cannot be set per test
+or per suite.
+
+If setting `forceHttp1` to `true` is what makes your suite pass, and the difference is
+not one of the documented behaviors below, please
+[open an issue](https://github.com/cypress-io/cypress/issues/new/choose) with what you
+are seeing. That points to a bug on the native browser network path, and we would rather
```
baus: Is bug the way we want to describe this. Maybe "discrepancy?"

docs/app/guides/native-network-interception.mdx
```diff
@@ -0,0 +1,300 @@
...
+fix it than have you keep the option set. `forceHttp1` is temporary, so anything that
```
baus: fix it than require you to keep the option set.

docs/app/guides/native-network-interception.mdx
```diff
@@ -0,0 +1,300 @@
...
+
+cy.visit('/')
+cy.reload()
+
+// Both navigations reach the intercept in Cypress 16
+cy.get('@config.all').should('have.length', 2)
+```
+
+Tests that assert a resource was served from cache, rather than asserting on what your
+application rendered, are the ones to look at.
```
baus: "are the ones the look at" -> "could be affected" ?

docs/app/guides/native-network-interception.mdx
AtofStryker: "Discrepancy" to me covers two different things:

- differences that are unintended (bugs)
- differences that are intentional (the documented behavior changes on this page)

This sentence only applies after the reader has ruled out the documented behaviors, so what's left is unintended by definition — which is why I want to keep "bug" here. The only thing "bug" misses is an intentional change we forgot to write down, and in that case we'd still want an issue for it. I changed the text to: "either a bug on the native browser network path or a gap in this page." Tell me if this works for you.

docs/app/guides/native-network-interception.mdx
AtofStryker: I used your wording. Thanks.

docs/app/guides/native-network-interception.mdx
AtofStryker: I like the direction, but "could be" undersells it — a test that asserts a stub was served from cache will fail here every time, not sometimes. I used your wording and made it stronger: "will be affected." Thank you!

docs/app/guides/native-network-interception.mdx
```diff
@@ -0,0 +1,390 @@
...
+
+<ProductHeading product="app" />
+
+# Native Network Interception
+
+## What changed
+
+Before Cypress 16, Cypress routed every request your application made through the
+legacy network path. That path is how Cypress observed, stubbed, and recorded network
+traffic, and it only spoke HTTP/1.1. Requests that your server would have served over
```
mschile:
```suggestion
traffic, and it only supported HTTP/1.1. Requests that your server would have served over
```

docs/app/guides/native-network-interception.mdx
```diff
@@ -0,0 +1,390 @@
...
+Changing `forceHttp1` restarts the Cypress server, and the option cannot be set per test
+or per suite.
+
+If setting `forceHttp1` to `true` is what makes your suite pass, and the difference is
+not one of the documented behaviors below, please
+[open an issue](https://github.com/cypress-io/cypress/issues/new/choose) with what you
+are seeing. That points to either a bug on the native browser network path or a gap in
+this page — either way, we would rather fix it than require you to keep the option set.
+`forceHttp1` is temporary, so anything that depends on it needs a real fix before it is
+removed.
```
mschile:
```suggestion
If setting `forceHttp1` to `true` allows your suite to pass, and the difference is
not one of the documented behaviors below, please
[open an issue](https://github.com/cypress-io/cypress/issues/new/choose) with what you
are experiencing so it can be investigated.
```

docs/app/guides/native-network-interception.mdx
```diff
@@ -0,0 +1,390 @@
...
+
+When you update these tests, assert on both the document and the subresources. Only the
+document load changed; the insecure-origin policy still applies to everything the browser
+fetches on its own.
+
+The Cypress 15 failure also varied by browser: Chromium timed out on the load, while
+Firefox rendered the raw compressed bytes as garbled text. Firefox stays on the legacy
+network path, so that behavior is unchanged.
+
+### Revalidated responses report `200` instead of `304`
```
mschile, line 260: I believe this unexpected behavior was also reported here: https://cypressio.slack.com/archives/C04FGLGRM4L/p1775226364966649

docs/app/references/migration-guide.mdx
```diff
@@ -212,6 +212,39 @@ it(
...
+
+```js
+{
+  forceHttp1: true
+}
+```
+
+:::
+
+If setting `forceHttp1` to `true` is what makes your suite pass, and the difference is not one of the documented behaviors, please [open an issue](https://github.com/cypress-io/cypress/issues/new/choose) with what you are seeing. That points to a bug on the native browser network path, and we would rather fix it than have you keep the option set. `forceHttp1` is temporary, so anything that depends on it needs a real fix before it is removed.
```
mschile:
```suggestion
If setting `forceHttp1` to `true` allows your suite to pass, and the difference is not one of the documented behaviors, please [open an issue](https://github.com/cypress-io/cypress/issues/new/choose) with what you are experiencing so it can be investigated.
```

docs/app/references/changelog.mdx
```diff
@@ -40,6 +40,7 @@ Refer to the [16 Migration Guide](/app/references/migration-guide#Migrating-to-C
...
 
+- Chrome, Chromium, and Edge intercept test traffic on the native browser network instead of the legacy network path, so your application connects to your server directly and negotiates whatever protocol the server supports (HTTP/1.1, HTTP/2, or HTTP/3), exactly as it does in production, rather than always being downgraded to HTTP/1.1. Firefox and WebKit continue to use the legacy network path because Cypress has not implemented a native browser network path for them yet, and Electron stays on the legacy path because it is deprecated as a test browser in Cypress 16. The new [`forceHttp1`](/app/references/configuration#forceHttp1) option is a temporary escape hatch that routes every browser through the legacy network path while you migrate or while an issue you have filed is fixed; if undocumented behavior is what makes your suite pass, [open an issue](https://github.com/cypress-io/cypress/issues/new/choose). See [Native network interception](/app/guides/native-network-interception) and the [Migration Guide](/app/references/migration-guide#Chrome-Chromium-and-Edge-use-the-native-browser-network). Addresses [#3708](https://github.com/cypress-io/cypress/issues/3708).
```
mschile: I would probably condense this quite a bit and have it focus on the feature. Maybe something like:

```suggestion
Cypress now supports HTTP/2/3 in Chrome, Chromium, and Edge. Your application is tested over the same protocol it uses in production - multiplexed requests, no six-connection ceiling - so modern apps load faster under test and behave the way your users actually experience them. See [Native network interception](/app/guides/native-network-interception) and the [Migration Guide](/app/references/migration-guide#Chrome-Chromium-and-Edge-use-the-native-browser-network). Addresses [#3708](https://github.com/cypress-io/cypress/issues/3708).
```

docs/app/references/migration-guide.mdx
```diff
@@ -212,6 +212,39 @@ it(
 
 Setting `blockHosts` in your Cypress configuration, or with `Cypress.config()` outside of test execution (for example at the top level of a support file or spec file), is unchanged.
 
+### Chrome, Chromium, and Edge use the native browser network
+
+In Cypress 16, Chrome, Chromium, and Edge intercept test traffic through the native browser network. Your application connects to your server directly and negotiates whatever protocol the server supports (HTTP/1.1, HTTP/2, or HTTP/3), exactly as it does in production.
+
+Firefox and WebKit continue to use the legacy network path because Cypress has not implemented a native browser network path for them yet. Electron is Chromium-based and could use one, but it is [deprecated as a test browser](#Electron-browser-deprecated) in Cypress 16, so it stays on the legacy path too. Suites that run only in Firefox, WebKit, or Electron are unaffected.
```
mschile:
```suggestion
Firefox, WebKit, and Electron continue to use the legacy network path and are unaffected.
```

docs/app/references/migration-guide.mdx
```diff
@@ -212,6 +212,39 @@ it(
...
+
+[`cy.intercept()`](/api/commands/intercept) works as before, and most suites need no edits. A few behaviors do differ in Chrome, Chromium, and Edge on the native browser network:
+
+- Intercepted responses report a decoded body and no `content-encoding`, `content-length`, or `transfer-encoding` header.
+- `req.httpVersion` is `undefined` and `res.httpVersion` is `null`.
+- Responses the browser refuses to deliver leave `interception.response` as `undefined`.
+- Responses answered inside Cypress, such as stubs, are not stored in the browser's HTTP cache.
+
+See [Native network interception](/app/guides/native-network-interception) for the full list with examples.
```
mschile: I wonder if we should just link to the guide instead of listing a few of the differences.

mschile (approved):
Added a couple suggestions but looks good overall!

docs/app/guides/native-network-interception.mdx
AtofStryker: Applied in 511d6aaa1.

docs/app/guides/native-network-interception.mdx
AtofStryker: I used your wording. Applied in 511d6aaa1.

docs/app/guides/native-network-interception.mdx
AtofStryker: That report is the legacy-path half of this section — the intercept saw the bare `304` with no body while DevTools showed the merged cached response. On the native browser network the intercept gets the same merged `200` DevTools shows, so that confusion goes away.

docs/app/references/migration-guide.mdx
AtofStryker: Applied in eb699534c.

docs/app/references/changelog.mdx
AtofStryker: I used your wording. Applied in eb699534c.

docs/app/references/migration-guide.mdx
AtofStryker: Applied in eb699534c.

docs/app/references/migration-guide.mdx
AtofStryker: Done in eb699534c — the list duplicated the guide's section headings and would have drifted; it's a single pointer now.

PR #6808, opened 2026-08-13
docs: remove references to the Cypress app proxying requests

baus:
## Summary

Removes user-facing statements that describe the Cypress app as proxying HTTP, HTTPS, and WebSocket traffic. Each passage was reworded rather than deleted, so the guidance it supported (WebSocket behavior, `cy.intercept('*')` overhead, `hosts` scoping, basic-auth headers, `ECONNRESET` troubleshooting) still reads correctly.

## Why

The docs asserted the proxy architecture in a dozen places as user-facing explanation, and those statements are no longer how we want the app's request handling described. Rewording keeps the practical advice intact while dropping the mechanism claim.

## Notes for reviewers

16 references removed across 8 files:

- `cross-origin-testing.mdx` — dropped the "Proxies all HTTP / HTTPS traffic" bullet from "What Cypress does under the hood"; removed "network proxying" from the same-origin strategies sentence
- `faq.mdx` — WebSocket answer, performance-testing answer, and "to your proxy" in the SSL certificate heading
- `trade-offs.mdx`, `network-requests.mdx` — the two other copies of the "transparently proxies WebSocket connections" sentence
- `test-performance.mdx` — wildcard-intercept overhead prose, its code comment, and the summary bullet
- `configuration.mdx` — `hosts`, `removeSRIAttributes` prose, and the options-table row
- `visit.mdx` — the basic-auth callout now names the `Authorization` header explicitly instead of referring to "this header … at the network proxy level"
- `troubleshooting.mdx` — `ECONNRESET` wording and the Electron fastest-path check

Deliberately left alone, and worth confirming these are the right calls:

- **The SSL certificate answer in `faq.mdx`** — the heading no longer says "to your proxy", but the answer still reads "Cypress modifies network traffic in real time and therefore must sit between your server and the browser." Removing that clause left the "No" with nothing to justify it, so the original answer stands until we decide what should replace it.
- **`error-messages.mdx`** (4 mentions) — "Whoops, we can't run your tests" names proxy bypass as a cause, and the GPO section exists *because* Chrome's `ProxySettings`/`ProxyMode` policies break Cypress. Rewording risks making working troubleshooting advice incoherent.
- **`troubleshooting.mdx:133`** — the `📁 proxy` entry in the app-data directory listing. Only correct to remove if the app no longer creates that folder; not verified against a real 16 install.
- **`troubleshooting.mdx:254`** — the `cypress:net-stubbing:*` DEBUG namespace description ("Network interception in the proxy layer").
- `changelog.mdx` (89 hits, historical), `proxy-configuration.mdx` and `advanced-installation.mdx` (corporate/upstream proxy setup), and all third-party proxy mentions (BrowserMob, Charles Proxy, VPNs, firewall guidance).

**Anchor change:** the SSL certificate heading dropped "to your proxy", so its slug changed. Nothing in the repo links to it, but external inbound links will no longer resolve to that section — happy to add a redirect if wanted.

---

> [!NOTE]
> **Low Risk**
> Documentation-only rewording with no code or config changes; main risk is slight messaging inconsistency where proxy terminology remains in untouched troubleshooting/error pages.
> 
> **Overview**
> Updates **user-facing docs** so they no longer describe the Cypress app as **proxying** HTTP, HTTPS, or WebSocket traffic. Wording is **replaced**, not deleted, so behavior guidance (WebSockets, `cy.intercept`, `hosts`, basic auth, performance, troubleshooting) still reads correctly.
> 
> **Cross-origin testing** drops the “Proxies all HTTP / HTTPS traffic” bullet and removes “network proxying” from the same-origin strategies list; the under-the-hood section now emphasizes URL matching and browser network APIs only.
> 
> **FAQ, network requests, and trade-offs** no longer say Cypress “transparently proxies” WebSockets; they state that WebSocket traffic works during tests and keep the existing caveats about stubbing frames.
> 
> **Test performance** reframes wildcard `cy.intercept('*')` overhead as **interception/processing** instead of routing every request through a proxy.
> 
> **Configuration** (`hosts`, `removeSRIAttributes`, options table) and **`cy.visit` basic auth** drop “proxy” / “proxy-rewritten” language in favor of Cypress-scoped or header-focused explanations.
> 
> **Troubleshooting (`ECONNRESET`)** refers to the connection between the browser and **Cypress** (and Electron using a different **network path**) instead of a “local proxy.”
> 
> The SSL FAQ answer still says Cypress must sit between server and browser (unchanged per PR notes). Other proxy mentions (error messages, `📁 proxy` folder, debug namespaces, corporate proxy setup) were intentionally left alone.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40214aa85b4904421c5a100253e12a30f5299244. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/api/commands/visit.mdx
+++ b/docs/api/commands/visit.mdx
@@ -155,9 +155,8 @@ cy.visit('https://wile:coyote@example.cypress.io')
 
 :::info
 
-Cypress will automatically attach this header at the network proxy level,
-outside of the browser. Therefore you **will not** see this header in the Dev
-Tools.
+Cypress attaches the `Authorization` header outside of the browser. Therefore
+you **will not** see this header in the Dev Tools.
 
 :::
 
--- a/docs/app/faq.mdx
+++ b/docs/app/faq.mdx
@@ -1139,7 +1139,7 @@ browser.
 
 [Check out the "Node Modules" example recipe.](/app/references/recipes#Fundamentals)
 
-### <Icon name="angle-right" /> Is there a way to give a proper SSL certificate so the page doesn't show up as "not secure"?
+### <Icon name="angle-right" /> Is there a way to give a proper SSL certificate to Cypress so the page doesn't show up as "not secure"?
 
 No, Cypress modifies network traffic in real time.
 
--- a/docs/app/faq.mdx
+++ b/docs/app/faq.mdx
@@ -1139,7 +1139,7 @@ browser.
 
 [Check out the "Node Modules" example recipe.](/app/references/recipes#Fundamentals)
 
-### <Icon name="angle-right" /> Is there a way to give a proper SSL certificate to your proxy so the page doesn't show up as "not secure"?
+### <Icon name="angle-right" /> Is there a way to give a proper SSL certificate so the page doesn't show up as "not secure"?
 
 No, Cypress modifies network traffic in real time and therefore must sit between
 your server and the browser. There is no other way for us to achieve that.
--- a/docs/app/faq.mdx
+++ b/docs/app/faq.mdx
@@ -100,8 +100,7 @@ testing).
 
 ### <Icon name="angle-right" /> We use WebSockets; will Cypress work with that?
 
-Yes. Cypress transparently proxies WebSocket connections, so your application's
-WebSocket traffic will work as expected during tests.
+Yes. Your application's WebSocket traffic will work as expected during tests.
 
 However, **stubbing or mocking individual WebSocket frames/messages** is not
 natively supported. If your tests need to control the messages sent over a
@@ -1140,10 +1139,9 @@ browser.
 
 [Check out the "Node Modules" example recipe.](/app/references/recipes#Fundamentals)
 
-### <Icon name="angle-right" /> Is there a way to give a proper SSL certificate to your proxy so the page doesn't show up as "not secure"?
+### <Icon name="angle-right" /> Is there a way to give a proper SSL certificate so the page doesn't show up as "not secure"?
 
-No, Cypress modifies network traffic in real time and therefore must sit between
-your server and the browser. There is no other way for us to achieve that.
+No, Cypress modifies network traffic in real time.
 
 ### <Icon name="angle-right" /> Is there any way to detect if my app is running under Cypress?
 
@@ -1621,8 +1619,8 @@ XState model state library.
 ### <Icon name="angle-right" /> Can Cypress be used for performance testing?
 
 Cypress is not built for performance testing. Because Cypress instruments the
-page under test, proxies the network requests, and tightly controls the test
-steps, Cypress adds its own overhead. Thus, the performance numbers you get from
+page under test and tightly controls the test steps, Cypress adds its own
+overhead. Thus, the performance numbers you get from
 Cypress tests are slower than "normal" use. Still, you can access the native
 `window.performance` object and grab the page time measurements, see the
 [Evaluate performance metrics](https://github.com/cypress-io/cypress-example-recipes#testing-the-dom)
--- a/docs/app/faq.mdx
+++ b/docs/app/faq.mdx
@@ -1139,9 +1139,10 @@ browser.
 
 [Check out the "Node Modules" example recipe.](/app/references/recipes#Fundamentals)
 
-### <Icon name="angle-right" /> Is there a way to give a proper SSL certificate to Cypress so the page doesn't show up as "not secure"?
+### <Icon name="angle-right" /> Is there a way to give a proper SSL certificate to your proxy so the page doesn't show up as "not secure"?
 
-No, Cypress modifies network traffic in real time.
+No, Cypress modifies network traffic in real time and therefore must sit between
+your server and the browser. There is no other way for us to achieve that.
 
 ### <Icon name="angle-right" /> Is there any way to detect if my app is running under Cypress?
 
--- a/docs/app/guides/cross-origin-testing.mdx
+++ b/docs/app/guides/cross-origin-testing.mdx
@@ -30,14 +30,13 @@ communicate with your remote application at all times. Unfortunately, browsers
 naturally try to prevent Cypress from doing this.
 
 To get around these restrictions, Cypress implements some strategies involving
-JavaScript code, the browser's internal APIs, and network proxying to _play by
-the rules_ of same-origin policy. It's our goal to fully automate the
+JavaScript code and the browser's internal APIs to _play by the rules_ of
+same-origin policy. It's our goal to fully automate the
 application under test without you needing to modify your application's code -
 and we are _mostly_ able to do this.
 
 ### What Cypress does under the hood
 
-- Proxies all HTTP / HTTPS traffic.
 - Changes the hosted URL to match that of the application under test.
 - Uses the browser's internal APIs for network level traffic.
 
--- a/docs/app/guides/network-requests.mdx
+++ b/docs/app/guides/network-requests.mdx
@@ -535,8 +535,7 @@ cy.wait('@new-user').then(console.log)
 
 ## Working with WebSockets
 
-Cypress transparently proxies WebSocket connections, so your application's
-WebSocket traffic will work as expected during tests.
+Your application's WebSocket traffic will work as expected during tests.
 
 ### Limitations
 
--- a/docs/app/guides/test-performance.mdx
+++ b/docs/app/guides/test-performance.mdx
@@ -176,7 +176,7 @@ Apply these optimizations in order. The highest-impact changes are at the top.
 1. [Eliminate arbitrary cy.wait()](#Avoid-arbitrary-waits-and-adjust-timeouts)
 1. [Slim support file and imports](#Keep-support-and-spec-imports-lean) — avoids bundling unused code on every spec startup
 1. [Block third-party requests with blockHosts](#Block-third-party-requests-with-blockHosts) — analytics and monitoring services consume time in test environments; blocked requests resolve immediately
-1. [Avoid wildcard cy.intercept('\*')](#Intercept-only-the-requests-you-need) — limits proxy overhead to requests the test actually needs
+1. [Avoid wildcard cy.intercept('\*')](#Intercept-only-the-requests-you-need) — limits interception overhead to requests the test actually needs
 1. [Use specific DOM selectors](#Use-specific-selectors) — broad selectors like `*` or `div` force the browser to collect every matching node before filtering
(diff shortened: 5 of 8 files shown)


docs/api/commands/visit.mdx
```diff
@@ -155,9 +155,8 @@ cy.visit('https://wile:coyote@example.cypress.io')
 
 :::info
 
-Cypress will automatically attach this header at the network proxy level,
-outside of the browser. Therefore you **will not** see this header in the Dev
-Tools.
+Cypress attaches the `Authorization` header outside of the browser. Therefore
+you **will not** see this header in the Dev Tools.
```
baus, line 160: @AtofStryker is it still correct that the header will not be in Dev Tools?

docs/api/commands/visit.mdx
```diff
@@ -155,9 +155,8 @@ cy.visit('https://wile:coyote@example.cypress.io')
...
```
AtofStryker: I checked this one in a real run, and we can't promise invisibility any more: on the default path `authorization: Basic ...` shows up in the DevTools request headers. It's absent with `forceHttp1: true`. Makes sense — the browser is the one making the request now.

Suggesting wording that leads with the invariant we actually test in `packages/driver/cypress/e2e/issues/573.cy.js` (attached to the visited origin, withheld from others) and softens the DevTools claim rather than dropping it.

```suggestion
Cypress attaches the `Authorization` header for you on requests to the origin
you visited; other origins do not receive it. Depending on the browser, the
header may not appear in the Dev Tools network panel even though it was sent.
```

docs/app/faq.mdx
```diff
@@ -100,8 +100,7 @@ testing).
 
 ### <Icon name="angle-right" /> We use WebSockets; will Cypress work with that?
 
-Yes. Cypress transparently proxies WebSocket connections, so your application's
-WebSocket traffic will work as expected during tests.
+Yes. Your application's WebSocket traffic will work as expected during tests.
```
AtofStryker: To your question about whether this is worth mentioning at all — I'd keep it. It's a recurring evaluation question and the answer got simpler rather than moot. No change needed here: the limitation in the paragraph right below still carries the actionable part.

docs/app/guides/network-requests.mdx
```diff
@@ -535,8 +535,7 @@ cy.wait('@new-user').then(console.log)
 
 ## Working with WebSockets
 
-Cypress transparently proxies WebSocket connections, so your application's
-WebSocket traffic will work as expected during tests.
+Your application's WebSocket traffic will work as expected during tests.
```
AtofStryker: Same as the FAQ — reads fine to me, and the Limitations section below still says what people need. No change needed.

docs/app/references/trade-offs.mdx
```diff
@@ -103,7 +103,7 @@ time.
 
 :::info
 
-**WebSocket support:** Cypress transparently proxies WebSocket connections, and
+**WebSocket support:** WebSocket connections work as expected during tests, and
 [`cy.intercept()`](/api/commands/intercept) can observe the WebSocket upgrade
 request via `resourceType: 'websocket'`. However, stubbing or mocking
 individual WebSocket **frames/messages** is not natively supported. The
```
AtofStryker: The half of this sentence the PR doesn't change is the one I'd fix: we don't intercept WebSockets on either path, so `cy.intercept()` can't observe the upgrade. `resourceType` as a matcher has also been deprecated since 14.0.0.

```suggestion
**WebSocket support:** WebSocket connections work as expected during tests, but
Cypress does not intercept them, so stubbing or mocking individual WebSocket
**frames/messages** is not natively supported. The
```

docs/app/guides/cross-origin-testing.mdx
```diff
@@ -30,14 +30,13 @@ communicate with your remote application at all times. Unfortunately, browsers
...
-the rules_ of same-origin policy. It's our goal to fully automate the
+JavaScript code and the browser's internal APIs to _play by the rules_ of
+same-origin policy. It's our goal to fully automate the
 application under test without you needing to modify your application's code -
 and we are _mostly_ able to do this.
 
 ### What Cypress does under the hood
 
-- Proxies all HTTP / HTTPS traffic.
 - Changes the hosted URL to match that of the application under test.
```
AtofStryker: I'd reword this bullet rather than drop it. "Intercepts" is true regardless of how we do it, and as it stands the list is down to two bullets that leave the reader unaware we touch traffic at all — while the HTTPS callout below still says we modify traffic in real time.

```suggestion
- Intercepts and can modify HTTP / HTTPS traffic.
- Changes the hosted URL to match that of the application under test.
```

docs/app/guides/cross-origin-testing.mdx
```diff
@@ -30,14 +30,13 @@ communicate with your remote application at all times. Unfortunately, browsers
 naturally try to prevent Cypress from doing this.
 
 To get around these restrictions, Cypress implements some strategies involving
-JavaScript code, the browser's internal APIs, and network proxying to _play by
-the rules_ of same-origin policy. It's our goal to fully automate the
+JavaScript code and the browser's internal APIs to _play by the rules_ of
+same-origin policy. It's our goal to fully automate the
 application under test without you needing to modify your application's code -
 and we are _mostly_ able to do this.
```
AtofStryker: Small thing — the edit left a couple of short lines here. Prettier is set to `proseWrap: preserve`, so `--check` won't flag it and it needs doing by hand.

```suggestion
JavaScript code and the browser's internal APIs to _play by the rules_ of
same-origin policy. It's our goal to fully automate the application under test
without you needing to modify your application's code - and we are _mostly_
able to do this.
```

Separately, and more importantly: the HTTPS callout a few lines down (~:52-60) is the passage the certificate finding lands on. It promises "you'll notice Chrome display a warning that the 'SSL certificate does not match'", which doesn't happen on the default path any more — Chrome gets the site's real certificate. Conditional phrasing that holds on both paths:

> Cypress may need to issue its own certificates in order to modify HTTPS traffic. When it does, Chrome warns that the certificate was issued by an unknown authority — this is expected. Under the hood Cypress acts as its own certificate authority and issues certificates dynamically, only for the origin currently under test.

(Worth noting the old wording was imprecise anyway — the certificate we minted did carry a matching SAN, so the browser's complaint was an untrusted issuer, not a name mismatch.)

docs/app/faq.mdx
```diff
@@ -1140,7 +1139,7 @@ browser.
 
 [Check out the "Node Modules" example recipe.](/app/references/recipes#Fundamentals)
 
-### <Icon name="angle-right" /> Is there a way to give a proper SSL certificate to your proxy so the page doesn't show up as "not secure"?
+### <Icon name="angle-right" /> Is there a way to give a proper SSL certificate so the page doesn't show up as "not secure"?
 
 No, Cypress modifies network traffic in real time and therefore must sit between
 your server and the browser. There is no other way for us to achieve that.
```
AtofStryker: You were right that dropping "to your proxy" left the "No" unsupported. The cause turns out to be simpler than the original answer suggested: the page is flagged because the Cypress app itself is served over `http` on `localhost` — an https iframe can't lift the page's security state — so no certificate you supply would change it. That's true on both paths.

```suggestion
No. The Cypress app itself is served over `http` on `localhost`, so the browser
flags the page as not secure no matter what certificate your site uses. Cypress
must also be able to modify traffic in real time, which can require it to
present its own certificate for the site under test.
```

On the anchor change from the heading: I wouldn't bother with a redirect. Fragments never reach the server, so nothing can catch the old slug, and I grepped the repo — nothing links to it, so readers just land at the top of the FAQ.

docs/app/faq.mdx
```diff
@@ -1621,8 +1620,8 @@ XState model state library.
 ### <Icon name="angle-right" /> Can Cypress be used for performance testing?
 
 Cypress is not built for performance testing. Because Cypress instruments the
-page under test, proxies the network requests, and tightly controls the test
-steps, Cypress adds its own overhead. Thus, the performance numbers you get from
+page under test and tightly controls the test steps, Cypress adds its own
+overhead. Thus, the performance numbers you get from
```
AtofStryker: I'd swap the clause instead of dropping it — the overhead is still real, we just don't need to say how. Rewrapped while I was in here.

```suggestion
Cypress is not built for performance testing. Because Cypress instruments the
page under test, intercepts network requests, and tightly controls the test
steps, Cypress adds its own overhead. Thus, the performance numbers you get from
```

docs/app/guides/test-performance.mdx
```diff
@@ -545,10 +545,10 @@ See the [Network Requests guide](/app/guides/network-requests) for a full discus
 
 #### Intercept only the requests you need
 
-Using a wildcard pattern to observe or intercept every HTTP request creates real overhead. Cypress routes every intercepted request through its proxy, so every network call the page makes is inspected, matched, and processed. Broad wildcard patterns that match dozens or hundreds of requests per page load multiply that cost across the entire suite.
+Using a wildcard pattern to observe or intercept every HTTP request creates real overhead: every network call the page makes is inspected, matched, and processed. Broad wildcard patterns that match dozens or hundreds of requests per page load multiply that cost across the entire suite.
```
AtofStryker: As written this isn't really specific to wildcards — we inspect and match requests either way. The cost being warned about is per *matched* request, so I'd name that:

```suggestion
Using a wildcard pattern to observe or intercept every HTTP request creates real overhead: every request that matches an intercept is paused and handed to your test code before it continues. Broad wildcard patterns that match dozens or hundreds of requests per page load multiply that cost across the entire suite.
```

docs/app/references/troubleshooting.mdx
```diff
@@ -654,7 +654,7 @@ upgrade to the latest Cypress:
 #### Quick fastest-path check
 
 To confirm the problem is external and unblock yourself immediately, run with
-the bundled Electron browser, which doesn't go through the same proxied path:
+the bundled Electron browser, which doesn't go through the same network path:
```
AtofStryker: This one now names the wrong browser: Electron stays on the old path and Chrome/Edge are the ones that changed, so Electron isn't the browser that avoids it. The triage value holds without the mechanism:

```suggestion
To confirm the problem is external and unblock yourself immediately, run with
the bundled Electron browser to see whether the failure follows the browser:
```

docs/api/commands/visit.mdx
AtofStryker: You will see the header in dev tools. going to drop a comment below

AtofStryker:
@baus I took some of the ideas in here and broaden them in #6812. Is it OK if I close this PR in favor of #6812? I have quite a few suggestions / areas we need to update in addition to this and figured that would be the easiest thing.

PR #6815, opened 2026-08-18
docs: deprecate `cy.exec()` and `execTimeout` in favor of `cy.task()`

### LINKED ISSUE #34639: Deprecate `cy.exec` in favor of `cy.task`
## Summary

We plan to deprecate `cy.exec` and remove it in a breaking version. Everything `cy.exec` does today can be done with [`cy.task`](https://docs.cypress.io/api/commands/task), which we recommend as the replacement.

## Why we are making this change

`cy.exec` runs a command through the shell on the machine running Cypress. That means its behavior depends on your operating system, your shell, your login environment, and whether a terminal is attached. 

This has been a long-running source of failures that are difficult for us to fix and difficult for you to diagnose:

- Path and argument handling differs on Windows depending on which shell is installed ([#789](https://github.com/cypress-io/cypress/issues/789))
- Commands hang indefinitely under some CI executors when no interactive terminal is available ([#19869](https://github.com/cypress-io/cypress/issues/19869))
- The shell that resolves your `PATH` is not always the shell you configured, producing intermittent "command not found" failures ([#6081](https://github.com/cypress-io/cypress/issues/6081))
- Chromium can freeze when the Cypress app is launched as a background job ([#24255](https://github.com/cypress-io/cypress/issues/24255))
- Failures do not always surface the full error output ([#5470](https://github.com/cypress-io/cypress/issues/5470))

Several of these trace back to how a shell is discovered and invoked rather than to Cypress code, so they are not fixable without either reimplementing shell handling ourselves or narrowing what `cy.exec` supports.

`cy.task` avoids this class of problem. It runs in the Node process that already loads your Cypress configuration, so there is no shell resolution, no dependence on an attached terminal, and no per-platform quoting rules. It also gives you real return values instead of parsed stdout, and it lets you handle errors in Node rather than in a shell.

## Migration

Every `cy.exec` call moves to a task registered in `setupNodeEvents`. The examples below cover some common patterns.

### Cleaning up and verifying downloaded files

Before:

```js
cy.exec('rm -rf cypress/downloads/*')

cy.exec('ls -t cypress/downloads | head -1')
  .its('stdout')
  .then((filename) => {
    cy.readFile(`cypress/downloads/${filename.trim()}`)
  })
```

After, in `cypress.config.js`:

```js
const fs = require('fs')
const path = require('path')

  e2e: {
    setupNodeEvents(on, config) {
      const downloads = path.join(config.projectRoot, 'cypress', 'downloads')

      on('task', {
        clearDownloads() {
          fs.rmSync(downloads, { recursive: true, force: true })
          fs.mkdirSync(downloads, { recursive: true })

          return null
        },

        latestDownload() {
          const files = fs.readdirSync(downloads)
            .map((name) => ({
              name,
              time: fs.statSync(path.join(downloads, name)).mtimeMs,
            }))
            .sort((a, b) => b.time - a.time)

          return files.length ? path.join(downloads, files[0].name) : null
        },
      })
    },
  },
})
```

In your test:

```js
cy.task('clearDownloads')

cy.task('latestDownload').then((filepath) => {
  expect(filepath).to.be.a('string')
  cy.readFile(filepath)
})
```

### Seeding test data or resetting an environment

If you are calling a Node script, call the module directly instead of spawning a process:

```js
const seedDatabase = require('./scripts/seed-database')

on('task', {
  async seed(scenario) {
    await seedDatabase(scenario)

    return null
  },
})
```

If you are calling something that is not Node, such as a database client or a container command, spawn it from the task:

```js
const { execFileSync } = require('child_process')

on('task', {
  resetDatabase() {
    execFileSync('docker', ['exec', 'db', 'reset-script'], { stdio: 'pipe' })

    return null
  },
})
```

Using `execFileSync` with an argument array rather than a single command string avoids the shell entirely, which is what removes the platform differences.

### Running a CLI that is part of the system under test

If the command line tool is what you are testing, the same pattern applies, and you can return structured output rather than parsing stdout in the browser:

```js
const { execFileSync } = require('child_process')

on('task', {
  listBucket(bucket) {
    const output = execFileSync('aws', ['s3', 'ls', bucket], {
      encoding: 'utf8',
    })

    return output.split('\n').filter(Boolean)
  },
})
```

### Things to know when migrating

- A task must return a value or `null`. Returning `undefined` fails the task.
- Tasks time out after 60 seconds by default. Use the `taskTimeout` configuration option or pass a `timeout` option to `cy.task` for longer-running work.
- Anything you register in `setupNodeEvents` runs in Node, so `require`, `fs`, and `child_process` are all available to you.
- Tasks are shared across your whole project, so commands duplicated across many spec files usually collapse into a single task.

The [`cy.task` documentation](https://docs.cypress.io/api/commands/task) covers registration, arguments, and timeouts in full.

jennifer-shehane:
## Summary

Marks `cy.exec()` and the `execTimeout` configuration option as deprecated across the docs, adds `cy.task()` examples covering the workflows that previously needed `cy.exec()`, and points the guides that recommended `cy.exec()` at `cy.task()` instead.

## Why

`cy.exec()` and `execTimeout` are deprecated in Cypress 15.21.0 and will be removed in a future major version. `cy.exec()` runs commands through a shell on the machine running Cypress, which produces platform-specific behavior that can't be fixed without reimplementing the command: Windows path/argument handling varies by installed shell, commands hang under CI executors with no interactive terminal, `PATH` resolution produces intermittent "command not found" failures, and error output isn't always surfaced. `cy.task()` runs in the Node process that loads the Cypress configuration, so none of that applies.

Deprecation issue: cypress-io/cypress#34639
Implementation: cypress-io/cypress#34645

Closes cypress-io/cypress#34639

## Notes for reviewers

**Deprecation markers**

- `docs/api/commands/exec.mdx` — deprecated title/frontmatter (`deprecated_label` sidebar badge) and a `Deprecated` caution pointing at `cy.task()`, plus a `15.21.0` History row.
- `docs/app/references/configuration.mdx` — `execTimeout` badged deprecated in both the Timeouts table (with a pointer to `taskTimeout`) and the Test Configuration list.
- `docs/api/table-of-contents.mdx` — deprecated badge on `.exec()`. Prettier realigned the surrounding table columns, which is most of that file's diff.

**New `cy.task()` examples** (`docs/api/commands/task.mdx`)

- **Clean up and verify downloaded files** — `clearDownloads` / `latestDownload`.
- **Run an external command or CLI** — `child_process.execFileSync()` with an argument array, plus a variant that returns command output to the test.

**Guides pointed away from `cy.exec()`**: `faq`, `trade-offs`, `best-practices`, `introduction-to-cypress`, `testing-your-app` (its three seed examples now use a `resetAndSeedDatabase` task, with the registration shown), `why-cypress`, `getcookies`, `arch`, `testing-type`, and every `See also` list that linked to it.

**Decisions worth a look**

- No migration guide section — deliberately out of scope for this change.
- No `15.21.0` changelog entry; that lands via the separate `docs: cypress release` PR. Because `#15-21-0` doesn't exist yet and `onBrokenAnchors` is set to `throw`, the History row in `exec.mdx` uses a plain-text version rather than a changelog link. Worth linking once the release entry exists.
- In `auth0-authentication.mdx` I swapped `cy.exec('curl -4 icanhazip.com')` for `cy.request('https://ipv4.icanhazip.com')`. That's the one change that alters an example's approach rather than just its command — happy to revert if you'd rather leave that guide alone.
- The Web Servers anti-pattern in `best-practices.mdx` still names `cy.exec()`, since the warning remains accurate for users on current versions.

`npm run build` (with `onBrokenLinks`/`onBrokenAnchors: throw`) and `npm run lint` both pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JMnN3xs3rEPGzUN8ZdJtA9)_

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime impact; the Auth0 example behavior change (HTTP request vs shell curl) is a minor doc accuracy consideration for reviewers.
> 
> **Overview**
> This PR updates the Cypress docs to reflect **`cy.exec()`** and **`execTimeout`** being deprecated in **15.21.0**, steering readers toward **`cy.task()`** and **`taskTimeout`**.
> 
> The **`exec`** command page gets deprecated frontmatter, a caution callout, and a History row; **`configuration.mdx`** marks **`execTimeout`** deprecated in the timeouts table and test-configuration list; the API table of contents shows a deprecated badge on **`.exec()`**.
> 
> **`task.mdx`** gains new sections for **download folder cleanup/verification** and **running external CLIs** via **`execFileSync`** (with argument arrays).
> 
> Across guides and API cross-links, **`cy.exec()`** references are removed or replaced: seeding examples use a **`resetAndSeedDatabase`** task, conditional examples use **`cy.task()`**, incorrect-chaining examples use **`cy.task()`**, and the Auth0 guide uses **`cy.request`** for public IP instead of **`curl`** via exec.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19f98e520ca00692f415dce5fc97e76928fb64ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AtofStryker:
I reviewed this alongside the code PR. The deprecation mechanics match the `cy.end()`/`Cypress.env()` precedent and nothing overclaims what the `@deprecated` tags do in editors. Two fixes before merge, two optional nits:

1. The `exec.mdx` caution block should name the version, matching `end.mdx`/`env.mdx`: "`cy.exec()` was deprecated in Cypress `15.21.0` and will be removed in a future major version."
2. `platform.mdx` still has the same shell-out example this PR converted in `arch.mdx` and `testing-type.mdx` — its "Correct Usage" section models `cy.exec(cmd).its('stdout')`.
3. (optional) The new downloads example in `task.mdx` hardcodes the downloads path — `config.downloadsFolder` is the resolved, user-customizable value. Maybe also note `trashAssetsBeforeRuns` already clears it before `cypress run`.
4. (optional) `shadow.mdx` and `find.mdx` still use `cy.exec()` in "Incorrect Usage" examples, if you want the sweep complete.

jennifer-shehane:
Thanks for the review. Pushed 58d85d3 covering 2, 3, and 4. Pushing back on 1.

**1. Version in the `exec.mdx` caution** — keeping it version-free, deliberately. `15.21.0` isn't released yet, so unlike `end.mdx` and `env.mdx` there's no changelog anchor to point at, and `onBrokenAnchors` is set to `throw`. The version is recorded in the History table on the page, and once the release entry lands it can move into the caution with a working link. Happy to change it now if you'd rather match the precedent immediately.

**2. `platform.mdx`** — converted. Swapping the `dir`/`ls` shell-out to a task would have deleted the platform conditional that's the point of the page, so I replaced the example with a platform-specific modifier key instead:

```javascript
it('selects all text in the note', () => {
  // if macOS do one thing, else do another
  const modifier = Cypress.platform === 'darwin' ? '{cmd}' : '{ctrl}'

  cy.get('#note').type(`${modifier}a`)
})
```

Say the word if you'd rather keep it file-listing shaped.

**3. `downloadsFolder`** — good catch, `downloadsFolder` is `isFolder: true` so the resolved config value is already absolute. The task now reads `config.downloadsFolder` directly instead of rebuilding the path from `config.projectRoot`, and I added a note that `trashAssetsBeforeRuns` already trashes the folder before `cypress run`, so the task is for clearing between tests within a run.

**4. `shadow.mdx` / `find.mdx`** — done, both now use `cy.task('log', 'hello')` as the non-DOM-yielding command.

After this, the only `cy.exec()` mentions left outside `exec.mdx` are the `execTimeout` / `resolvedNodePath` rows in the configuration reference (which describe the option itself) and the Web Servers anti-pattern in `best-practices.mdx`, which I left because the warning is still accurate for anyone on a current version.

---
_Generated by [Claude Code](https://claude.ai/code)_

AtofStryker:
Verified 58d85d3 — 2, 3, and 4 all look good, and I confirmed the `trashAssetsBeforeRuns` note against the implementation: `cypress run` trashes `downloadsFolder` when the option is enabled, and only in run mode, so the "between tests within a run" framing is exactly right. The modifier-key example in `platform.mdx` works for me — no need to keep it file-listing shaped.

On 1, this is unblocked for two reasons:

- The precedent doesn't actually link the version — `end.mdx` and `env.mdx` both state it as plain text (`` `.end()` was deprecated in Cypress `15.15.0` and will be removed in a future major version ``), so `onBrokenAnchors` was never in play for matching them.
- The 15.21.0 changelog entry just merged in #6817 to `release/15.21.0` — the same branch this PR targets — so even a linked version resolves now.

Either way, you should now be able to add the version to the caution: "`cy.exec()` was deprecated in Cypress `15.21.0` and will be removed in a future major version."

jennifer-shehane:
Reversing my pushback on 1 — the version is in as of a8987c3, matching the `end.mdx` / `env.mdx` wording:

> `cy.exec()` was deprecated in Cypress `15.21.0` and will be removed in a future major version.

The `execTimeout` note in the configuration reference now names the version too. It's plain text rather than a changelog link since `#15-21-0` doesn't exist yet and `onBrokenAnchors` is set to `throw`; it can become a link once the release entry lands.

---
_Generated by [Claude Code](https://claude.ai/code)_

AtofStryker (approved):
All review items are resolved — the version wording in a8987c3 matches the `end.mdx`/`env.mdx` precedent exactly, and I verified the `trashAssetsBeforeRuns` note and the `downloadsFolder` change against the implementation. Thanks for the quick turnaround and for carrying the version into the `execTimeout` config row too. :ship:

PR #6818, opened 2026-08-18
fix: skip analytics under Cypress on the docs site

marktnoonan:
The Cypress suite visits every docs page (cypress/support/visitAllPages.ts), and CI serves a real production build, so the analytics tags fire on each visit, unless they are filtered out downstream. This PR avoids firing such events at all when Cypress is the visitor.

---

> [!NOTE]
> **Low Risk**
> Test-only Cypress networking configuration with no impact on production site behavior or user-facing docs.
> 
> **Overview**
> **Cypress e2e** now blocks outbound requests to Google Tag Manager, Google Analytics, FullStory, and Pendo while the suite crawls every docs page against a production build.
> 
> That stops those full-site visits from creating real analytics visitors in CI and local runs, complementing any in-app Cypress detection described in the PR goal.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c6cddcb9fad01c9bc02991120a947825a8227654. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jennifer-shehane:
Reviewed this and I think the goal (don't let the docs crawl inflate Pendo/GA4/FullStory) can be reached with a lot less machinery.

### The mechanism here does work

`window.Cypress` is reliable in this context — the proxy injects `packages/runner/injection/main.js` at the top of the AUT's `<head>`, and its first statement is `window.Cypress = parent.Cypress`, so it's set before Docusaurus's head tags run.

The UA half is weaker: `Cypress/<version>` appears in the **Electron** default UA only, and it's explicitly stripped when `experimentalModifyObstructiveThirdPartyCode` is enabled (`packages/server/lib/browsers/electron.ts`). Under `--browser chrome` it's absent entirely, so `window.Cypress` is doing essentially all the work.

CI doesn't test a deployed build — `.circleci/config.yml` runs `npm run build` itself and then `npm run serve` on `localhost:3000`. The artifact the Cypress suite crawls is built by CI, for CI; Netlify/Vercel build production separately. So nothing requires the build under test to carry real analytics, which opens up two simpler options.

### Option 1 — `blockHosts` (no site changes at all)

Cypress has a feature for this. In `cypress.config.ts`:

```ts
blockHosts: [
  '*googletagmanager.com',
  '*google-analytics.com',
  '*fullstory.com',
  '*pendo.io',
],
```

The proxy 503s matching requests, so `gtm.js` never loads and neither does anything the container carries — including tags added to GTM later. That replaces all four changed files with four lines, ships nothing Cypress-specific to real visitors, and also covers local runs against a production build. It's consistent with how Osano is already handled (stubbed in `cypress/support/e2e.ts`).

The inline snippets still execute and create `dataLayer`, but nothing is sent. If you want zero analytics markup on the page in CI, see option 2.

### Option 2 — build-time flag (no vendored plugin, no runtime detection)

`preset-classic` only registers the GTM plugin when the option is truthy, so this is sufficient in `docusaurus.config.js`:

```js
const analytics = !process.env.DISABLE_ANALYTICS
// ...
        ...(analytics ? { googleTagManager: { containerId: 'GTM-KNKBWLD' } } : {}),
```

plus one line in `plugins/fullstory.js` (`const isProd = process.env.NODE_ENV === 'production' && analytics`) and `DISABLE_ANALYTICS=1 npm run build` in the `install-and-persist` job. Same result, keeps the official plugin, no browser-side detection.

### Concerns with the current approach

1. **Vendored upstream code.** `plugins/googleTagManager.js` forks ~30 lines of `@docusaurus/plugin-google-tag-manager` while that plugin stays installed via `preset-classic`. It will drift — Docusaurus v4 is already changing this area (the deprecated `googleAnalytics` option now throws).
2. **Cypress detection ships to production** — a permanent branch in the HTML every real visitor downloads, to solve a CI-only problem.
3. `id: 'docs-google-tag-manager'` is inert for a single-instance plugin.


---
_Generated by [Claude Code](https://claude.ai/code)_

jennifer-shehane (changes requested):
See comment

marktnoonan:
blockHosts is definitely better, yes. I'm surprised that wasn't the first solution offered: Cursor's comment:

> blockHosts is even documented in this repo for analytics in tests.

Updated the PR

jennifer-shehane (approved):

PR #6867, opened 2026-08-27
docs: remove `cy.exec()` and `execTimeout` for Cypress 16

mschile:
## Summary

Documents the removal of `cy.exec()` and the `execTimeout` configuration option in Cypress 16. Deletes the `cy.exec()` API page, adds a migration section covering the move to `cy.task()`, and removes `execTimeout` from the configuration reference.

## Why

`cy.exec()` was deprecated in Cypress `15.21.0` (cypress-io/cypress#34639) and is removed in 16.0 by cypress-io/cypress#34697, which closes cypress-io/cypress#34694. Calling it now throws an error pointing at `cy.task()`, and setting `execTimeout` prints an "option removed" warning.

`cy.exec()` ran a command through the shell on the machine running Cypress, so its behavior depended on the operating system, the shell, the login environment, and whether a terminal was attached. `cy.task()` runs in the Node.js process that already loads the Cypress configuration, which removes that whole class of problem, so the migration section is written around moving each call into a task.

## Notes for reviewers

This targets `release/16.0.0` rather than `main`, since that is where the "Migrating to Cypress 16.0" section lives. The page removal follows the established pattern: delete the `.mdx`, drop the table of contents row, and add a `301` in `netlify.toml`. The redirect points at `/api/commands/task`, matching how `route`, `route2`, and `server` redirect to `intercept`; `/api/table-of-contents` is for removals with no replacement, such as `cy.end()`.

One thing differs from that precedent and is worth a look. `cy.end()` had no surviving links anywhere in the repo, but `/api/commands/exec` was linked from 14 places, nine of them historical changelog entries. Since `onBrokenLinks` is `throw`, those had to be de-linked or the build fails. They are now plain code spans, with the surrounding wording untouched, so the changelog still records what happened in each release without pointing at a page that no longer exists.

Also updated the AI upgrade prompt in the migration guide. It enumerates the breaking changes to work through, so leaving `cy.exec()` out of step 5 would have quietly skipped it during an assisted upgrade.

Verified with `npm run lint:fix` and `npm run build`. The build is the meaningful check here: it resolves every link and anchor, including the 14 de-linked references and the new `#cyexec-removed` anchor used by the summary table.

---

> [!NOTE]
> **Low Risk**
> Documentation and redirect changes only; no product runtime or test behavior is modified.
> 
> **Overview**
> Documents **Cypress 16** removal of **`cy.exec()`** and **`execTimeout`**: the dedicated API page is deleted, the API TOC and configuration reference no longer list them, and the 16.0 config history records removal in favor of **`taskTimeout`** with **`cy.task()`**.
> 
> The **16.0 migration guide** gains a **`cy.exec()` removed** section (before/after examples, `setupNodeEvents` tasks, `execFileSync` for CLIs, `execTimeout` cleanup) plus a summary-table row; the **AI upgrade prompt** now tells assistants to replace **`cy.exec()`** and drop **`execTimeout`** in step 5.
> 
> **`/api/commands/exec`** **301** redirects to **`/api/commands/task`**. Historical **changelog** mentions of **`cy.exec()`** are plain code spans so **`onBrokenLinks`** stays green after the page is gone.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 30bf6b7fe6a8c2ffc5fd300c93ea33fb67cf4139. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


netlify.toml
```diff
@@ -577,6 +577,12 @@
   status = 301
   force = true
 
+[[redirects]]
+  from = "/api/commands/exec"
+  to = "/api/table-of-contents"
```
jennifer-shehane: I wonder if this should reroute to task?? May be weird.

jennifer-shehane (approved):

netlify.toml
mschile: I think that makes sense, updated [154f1ac](https://github.com/cypress-io/cypress-documentation/pull/6867/commits/154f1ac52f5fac1b576a903f0142d6e4c30d73c9).

AtofStryker (approved):
Nice work on this one — the migration section reads really well, and redirecting to `/api/commands/task` instead of the table of contents was the right call since `cy.exec()` actually has a replacement page.

Approving, with one ask: can you add the imports for `latestDownload`? With a single block, `:::cypress-config-plugin-example` renders a complete config file, so on the deploy preview `fs`, `path`, and `downloads` come out undefined — copying it throws a `ReferenceError`. The `cy.task()` page's version of this example has the shape to borrow: an imports block, then a body starting with `const downloads = config.downloadsFolder`. Worth doing the same for the `resetDatabase` snippet, which gets the `execFileSync` import for free.

Two small nits, take or leave:

- Three of the de-linked changelog entries (1.1.3, 1.0.0-era, 0.20.1) never had backticks, so they're now plain text `cy.exec()`. Backticks would make all nine match.
- The PR description still says the redirect goes to `/api/table-of-contents` — stale after the last commit.

Everything else checked out: `cy.exec()` throws and `execTimeout` warns in [#34697](https://github.com/cypress-io/cypress/pull/34697) exactly as described, and the new `cy.task()` anchors all resolve. Should be quick to wrap up!

PR #6889, opened 2026-09-08
docs: add missing visibilityStrategy migration patterns

### LINKED ISSUE #6887: Visibility strategy: rewrite patterns only cover negative assertions
"Updating tests that depended on the legacy algorithm" gives three rewrite patterns: scroll
clipping, collapsed `overflow: hidden` containers, and text truncation. All three rewrite a
`should('not.be.visible')`.

Two common cases have no pattern:

1. **The positive case.** `cy.get(sel).should('be.visible').click()` is the documented way to wait
   until an element is ready to interact with. Under `visibilityStrategy: "modern"` that wait is
   satisfied while the element is still clipped or still covered, so the following command races the
   UI. A pattern for "wait until the user could actually see and use this" would cover by far the
   most common usage.
2. **Coverage.** None of the three patterns addresses an element covered by an opaque `fixed` or
   `sticky` element, which `"modern"` reports as visible.

Given that `visibilityStrategy` is documented as deprecated and slated for removal, these two
patterns are the migration path for anyone who relied on the legacy behaviour.

Related: cypress-io/cypress#34795, which asks that the `visibilityStrategy` option not be removed until these patterns exist.

jennifer-shehane:
## Summary

Fills the gaps in "Updating tests that depended on the legacy algorithm" on the [Interacting with Elements](https://docs.cypress.io/app/core-concepts/interacting-with-elements) page. The section previously offered three rewrite patterns, all of them rewrites of `should('not.be.visible')`, leaving the most common usage and two of the legacy-only behaviors we list with no migration path at all. This adds those patterns, corrects one inaccurate bullet, and moves the reusable checks into custom queries.

## Why

Closes #6887

`visibilityStrategy` is documented as deprecated and slated for removal, so this section is the migration path for anyone who relied on the legacy algorithm. Two gaps made that path incomplete:

**The positive case had no pattern.** `cy.get(sel).should('be.visible').click()` is the most common usage by far, and under `'modern'` the assertion stops acting as the wait it used to be. Investigating this turned up something worth documenting plainly: action commands never changed. `ensureElIsNotCovered` in `packages/driver/src/cy/actionability.ts` sits outside the strategy branch, so a covered element still has its `.click()` refused under `'modern'`. The assertion was never what made the action wait, so in most cases the fix is to delete it. The page now says that, and covers the "they disappeared from the page" failure that shows up when the earlier-resolving wait lets an action start sooner.

**Coverage by a `fixed` or `sticky` element had no pattern**, despite being on our own list of legacy-only behaviors.

Two further problems surfaced while writing it:

- The `transform: scale(0)` bullet was wrong. `modernIsHidden` has an explicit zero-dimension guard on `getBoundingClientRect()`, so the modern algorithm catches it too. Removed from the legacy-only list.
- `backface-visibility` rotation is on that list and also had no pattern. Added.

The three existing patterns used `.then()`, which does not retry. They were replacing a retrying assertion with a non-retrying one and quietly giving up that behavior. All patterns now use `.should()`.

## Notes for reviewers

**Please look hardest at the two custom queries in `cypress/support/commands.js`.** They are the most load-bearing new content, and a self-review caught several defects in an earlier revision of this branch that I have since fixed. Worth confirming I got them right:

- `isCovered()` mirrors legacy's `elIsNotElementFromPoint`, including the `pointer-events: none` exemption where the hit test falls through to an ancestor.
- Both queries check every matched element via `.every()`, since `should('not.be.visible')` applies to the whole set rather than just the first element.
- `isVisibleToUser()` hit tests every element it is given, which is stricter than legacy's fixed/sticky-only rule. This is called out in the page, because over-applying the coverage check is exactly the trap described in cypress-io/cypress#34795, where it turned passing specs red.

**Queries rather than commands.** These use `Cypress.Commands.addQuery`, not `Cypress.Commands.add`. Both checks are synchronous and idempotent, and a non-retrying command would receive its subject once and hold it, undoing the retry behavior the section promises. Happy to revisit if you would rather they read as plain commands, but the retry loss is the tradeoff.

**One thing I deliberately did not include.** An earlier revision offered a geometric fallback for the `backface-visibility` case, reading `m33` off the element's computed transform. It could never fail: in a typical card flip the rotation lives on an ancestor and the face itself carries a static `rotateY(180deg)`, so the value is identical in both states. Reproducing the legacy check faithfully means accumulating the transform across every `preserve-3d` ancestor, which is more DOM logic than the assertion is worth. The page now recommends asserting on component state and explains why.

**Also unresolved, and out of scope here.** `Cypress.dom.isStrictlyHidden` and `Cypress.dom.isHiddenByAncestors` are still attached to `Cypress.dom` at runtime and are not strategy-gated. Together they are the legacy algorithm, which undercuts the "3,000 lines to reimplement" premise in cypress-io/cypress#34795. They are undocumented and absent from `cli/types/cypress.d.ts`, so I have not documented them as a supported path. Whether we want to bless that escape hatch is a product decision worth making deliberately.

**Scope note.** I touched two pre-existing em dashes outside the migration section (the `opacity: 0` bullet in Default Behavior, and the collapsed-container paragraph) so the page reads consistently. Easy to drop if you would rather keep the diff tight.

**Verification.** `npm run build` passes. That matters more than usual here: the site sets `onBrokenAnchors: 'throw'` and `onBrokenLinks: 'throw'`, so the build confirms the new cross-reference and every new link resolves. I also checked the rendered anchor ids and the ten `codeBlockTitle` elements in `dist/`.

https://claude.ai/code/session_01BZraHsGceNoUhqc9oKzMzQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01BZraHsGceNoUhqc9oKzMzQ)_

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes with no runtime or test-runner behavior; main risk is inaccurate migration advice, which reviewers should validate against the custom query examples.
> 
> **Overview**
> Expands **Updating tests that depended on the legacy algorithm** on the Interacting with Elements page so teams migrating off deprecated `visibilityStrategy: 'legacy'` have concrete rewrites for cases that were missing or wrong.
> 
> The guide now distinguishes **`should('be.visible')` as a pre-action wait** (often redundant because action commands still enforce coverage and scroll) from **`should('not.be.visible')` assertions** that fail under the modern `checkVisibility()` path. It documents custom query examples **`isCovered`** and **`isVisibleToUser`**, adds patterns for **fixed/sticky overlap**, **backface-hidden rotations** (prefer component state over geometry), and tightens existing scroll/collapse/ellipsis examples by switching **`.then()` to `.should()`** so retries match the assertions they replace.
> 
> **Docs corrections:** removes **`transform: scale(0)`** from the legacy-only list (modern already treats zero bounding rects as hidden) and makes minor punctuation/consistency edits elsewhere on the page.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0be1d81b3f989f29d33e16a26654bfce4de6969d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/app/core-concepts/interacting-with-elements.mdx
+++ b/docs/app/core-concepts/interacting-with-elements.mdx
@@ -25,7 +25,7 @@ As of Cypress 16, the default visibility algorithm delegates to the browser's na
   - `display: none` on the element or any ancestor.
   - `visibility: hidden` or `visibility: collapse`.
   - `content-visibility: hidden`, or `content-visibility: auto` when not rendered.
-  - `opacity: 0` (when asserting visibility — see Opacity note below).
+  - `opacity: 0` (when asserting visibility, see the Opacity note below).
 
 :::info
 
@@ -93,13 +93,13 @@ Every pattern below uses [`.should()`](/api/commands/should) rather than [`.then
 
 ##### Waiting for an element before you interact with it
 
-`cy.get(selector).should('be.visible').click()` is a widespread pattern. Its value was never only the assertion — under the legacy algorithm, the assertion also kept retrying until the element was no longer clipped or covered.
+`cy.get(selector).should('be.visible').click()` is a widespread pattern. Its value was never only the assertion. Under the legacy algorithm, the assertion also kept retrying until the element was no longer clipped or covered.
 
 Under the modern algorithm that assertion resolves as soon as the browser considers the element rendered, so the wait ends earlier than it used to.
 
 **Action commands did not change.** Before every action, Cypress still scrolls the element into view and then verifies that it is not hidden, disabled, detached, readonly, animating, or [covered by another element](#Covering). The coverage check runs under both strategies, so `.click()` on a covered element is still refused under `'modern'`.
 
-That means the assertion was never what made the action wait. When an action follows, remove it — the action command already waits for strictly more conditions than `be.visible` does:
+That means the assertion was never what made the action wait. When an action follows, remove it. The action command already waits for strictly more conditions than `be.visible` does:
 
 ```javascript
 // Before
@@ -117,7 +117,7 @@ cy.get('#submit').click()
 
 Because the assertion resolves sooner, a subsequent action starts sooner. If your application replaces the element during that interval, the action fails with `we initially found matching element(s), but while waiting for them to become actionable, they disappeared from the page`. The message describes a detached element and does not mention visibility, which makes the cause easy to miss.
 
-Removing the redundant `should('be.visible')` does not fix this on its own. Wait on an application-level signal that the element is settled — a loading indicator disappearing, a request completing via [`cy.intercept()`](/api/commands/intercept), or a state class or attribute your application sets — and then perform the action.
+Removing the redundant `should('be.visible')` does not fix this on its own. Wait on an application-level signal that the element is settled, such as a loading indicator disappearing, a request completing via [`cy.intercept()`](/api/commands/intercept), or a state class or attribute your application sets. Then perform the action.
 
 :::
 
@@ -181,7 +181,7 @@ cy.get('#scroll-container button').should(($el) => {
 
 ##### Children inside a collapsed container that hides via `overflow: hidden` + `max-height: 0`
 
-Many UI libraries hide collapsible content by setting `max-height: 0` and `overflow: hidden` on a wrapper. The wrapper itself reports as hidden under the modern algorithm (its bounding rect collapses to zero), but children with their own non-zero dimensions inside that wrapper are not detected as hidden — `checkVisibility()` doesn't account for ancestor clipping. Most libraries also set `aria-hidden="true"` on the closed wrapper — assert on that instead:
+Many UI libraries hide collapsible content by setting `max-height: 0` and `overflow: hidden` on a wrapper. The wrapper itself reports as hidden under the modern algorithm (its bounding rect collapses to zero), but children with their own non-zero dimensions inside that wrapper are not detected as hidden, because `checkVisibility()` doesn't account for ancestor clipping. Most libraries also set `aria-hidden="true"` on the closed wrapper, so assert on that instead:
 
 ```javascript
 // Before
--- a/docs/app/core-concepts/interacting-with-elements.mdx
+++ b/docs/app/core-concepts/interacting-with-elements.mdx
@@ -109,57 +109,54 @@ cy.get('#submit').should('be.visible').click()
 cy.get('#submit').click()
 ```
 
-:::info
-
-<strong>
-  If a `.click()` or `.type()` now fails with "they disappeared from the page"
-</strong>
-
 Because the assertion passes sooner, the action that follows it starts sooner. If your application replaces the element in that window, the action fails with `we initially found matching element(s), but while waiting for them to become actionable, they disappeared from the page`. The error describes a detached element and never mentions visibility, so the cause is easy to miss.
 
 Removing the redundant `should('be.visible')` does not fix this on its own. Wait on a signal from your application that the element has settled, such as a loading indicator disappearing, a request completing with [`cy.intercept()`](/api/commands/intercept), or a class or attribute your application sets. Then perform the action.
 
-:::
-
-When you need to assert that a user can see and use an element, combine the built-in assertion with a coverage check in a single `.should()`:
+When you need to assert that a user can see and use an element, combine [`Cypress.dom.isVisible()`](/api/cypress-api/dom) with a coverage check. Define both as [custom queries](/api/cypress-api/custom-queries) so you can reuse them across your suite:
 
 ```javascript
-cy.get('#confirmation-banner').should(($el) => {
-  expect($el).to.be.visible
-
-  const rect = $el[0].getBoundingClientRect()
-  const elAtPoint = $el[0].ownerDocument.elementFromPoint(
+// cypress/support/commands.js
+const isCovered = (el) => {
+  const rect = el.getBoundingClientRect()
+  const elAtPoint = el.ownerDocument.elementFromPoint(
     rect.left + rect.width / 2,
     rect.top + rect.height / 2
   )
 
-  expect($el[0].contains(elAtPoint), 'element is not covered').to.be.true
+  return !el.contains(elAtPoint)
+}
+
+Cypress.Commands.addQuery('isCovered', function () {
+  return ($el) => isCovered($el[0])
+})
+
+Cypress.Commands.addQuery('isVisibleToUser', function () {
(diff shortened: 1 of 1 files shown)


jennifer-shehane:
@jgonzalezalbisu do you feel this is helpful?

jgonzalezalbisu:
@jennifer-shehane yes, genuinely helpful — thank you for turning this around so quickly, and for going well past what I reported.

The split between "assertion used as a wait" and "assertion that something is hidden" is the distinction that was missing, and the first half is answered better here than in what I asked for. Action commands never changed, so for `should('be.visible').click()` the fix is usually to delete the assertion. I had verified that coverage is still enforced before actions and used it to argue for exposing a strict check; turning it into the migration pattern instead is the more useful conclusion, and it means most people's fix is to remove a line rather than add one.

Two other things I appreciated: the `.then()` → `.should()` correction on the pre-existing patterns is a real fix, since those were quietly trading a retrying assertion for a non-retrying one. And the paragraph about `we initially found matching element(s)... they disappeared from the page` describes exactly the failure I hit; nothing in that error points at visibility, and I would not have connected the two without this page saying so.

Also happy to confirm from the outside that removing the `transform: scale(0)` bullet is right — `modernIsHidden`'s zero-dimension guard catches it, and it reports as hidden under both strategies.

### On the internal functions — you're right, and my premise was overstated

I checked before saying so: `Cypress.dom.isStrictlyHidden` and `isHiddenByAncestors` are both functions on `Cypress.dom` at runtime in 16.0.0, neither is strategy-gated, and composing them the way the driver's own legacy `isHidden` does — including passing `isHidden` as the `recurse` argument, which matters for `<option>` inside `<select>` — agrees with the legacy algorithm on everything I could throw at it: 138 elements across a real page plus injected edge cases (`<option>`/`<optgroup>` in visible and hidden selects, `display: contents`, `backface-visibility`, `visibility: collapse`, clipping, scroll-out, coverage, sticky, empty inline), with zero disagreements.

So "thousands of lines to reimplement" in cypress-io/cypress#34795 is wrong, and it's about six lines against those two functions. I'd rather state that accurately than leave the exaggeration standing. What remains of the concern is much smaller: they're undocumented and absent from `cli/types/cypress.d.ts`, so building on them is the kind of unsupported-internals dependency that breaks silently on a major. Documenting and typing them would resolve most of what I was asking for, and it's a far smaller ask than a new assertion.

### On the two queries

You flagged these as the load-bearing part, so:

**`isVisibleToUser()` may promise more than it delivers.** It's `Cypress.dom.isVisible()` plus a coverage check, and `Cypress.dom.isVisible()` follows the configured strategy — so under `'modern'` the query returns modern plus coverage, still missing `overflow` clipping and scroll-out, two of the three legacy-only behaviors this page lists. Someone reading the name could reasonably expect all three back. It also means the query behaves differently depending on the strategy in effect, becoming legacy-plus-coverage under `'legacy'`. A sentence on what it does and does not cover would set expectations.

**`every()` versus `some()`.** The page says both queries check every matched element "the way `be.visible` and `not.be.visible` do". That holds for the negative case, but chai-jQuery implements `visible` as `$el.is(':visible')`, and jQuery's `.is()` is true when *any* element matches — on a set of two where only the first is visible, `should('be.visible')` passes under both strategies. So `isVisibleToUser()` is stricter than the assertion it stands in for in a second way, on top of the hit testing. Either `some()` or a small correction to that sentence.

**The off-screen consequence might be worth repeating** in the `isVisibleToUser()` paragraph. The `isCovered()` section explains that `elementFromPoint` returns `null` outside the viewport and that this counts as covered, which matches legacy — but legacy only hit tested `fixed`/`sticky` elements, which are in the viewport by definition, so it never met that case. Applied to arbitrary elements, anything below the fold reports as not visible to the user, even though a user could scroll to it and Cypress itself scrolls before acting. It's the most likely way someone gets caught out, and it currently sits a section away from the warning.

`isCovered()` itself reads as a faithful port of `elIsNotElementFromPoint` — descendant-or-self short circuit, the `pointer-events: none` fall-through on the element or its parent combined with the hit result being an ancestor, and `null` treated as covered. Two narrow divergences, neither a blocker for documentation: legacy takes the center point from `getElementPositioning().fromElViewport`, which accounts for nested-iframe offset and scaling where `getBoundingClientRect()` does not, and legacy's ancestor walk crosses shadow boundaries where `Node.contains()` does not.

### Two additions you might consider

**`visibilityStrategy` also works as a suite and test override, and the docs don't mention it.** With `'modern'` configured globally, both `describe('...', { visibilityStrategy: 'legacy' }, ...)` and the same override on an individual `it` take effect — I confirmed both. For a migration page that seems worth a line: it turns an all-or-nothing global flag into something you can scope to the handful of suites that need it while everything else moves to `'modern'`. That reads much more like the temporary migration aid the option is described as.

**A short "which pattern do I need?" table at the top of the section.** There are six patterns now, and someone arriving from a red test has to read all of them to find theirs. Three columns — what you see failing, what changed, which pattern — would get people to the right one quickly. Cheapest of these suggestions and probably the highest impact for whoever lands here mid-debugging.

Either way, this is a big improvement on where the section was. We're holding off on changing anything on our side until this and cypress-io/cypress#34795 settle, since the right approach depends on which of these becomes the supported path — no rush from us.

---

**TL;DR** — three things worth changing, all in the two queries:

1. `isVisibleToUser()` is `Cypress.dom.isVisible()` plus coverage, so under `'modern'` it still misses `overflow` clipping and scroll-out — two of the three legacy-only behaviors this page lists. Worth a line on what it does and does not cover.
2. It uses `every()`, but `be.visible` is effectively `some()` (chai-jQuery's `$el.is(':visible')`), so it is stricter than the assertion it replaces. Either `some()` or a correction to that sentence.
3. The off-screen consequence of `elementFromPoint` returning `null` deserves repeating in the `isVisibleToUser()` paragraph, not only under `isCovered()`.

Two optional additions: document that `visibilityStrategy` also works as a suite and test override, and a short "which pattern do I need?" table at the top of the section.

Everything else above is confirmation rather than a request — including that you're right about `Cypress.dom.isStrictlyHidden` / `isHiddenByAncestors`, and that my "thousands of lines" premise in cypress-io/cypress#34795 was overstated.

jennifer-shehane:
Made some updates based on feedback.

jgonzalezalbisu:
@jennifer-shehane thanks for picking all three up, and for turning the `isVisibleToUser()` caveats into a clearer list than the one I described — "reach for it only where you expect the element to be on screen already" is the rule I was circling without stating.

Appreciate how quickly this moved.

PR #6909, opened 2026-09-17
docs(changelog): note 16.0.0 dropped arm64 support on Ubuntu 22.04

baus:
## Summary

Adds a Breaking Changes entry to the 16.0.0 changelog recording that Cypress 16 no longer runs on Linux arm64 under Ubuntu 22.04.

Cypress 16.0.0 requires `GLIBCXX_3.4.32` on arm64, which Ubuntu 22.04 does not provide — Cypress exits at startup with ``version `GLIBCXX_3.4.32' not found``. This affects the `ubuntu-22.04-arm` GitHub Actions runner. Linux x64 is unaffected and still supports Ubuntu >=22.04.

The install page was already corrected in #6888, but the 16.0.0 changelog did not mention it, and the changelog is where users look when an upgrade breaks a previously working setup.

Reported in cypress-io/cypress#34792.

## Notes

- Placed next to the Node.js support bullet, since both are platform requirements.
- The entry is added to an already-released 16.0.0 section; `cli/CHANGELOG.md` in `cypress-io/cypress` is intentionally left untouched.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changelog update with no runtime or build behavior changes.
> 
> **Overview**
> Adds a **Breaking Changes** bullet to the Cypress **16.0.0** section in `changelog.mdx`, placed after the Node.js version requirement.
> 
> The new entry documents that **Linux arm64** needs Ubuntu **≥24.04** (or a distro with `GLIBCXX_3.4.32+`), calls out the startup failure on Ubuntu 22.04 arm64 including **`ubuntu-22.04-arm`**, notes that **Linux x64** is unchanged, and links to the install **Operating System** docs and issue **#34792**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a80ec332eda51a72c8bf0f2b9994cf663459ea7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/app/releases/changelog.mdx
```diff
@@ -58,6 +58,7 @@ Refer to the [16 Migration Guide](/app/references/migration-guide#Migrating-to-C
...
 - Removed support for Node.js 20 and Node.js 25. Cypress now requires Node.js 22.x, 24.x, or >=26.x. Addresses [#33705](https://github.com/cypress-io/cypress/issues/33705) and [#33778](https://github.com/cypress-io/cypress/issues/33778).
+- Linux arm64 now requires Ubuntu >=24.04, or another distribution providing `GLIBCXX_3.4.32` or later. On Ubuntu 22.04 arm64 — including the `ubuntu-22.04-arm` GitHub Actions runner — Cypress exits at startup with ``version `GLIBCXX_3.4.32' not found``. Linux x64 is unaffected and continues to support Ubuntu >=22.04. See [Operating System](/app/get-started/install-cypress#Operating-System). Addresses [#34792](https://github.com/cypress-io/cypress/issues/34792).
```
AtofStryker, line 61: States the requirement and which distros meet it, instead of the single Ubuntu 22.04 symptom. Debian 12 arm64 and Debian 11 fall out of the same change.

````suggestion
- Linux arm64 now requires a distribution that ships `GLIBCXX_3.4.32` or later (libstdc++ from GCC 13.2+). Ubuntu >=24.04, Debian >=13, and Fedora >=43 meet this requirement. Linux x64 continues to support Ubuntu >=22.04 and Debian >=12. Debian 11 is no longer supported on any architecture, since its LTS ended in August 2026. See [Operating System](/app/get-started/install-cypress#Operating-System). Addresses [#34792](https://github.com/cypress-io/cypress/issues/34792).
````

AtofStryker (changes requested):
Thanks for picking this up. The entry is right about arm64, but I dug into the shipped 16.1.0 binaries and the picture is a bit wider than Ubuntu 22.04, so I'd like to adjust the wording and the install page together.

What I checked: I parsed the version requirements out of the linux-x64 and linux-arm64 addons in the 16.1.0 zips, then ran `ldd -r` against them inside `debian:11`, `debian:12`, and `debian:13` containers on both architectures.

| Distro | x64 | arm64 |
|---|---|---|
| Debian 11 | `better_sqlite3.node` fails to load (needs `GLIBCXX_3.4.29`) | fails at startup |
| Debian 12 | OK | fails at startup |
| Debian 13 | OK | OK |
| Ubuntu 22.04 | OK | fails at startup |
| Ubuntu 24.04 | OK | OK |

Two things fall out of that:

1. **Debian 12 arm64 is broken the same way Ubuntu 22.04 arm64 is.** The addon needs libstdc++ from GCC 13.2 or newer, and Debian 12 ships GCC 12. The install page still says `Debian >=11` with no arm64 qualifier, so a Debian 12 arm64 user reading this entry and following the link would conclude they're supported.
2. **Debian 11 x64 is not fully unaffected.** The better-sqlite3 binding moved from `GLIBCXX_3.4.20` to `3.4.29` in 16, and Debian 11 tops out at `3.4.28`. It loads lazily, so `cypress run` starts fine but Test Replay fails when it opens its database. Debian 11 LTS ended on 2026-08-31, so I'd rather drop it from the supported list than document a partial break.

Concretely I'm asking for:

- Reword the entry to state the requirement rather than the Ubuntu 22.04 symptom (suggestion inline). This also clears the bugbot em-dash comment.
- Update the install page in this PR so the link resolves to something consistent: `Debian >=12 _(arm64 requires Debian >=13)_`.
- Add one line to the 16.0 migration guide requirements table. Mike noted on [#34792](https://github.com/cypress-io/cypress/issues/34792) that the guide does not mention this. It should go near the "Node.js 20 and 25 no longer supported" section. Before this I was fine with just updating the changelog, but now that the impact is to multiple distros, it's likely worth the callout.

Happy to pair on any of it.

docs/app/releases/changelog.mdx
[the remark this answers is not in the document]
AtofStryker: Agreed. The suggestion above rewrites the bullet without the runner aside, so both em dashes go away.

MikeMcC399: The Debian entry in https://docs.cypress.io/app/get-started/install-cypress#Operating-System would need updating for consistency, since it currently states:

> Debian >=11

MikeMcC399:
- See also https://github.com/cypress-io/cypress-documentation/issues/6936

docs/app/releases/changelog.mdx
MikeMcC399: - https://github.com/cypress-io/cypress-documentation/pull/6937 resolved my previous comment (now hidden)

PR #6912, opened 2026-09-18
[DO NOT MERGE] fix(favicon): replace the icon set and stop Chrome ignoring the SVG

emilmilanov:
> **Draft — verification only, not for merge.** Opened to confirm the favicon behaves correctly on
> docs.cypress.io before the shared package lands. The real change will come from
> `@cypress-design/favicon` (cypress-io/cypress-design#723) once that is published.

## What & why

Two problems, one of which is invisible rather than broken.

**The icon has no background.** Every frame of `static/favicon.ico` has a fully transparent centre —
no disc, just the green ring and a navy `#1B1E2E` wordmark floating on nothing. On Chrome's dark tab
strip (`#35363A`) that is close to unreadable.

**Chrome was never using our adaptive SVG.** `plugins/fav-icon.js` injects `<link rel="icon"
href="/favicon.ico" sizes="any">`, the widely copied recipe for making browsers prefer an SVG.
Measured against **Chrome 153.0.8010.48** and **Safari 26.5** by serving five markup variants and
recording which file each engine actually fetched, it does the opposite:

| Markup | Chrome 153 | Safari 26.5 | Fetched |
| --- | --- | --- | --- |
| SVG first, `sizes="any"` on ICO (what we ship) | ICO | ICO | **ICO only** |
| `sizes="any"` on both | ICO | ICO | **ICO only** |
| Explicit sizes on the ICO | ICO | ICO | **ICO only** |
| ICO first, SVG last | ICO | ICO | both |
| **SVG only, no ICO link** | **SVG, adapts** | **SVG** | SVG |

A declared ICO wins outright in both engines, and in three of four variants Chrome did not even
download the SVG. Removing that tag is what makes the adaptive icon work. The file is still shipped
and still served at `/favicon.ico` — crawlers, unfurlers, RSS readers and Safari before 26 request it
there by convention, without reading our markup.

## How

`plugins/fav-icon.js` is the only source of these head tags (`docusaurus.config.js` sets
`favicon: undefined`). It now injects three: the SVG, the apple-touch icon, and a manifest. The
reasoning for the omitted ICO tag is in a comment there, so it does not get re-added.

Assets are generated from one adaptive source SVG. `favicon.ico` is now a real container with five
PNG frames (16/24/32/48/64) baked from the **dark** variant with the disc filled — it cannot adapt,
so it has to be one image that stays legible on either tab strip. Large rasters are supersampled to
remove a seam where the ring's four quarter paths meet.

## Known limitation

Safari renders the SVG but never evaluates `prefers-color-scheme` — it always resolves light. A
Safari dark-mode tab shows the light variant while Chrome and Firefox show the dark one.


emilmilanov:
Superseded by #6940, which adopts the published @cypress-design/favicon instead of inlining the assets.

